PR #3, opened 2020-08-11
Fix dependencies to point to new agent repo

empovit:
Replace `ori-amizur/introspector` with `openshift/assisted-installer-agent` and run `go mod tidy` to clean up unneeded dependencies.


empovit:
Converted to draft until test-infra passes

PR #4, opened 2020-08-12
MGMT-1696 Support customizable HTTPS transport

empovit:


src/inventory_client/inventory_client.go
```diff
@@ -41,20 +46,64 @@ type EnabledHostData struct {
...
 	if err != nil {
 		return nil, err
 	}
 
-	clientConfig.Transport = requestid.Transport(http.DefaultTransport)
+	var certs *x509.CertPool
+	if insecure {
+		logger.Warn("Certificate verification is turned off. This is not recommended in production environments")
+	} else {
+		certs, err = readCACertificate(caPath)
```
romfreiman, line 61: why is it required? If it's CA authority u dont need cret, right?
I mean while working with api.openshift u will pass nothing

src/inventory_client/inventory_client.go
```diff
@@ -41,20 +46,64 @@ type EnabledHostData struct {
...
+	caData, err := ioutil.ReadFile(capath)
+	if err != nil {
+		return nil, err
+	}
+
+	pool := x509.NewCertPool()
+	if !pool.AppendCertsFromPEM(caData) {
+		return nil, fmt.Errorf("failed to load certificate: %s", capath)
+	}
+
```
romfreiman, line 103: please log that u read it successfully

src/inventory_client/inventory_client.go
empovit: If it's a valid public CA you won't use this option. It's only if you use a self-signed CA and still want verification (most on-premised environments to my knowledge).

src/inventory_client/inventory_client.go
empovit: Hmm, it's a normal flow. Do you think logging it with `debug` could be useful? Or with `info`?

src/inventory_client/inventory_client.go
romfreiman: right... but...

src/inventory_client/inventory_client.go
romfreiman: it happens once. so info is good

src/inventory_client/inventory_client.go
```diff
@@ -41,20 +46,66 @@ type EnabledHostData struct {
...
 	return &inventoryClient{logger, assistedInstallClient, strfmt.UUID(clusterId)}, nil
 }
 
+func readCACertificate(capath string, logger *logrus.Logger) (*x509.CertPool, error) {
+
+	if capath == "" {
+		return nil, nil
+	}
+
+	caData, err := ioutil.ReadFile(capath)
```
tsorya, line 94: certificate is file on filesystem? how it is mounted?

empovit: Not mounted yet. Will be done at a later stage. Currently we'll support only public certificates or `--insecure`. I considered other options for loading certificates, but this one seems to give most flexibility without changing code. AFAIK there's no problem to send and mount a file in ignition.

eranco74: I assume that in case the install command will have the `--cacert` option it will also have the relevant volume mount.

src/inventory_client/inventory_client.go
```diff
@@ -41,20 +46,66 @@ type EnabledHostData struct {
...
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+		}).DialContext,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: insecure,
```
eranco74, line 79: What is the current behavior?
By default InsecureSkipVerify is set to true or false?

empovit: False, i.e. secure by default. Meaning it will fail with a custom CA or incorrect SAN when trying HTTPS.

PR #5, opened 2020-08-13
Igal/upload logs

tsorya:


src/ops/ops.go
```diff
@@ -347,3 +348,12 @@ func (o *ops) GetMCSLogs() (string, error) {
 
 	return string(logs), nil
 }
+
+func (o *ops) UploadInstallationLogs() (string, error) {
+	command := "podman"
+	args := []string{"run", "--rm", "--privileged", "-v", "/run/systemd/journal/socket:/run/systemd/journal/socket",
+		"-v", "/var/log:/var/log", "docker.io/tsorya/logs_sender:latest", "logs_sender", "-tag", "agent", "-tag", "installer",
```
romfreiman, line 355: nice image :)

tsorya, line 355: fix it

romfreiman: rega, so we have to build this one as well? where?

tsorya: nope it was bug

PR #12, opened 2020-08-23
MGMT-1626  -support v4.6 in assisted-installer

yevgeny-shnaidman:


src/utils/utils.go
```diff
@@ -11,8 +11,8 @@ import (
 
 	"github.com/openshift/assisted-installer-agent/pkg/journalLogger"
 
-	ignition "github.com/coreos/ignition/config/v2_2"
-	"github.com/coreos/ignition/config/v2_2/types"
+	ignition "github.com/coreos/ignition/v2/config/v3_1"
```
eranco74, line 14: This break current (4.5) installation right?

src/utils/utils.go
```diff
@@ -78,18 +78,18 @@ func SetFileInIgnition(ignitionData []byte, filePath, fileContents string, mode
...
+	rootUser := "root"
 	file := types.File{
 		Node: types.Node{
-			Filesystem: "root",
-			Path:       filePath,
-			Overwrite:  nil,
-			Group:      nil,
-			User:       nil,
+			Path:      filePath,
+			Overwrite: nil,
```
eranco74, line 85: @tsorya , @yevgeny-shnaidman what do you think about moving this to the ignition generator?
We do similar things there

src/utils/utils.go
yevgeny-shnaidman: right, but we won't support it once we move to 4.6

src/utils/utils.go
yevgeny-shnaidman: good idea, we can do it once @mhrivnak code is delivered, and we can gather all related ignition code at one place

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -188,6 +191,77 @@ func (c controller) PostInstallConfigs(wg *sync.WaitGroup) {
 	c.sendCompleteInstallation(true, "")
 }
 
+func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
+	defer wg.Done()
+	for {
+		time.Sleep(GeneralWaitTimeout)
+		exists, err := c.kc.IsMetalProvisioningExists()
```
tsorya, line 198: it needs to be check more than once?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -188,6 +191,77 @@ func (c controller) PostInstallConfigs(wg *sync.WaitGroup) {
...
+		if err == nil && exists {
+			c.log.Infof("Provisioning CR exists, no need to update BMHs")
+			return
+		}
+
+		bmhs, err := c.kc.ListBMHs()
+		if err != nil {
+			c.log.WithError(err).Errorf("Failed to BMH hosts")
```
tsorya, line 206: failed to list?

tsorya: sure it worth error? maybe warning is enough?

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -50,6 +50,7 @@ func main() {
 	wg.Add(2)
 	go assistedController.ApproveCsrs(done, &wg)
 	go assistedController.PostInstallConfigs(&wg)
+	go assistedController.UpdateBMHs(&wg)
```
tsorya, line 53: don't we need to delete 	go assistedController.ApproveCsrs(done, &wg)
 ?

src/utils/openshift_version.go
```diff
@@ -15,6 +15,11 @@ var getOpenshiftMapping = func(openshiftVersion string) map[string]string {
 		"4.5": {
 			rhcos: "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.5/4.5.1/rhcos-4.5.1-x86_64-metal.x86_64.raw.gz",       // 4.5.1 image in the openshift mirror
 			mco:   "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:58bed0f5c4fbe453b994f6b606fefe3e2aaf1d10dcbdf5debb73b93007c7bee5"}, // MCO from latest 4.5 nightly https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview/4.5.0-0.nightly-2020-05-14-021132/
+		"4.6": {
+			// rhcos: "https://releases-art-rhcos.svc.ci.openshift.org/art/storage/releases/rhcos-4.6/46.82.202007212240-0/x86_64/rhcos-46.82.202007212240-0-metal.x86_64.raw.gz",
```
tsorya, line 19: why we leave comments?

src/utils/utils.go
```diff
@@ -62,7 +62,7 @@ func GetFileContentFromIgnition(ignitionData []byte, fileName string) ([]byte, e
 	}
 	for i := range bm.Storage.Files {
 		if bm.Storage.Files[i].Path == fileName {
-			pullSecret, err := dataurl.DecodeString(bm.Storage.Files[i].Contents.Source)
+			pullSecret, err := dataurl.DecodeString(*bm.Storage.Files[i].Contents.Source)
```
tsorya, line 65: do we still need pull secret ? @eranco74 ?

src/utils/utils.go
tsorya: agree

deploy/assisted-installer-controller/assisted-installer-controller-role.yaml
```diff
@@ -70,8 +70,33 @@ rules:
...
     resources:
       - etcds
     verbs:
       - patch
+  - apiGroups:
+      - metal3.io
+    resources:
+      - baremetalhosts
+    verbs:
+      - get
```
tsorya, line 83: learned "*" :) for all verbs

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -188,6 +191,77 @@ func (c controller) PostInstallConfigs(wg *sync.WaitGroup) {
...
+}
+
+func (c controller) updateBMHStatus(bmhList metal3v1alpha1.BareMetalHostList) bool {
+	allUpdated := true
+	for _, bmh := range bmhList.Items {
+		c.log.Infof("Checking bmh %s", bmh.Name)
+		annotations := bmh.GetAnnotations()
+		content := []byte(annotations[metal3v1alpha1.StatusAnnotation])
+		if annotations[metal3v1alpha1.StatusAnnotation] == "" {
+			c.log.Infof("Skipping setting status of BMH host %s, status annotation not present", bmh.Name)
```
tsorya, line 225: if we don't have annotations we don't need to update?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -188,6 +191,77 @@ func (c controller) PostInstallConfigs(wg *sync.WaitGroup) {
...
+
+		allUpdated := c.updateBMHStatus(bmhs)
+		if allUpdated {
+			c.log.Infof("Updated all the BMH CRs, finished successfully")
+			return
+		}
+	}
+}
+
+func (c controller) updateBMHStatus(bmhList metal3v1alpha1.BareMetalHostList) bool {
```
tsorya, line 218: No one is updating this status? in parallel ?

src/assisted_installer_controller/assisted_installer_controller.go
yevgeny-shnaidman: i prefer to check it every loop, since in case there is a network gleech , i won't accidentally start updating BMHs.

src/assisted_installer_controller/assisted_installer_controller.go
yevgeny-shnaidman: error is more visible in the logs, in this case i preferred error, since controller is short-lived

src/main/assisted-installer-controller/assisted_installer_main.go
yevgeny-shnaidman: i thought to delete it, but now there is a talk of maybe removing BMHs at all for 4.6 ( in a later commits), so i will leave it here for now

src/assisted_installer_controller/assisted_installer_controller.go
yevgeny-shnaidman: no. This code is called in case provisioning for metal3 is not defined, in whch case BMO is not running, and it is the code that updates the statuses

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: u will not have lots of errors till masters join?

src/assisted_installer_controller/assisted_installer_controller.go
yevgeny-shnaidman: fixed

src/assisted_installer_controller/assisted_installer_controller.go
yevgeny-shnaidman: as long as we can connect to any api server (master or bootstrap), we will be able to list.

PR #19, opened 2020-08-31
Use IntVar for installation timeout

eranco74:


filanov (changes requested):
assisted-service pass duration 
please take a look at https://github.com/openshift/assisted-installer/pull/20

PR #20, opened 2020-08-31
Fix timeout duration - user pass a timeout duration and not int number

filanov:


filanov:
Closing PR because @eranco74 want to use int

PR #21, opened 2020-09-01
Move installation timeout to be unsigned integer, Verify integer over…

YuviGold:
…flow


romfreiman:
How do we verify it works?

eranco74 (commented):
/lgtm

YuviGold:
> How do we verify it works?

It already supported an integer. I just changed it to an unsigned integer. I added a test demonstrating an integer overflow and next time it will results a proper error.

PR #23, opened 2020-09-02
dummy push for DS image build

gamli75:
@eranco74 @filanov please approve


eranco74:
OK, but please fix what need to be fix so we don't push this kind of commits in the future

/lgtm

PR #26, opened 2020-09-07
MGMT-2030 Add Red Hat CA Cert to ignition if present

rollandf:
In case that the Red Hat IT root CA cert is present in the live ISO, add it to the node ignition file.
This allows container images from brew to be pulled. It is needed for the assisted-controller image.

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/openshift/assisted-installer/src/common"
 
+	"net/url"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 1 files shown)


rollandf:
@eranco74 @gamli75

romfreiman:
https://github.com/openshift/assisted-service/blob/01d9152b460a5e800b86c1e5fbc1bbf4ee43eda3/internal/installcfg/installcfg.go#L73 this should do the job. Not sure u need this PR

romfreiman:
Also, it will be overwritten in on-prem exactly due to the same parameter.

rollandf:
> https://github.com/openshift/assisted-service/blob/01d9152b460a5e800b86c1e5fbc1bbf4ee43eda3/internal/installcfg/installcfg.go#L73 this should do the job. Not sure u need this PR

without the controller did succeed in pulling the image.

romfreiman:
I really dont like the fact that we keep on using the pointer ignition from within the node. I assume that @carbonin will have to rewrite all this logic here MGMT-1989.
IMO the correct way to propagate this certificate to OCP is by using additionalTrustBundle param in install-config - then OCP will handle with it.
https://docs.openshift.com/container-platform/4.2/networking/configuring-a-custom-pki.html
https://coreos.slack.com/archives/C017MQVGY0L/p1599129815092500?thread_ts=1599127216.091600&cid=C017MQVGY0L

You can set it using the new API the @carbonin introduced

rollandf:
Closing, handling with InstallConfig

PR #30, opened 2020-09-13
Igal/MGMT-2145 Assisted-controller must not try to update already installed nodes or nodes in error

tsorya:


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -61,9 +61,16 @@ func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inv
 
 func (c *controller) WaitAndUpdateNodesStatus() {
 	c.log.Infof("Waiting till all nodes will join and update status to assisted installer")
-	assistedInstallerNodesMap := c.getInventoryNodesMap()
-	for len(assistedInstallerNodesMap) > 0 {
+	for {
 		time.Sleep(GeneralWaitTimeout)
+		assistedInstallerNodesMap, err := c.ic.GetHosts([]string{models.HostStatusDisabled,
```
eranco74, line 66: This is very misleading.
Consider declaring a var (Ignore/filterStatuses) with this statuses so it will be clear that you don't want these hosts.
Also why did you move into the for loop?

eranco74 (commented):
Looks good just a nit

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: cause worker can get error while we are running already. Looks like better to have updated status all the time

eranco74 (commented):
/lgtm

PR #31, opened 2020-09-14
Eran/mgmt 2116

eranco74:


src/inventory_client/inventory_client.go
```diff
@@ -80,6 +82,11 @@ func CreateInventoryClient(clusterId string, inventoryURL string, pullSecret str
 			RootCAs:            certs,
 		},
 	})
+	// Add retry settings
+	clientConfig.Transport = rehttp.NewTransport(transport,
+		rehttp.RetryMaxRetries(10),
+		rehttp.ExpJitterDelay(2, time.Duration(2)),
```
empovit, line 88: Capped at only 2 nanoseconds?

empovit: According to https://godoc.org/github.com/PuerkitoBio/rehttp#ExpJitterDelay, it can return a delay between 0 and 2 * 2 ^ 10 (2048) nanoseconds, but no more than 2 nanoseconds. Does it make sense?

src/inventory_client/inventory_client.go
```diff
@@ -80,6 +81,11 @@ func CreateInventoryClient(clusterId string, inventoryURL string, pullSecret str
 			RootCAs:            certs,
 		},
 	})
+	// Add retry settings
+	clientConfig.Transport = rehttp.NewTransport(transport,
+		rehttp.RetryMaxRetries(10),
+		rehttp.ExpJitterDelay(2, time.Duration(10)*time.Minute),
```
empovit, line 87: Formatting :)

src/inventory_client/inventory_client.go
```diff
@@ -80,6 +81,11 @@ func CreateInventoryClient(clusterId string, inventoryURL string, pullSecret str
 			RootCAs:            certs,
 		},
 	})
+	// Add retry settings
+	clientConfig.Transport = rehttp.NewTransport(transport,
```
tsorya, line 85: nice

src/inventory_client/inventory_client.go
empovit: 2048 nanoseconds (=2.048 microseconds=0.002048 ms) seem too short. Not comparable even to a typical round-trip time. Here I simulated a progression of delays with the values in your code.
```
attempt: 1, delay: 0s
attempt: 2, delay: 3ns
attempt: 3, delay: 0s
attempt: 4, delay: 9ns
attempt: 5, delay: 11ns
attempt: 6, delay: 79ns
attempt: 7, delay: 89ns
attempt: 8, delay: 453ns
attempt: 9, delay: 138ns
attempt: 10, delay: 1.672µs
```

src/inventory_client/inventory_client.go
eranco74: rehttp doesn't seem to work.:leftwards_arrow_with_hook: 
Wrote a simple retry roundtripper myself

src/inventory_client/retry_roundtripper.go
```diff
@@ -0,0 +1,48 @@
...
+	return rrt.retry(rrt.maxTries, b, rrt.Proxied.RoundTrip, req)
+
+}
+
+func (rrt RetryRoundTripper) retry(maxTries uint, backoff *backoff.Backoff, fn func(req *http.Request) (res *http.Response, e error), req *http.Request) (res *http.Response, err error) {
+	var i uint
+	for i = 0; i < maxTries; i++ {
+		res, err = fn(req)
+		if err != nil {
+			rrt.log.Warnf("Failed executing HTTP call: %s attempt number %d. Error: %s", req.Method, i, err)
```
empovit, line 37: IMO `Error: %e` would be better. Also, if you're including the method it's probably worth including the URL as well.
Attempt numbers will be printed zero-based, might be confusing if you see `attempt number 9` and don't realize it's the last of 10 attempts.

src/inventory_client/retry_roundtripper.go
```diff
@@ -0,0 +1,48 @@
...
+			if i < maxTries {
+				delay := backoff.Duration()
+				rrt.log.Warnf("Going to retry in: %s", delay)
```
filanov, line 40: delay.String()

src/inventory_client/retry_roundtripper.go
```diff
@@ -0,0 +1,48 @@
...
+		Min:    rrt.delay,
+		Max:    rrt.maxDelay,
+		Factor: 2,
+		Jitter: false,
+	}
+	return rrt.retry(rrt.maxTries, b, rrt.Proxied.RoundTrip, req)
+
+}
+
+func (rrt RetryRoundTripper) retry(maxTries uint, backoff *backoff.Backoff, fn func(req *http.Request) (res *http.Response, e error), req *http.Request) (res *http.Response, err error) {
```
tsorya, line 32: sure there is no package for it?

src/inventory_client/retry_roundtripper.go
```diff
@@ -0,0 +1,48 @@
...
+	var i uint
+	for i = 1; i <= maxTries; i++ {
+		res, err = fn(req)
+		if err != nil {
+			rrt.log.Warnf("Failed executing HTTP call: %s %s attempt number %d. Error: %s", req.Method, req.URL, i, err)
```
ori-amizur, line 37: Either use err.Error() or use log.WithError

ori-amizur (commented):
Maybe some tests?

src/inventory_client/retry_roundtripper.go
empovit, line 37: Suggestion: log this only if you're going to retry, i.e. inside `if i <= maxTries {`. Otherwise you may see the same error printed here and by the client code where it gets from `return res, err`.

PR #34, opened 2020-09-14
Fixed time.Duration lint failure

eranco74:
"/src/inventory_client/retry_roundtripper.go:23:11: cannot use rrt.delay (variable of type time.Duration) as time.Duration value in struct literal"

https://github.com/golangci/golangci-lint/issues/995


eranco74:
I'll run a few /retest to make sure

PR #35, opened 2020-09-14
Accept proxy settings in install command

masayag:
On D/S image, nsenter is being executed without environment variables,
therefore no proxy configuration are enabled to 'coreos-installer'
command which attempt to download a resource.
    
While the issue isn't happened to commands running on 'bash' image used
for U/S assisted-installer image, D/S image uses ubi8-minimal which
behaves differently and requires the env-vars to be provided explicitly
(assuming we'd like to avoid hacking the process id of the agent in
order to inherit the already-provided env-vars for proxy from it).
    
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1877757
Signed-off-by: Moti Asayag <masayag@redhat.com>

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -55,11 +55,24 @@ const (
 type ops struct {
 	log       *logrus.Logger
 	logWriter *utils.LogWriter
+	cmdEnv    []string
 }
 
 // NewOps return a new ops interface
 func NewOps(logger *logrus.Logger) Ops {
-	return &ops{logger, utils.NewLogWriter(logger)}
+	cmdEnv := os.Environ()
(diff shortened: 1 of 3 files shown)


src/ops/ops.go
```diff
@@ -82,6 +82,18 @@ func (o *ops) ExecCommand(liveLogger io.Writer, command string, args ...string)
 		cmd.Stdout = &stdoutBuf
 		cmd.Stderr = &stdoutBuf
 	}
+	if config.GlobalConfig.HTTPProxy != "" || config.GlobalConfig.HTTPSProxy != "" {
+		cmd.Env = os.Environ()
```
eranco74, line 86: Consider calculating the proxy env once in the ops builder and just appending it to the cmd.env here

PR #38, opened 2020-09-16
OCPBUGSM-17295 Set proxy env vars for controller

masayag:
The assisted-installer requires to use the same proxy settings as
specified by the user.
    
The assisted-installer-controller runs on OCP cluster and should use the
cluster-wide proxy, based on its processed status in order to assure
in-cluster communication is unproxified.

### DIFF
--- a/src/k8s_client/k8s_client.go
+++ b/src/k8s_client/k8s_client.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 
 	"github.com/openshift/assisted-installer/src/utils"
 	"k8s.io/apimachinery/pkg/labels"
@@ -17,6 +18,7 @@ import (
 
 	bmoapis "github.com/metal3-io/baremetal-operator/pkg/apis"
 	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	operatorv1 "github.com/openshift/client-go/operator/clientset/versioned"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -52,6 +54,7 @@ type K8SClient interface {
 	ListBMHs() (metal3v1alpha1.BareMetalHostList, error)
 	UpdateBMHStatus(bmh *metal3v1alpha1.BareMetalHost) error
 	UpdateBMH(bmh *metal3v1alpha1.BareMetalHost) error
+	SetEnvVars() error
 }
 
 type K8SClientBuilder func(configPath string, logger *logrus.Logger) (K8SClient, error)
@@ -62,7 +65,8 @@ type k8sClient struct {
 	ocClient      *operatorv1.Clientset
 	runtimeClient runtimeclient.Client
 	// CertificateSigningRequestInterface is interface
-	csrClient certificatesv1beta1client.CertificateSigningRequestInterface
+	csrClient   certificatesv1beta1client.CertificateSigningRequestInterface
+	proxyClient configv1client.ProxyInterface
 }
 
 func NewK8SClient(configPath string, logger *logrus.Logger) (K8SClient, error) {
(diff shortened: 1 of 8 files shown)


eranco74:
Unsure why this is required.
cluster proxy should handle it

src/ops/ops.go
```diff
@@ -59,9 +59,9 @@ type ops struct {
 }
 
 // NewOps return a new ops interface
-func NewOps(logger *logrus.Logger) Ops {
+func NewOps(logger *logrus.Logger, proxySet bool) Ops {
 	cmdEnv := os.Environ()
-	if config.GlobalConfig.HTTPProxy != "" || config.GlobalConfig.HTTPSProxy != "" {
+	if proxySet && (config.GlobalConfig.HTTPProxy != "" || config.GlobalConfig.HTTPSProxy != "") {
```
tsorya, line 64: maybe there is no need for config.GlobalConfig.HTTPProxy ?

src/k8s_client/k8s_client.go
```diff
@@ -184,6 +191,25 @@ func (c *k8sClient) GetConfigMap(namespace string, name string) (*v1.ConfigMap,
 	return cm, nil
 }
 
+func (c *k8sClient) SetEnvVars() error {
```
tsorya, line 194: i think better name is SetProxyEnvVars or something with proxy

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -28,20 +28,25 @@ func main() {
 
 	kc, err := k8s_client.NewK8SClient("", logger)
 	if err != nil {
-		log.Fatalf("Failed to create k8 client %e", err)
+		log.Fatalf("Failed to create k8 client %v", err)
```
tsorya, line 31: lets move it to log.WithError

src/ops/ops.go
masayag: it is needed for the assisted-installer only. but since this code is shared between the installer and the controller - only one requires it.

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -28,20 +28,25 @@ func main() {
...
 
+	err = kc.SetEnvVars()
+	if err != nil {
+		log.Fatalf("Failed to set env vars for installer-controller pod %v", err)
+	}
+
 	client, err := inventory_client.CreateInventoryClient(Options.ControllerConfig.ClusterID, Options.ControllerConfig.URL, Options.ControllerConfig.PullSecretToken, Options.ControllerConfig.SkipCertVerification, Options.ControllerConfig.CACertPath, logger)
 	if err != nil {
-		log.Fatalf("Failed to create inventory client %e", err)
+		log.Fatalf("Failed to create inventory client %v", err)
```
tsorya, line 44: lets move it to log.WithError

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -28,20 +28,25 @@ func main() {
...
-		log.Fatalf("Failed to create k8 client %e", err)
+		log.Fatalf("Failed to create k8 client %v", err)
 	}
 
 	logger.Infof("Start running assisted-installer with cluster-id %s, url %s",
 		Options.ControllerConfig.ClusterID, Options.ControllerConfig.URL)
 
+	err = kc.SetEnvVars()
+	if err != nil {
+		log.Fatalf("Failed to set env vars for installer-controller pod %v", err)
```
tsorya, line 39: lets move it to log.WithError

src/ops/ops.go
tsorya: got it , config.GlobalConfig.HTTPProxy  is for installer

tsorya, line 64: we must not see config.GlobalConfig.HTTPProxy in controller right? so why we need proxySet ?

src/main/assisted-installer-controller/assisted_installer_main.go
masayag: log doesn't have 'WithError', the assisted-service uses a different logging lib

eranco74 (commented):
/lgtm
/approve

src/ops/ops.go
masayag: since this code is used both by the installer and the controller  - we need to differentiate who invokes it.

for the controller - we are using the cluster-wide-proxy - by SetProxyEnvVars in its main. Then we don't need to set it any more, and mainly not to rely on env-vars for proxy other than the cluster ones.

the installer however, requires them in order to wrap the calls to 'nsenter' that doesn't inherit it from the env. This issue (nsenter fails to use env-vars) was detected only on d/s base image.

PR #40, opened 2020-09-17
WIP/On-Hold: Set etc/hosts file to resolve dnsname

pawanpinjarkar:
Assisted service sets a custom dns name for service base url for onprem use by adding all the ips of the assisted service node's into /etc/hosts file of the agent machines. Once the agent machine reboots during the cluster installation, the entries in etc/hosts file on agent nodes are lost. This fix adds the required entries back into the agent's etc/host file via ignition config.

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -3,6 +3,7 @@ package installer
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/openshift/assisted-installer/src/common"
 
@@ -426,5 +427,10 @@ func (i *installer) setEtcHostInIgnition(ignitionPath string) error {
 		i.log.Infof("No IPs to set, continuing")
 		return nil
 	}
-	return i.ops.SetFileInIgnition(ignitionPath, "/etc/hosts", fmt.Sprintf("data:,%s", dataurl.EncodeBytes([]byte(serviceIPs))), 420)
+	ips := strings.Split(serviceIPs, " ")
+	content := ""
+	for _, ip := range ips {
+		content = content + fmt.Sprintf(ip+" assisted-api.local.openshift.io\n")
+	}
+	return i.ops.SetFileInIgnition(ignitionPath, "/etc/hosts", dataurl.EncodeBytes([]byte(content)), 420)
 }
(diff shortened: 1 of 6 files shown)


src/installer/installer.go
```diff
@@ -412,3 +419,12 @@ func (i *installer) setHostnameInIgnition(ignitionPath string) error {
 	}
 	return i.ops.SetFileInIgnition(ignitionPath, "/etc/hostname", fmt.Sprintf("data:,%s", i.Hostname), 420)
 }
+
+func (i *installer) setEtcHostInIgnition(ignitionPath string) error {
```
romfreiman, line 423: @carbonin if I would you I would not allow this to be merged :)

carbonin:
re: https://github.com/openshift/assisted-installer/pull/40/files#r490447411

There is a plan in https://issues.redhat.com/browse/MGMT-2221 to move the host-side ignition customization to the assisted-service rather than making these changes on the node.

Where are these ips coming from? If you link to the source I'll make sure they're included in the ignition that will get downloaded.
Alternatively, if this is an urgent fix, we can make the change here first and I'll move this code over to the service in my refactoring.

carbonin:
Spoke with @pawanpinjarkar and we decided to move this to assisted-service.
1. This is not host-specific so this addition can, and should, be applied to both the master and worker ignition files in their current state
2. As @romfreiman mentioned, I'm going to be moving all of this ignition-munging logic into assisted-service soon so it would be wasted effort to make this change here then have to move it to assisted-service anyway.

The proposed change would be to build upon https://github.com/openshift/assisted-service/pull/185 by adding the `/etc/hosts` file to both `master.ign` and `worker.ign` after they are created, and before they are uploaded to S3. This can be done in parallel with the work I'm doing to move to host-based ignition files as the change would be applied identically, no matter which is merged first, the conflicts should be simple to resolve.

eranco74:
@pawanpinjarkar I'm closing this PR

pawanpinjarkar:
Created a new PR https://github.com/openshift/assisted-installer/pull/45 which is still needed after @carbonin completes his ignition changes.

PR #42, opened 2020-09-17
MGMT-2221 Replace master/worker ignition files with a separate one per host

carbonin:
This PR removes the ignition customization for configuring `/etc/hostname` from the assisted-installer and instead downloads an ignition file that will have the hostname file already configured.

This will allow us to consolidate user-provided ignition customizations (which will vary by host) with all the changes we make to the ignition files.


src/installer/installer.go
```diff
@@ -97,18 +97,12 @@ func (i *installer) InstallNode() error {
 	}
 
 	i.UpdateHostInstallProgress(models.HostStageInstalling, i.Config.Role)
-	ignitionFileName := i.Config.Role + ".ign"
+	ignitionFileName := i.Config.HostID + ".ign"
```
eranco74, line 100: Consider having the role in as part of the file name.

eranco74 (commented):
Looks good.

src/installer/installer.go
carbonin: I assume we don't have any plans to change the role or a host, right?

carbonin: Not an argument against the suggestion. Just curious as it might make this a bit strange.

eranco74:
@carbonin please rebase

src/installer/installer.go
```diff
@@ -324,7 +318,12 @@ func (i *installer) getInventoryHostsMap(hostsMap map[string]inventory_client.Ho
 			return nil, err
 		}
 		// no need for current host
-		delete(hostsMap, i.Hostname)
```
carbonin, line 327: This was added after my initial change to remove the `Hostname` field.

Generally I feel like this map should be keyed by id, but I don't think that kind of change is in the scope of this PR. Additionally I would rather not have to provide the hostname to the installer just to save a few iterations through this map.

eranco74:
@carbonin DO NOT MERGE: override assisted-service for tests?

carbonin:
Yeah, I needed the assisted service go client from my other PR so I added a replace line to go.mod to point this branch at my fork of assisted service.

PR #43, opened 2020-09-17
OCPBUGSM-17295 Implement Proxy function

masayag:
ProxyFromEnvVars provides an alternative to http.ProxyFromEnvironment
since it is being initialized only once and that happens by k8s before
proxy settings was obtained from the cluster.
While this is no issue for k8s client, it prevents any out-of-cluster
traffic from using the proxy.

By providing implementation of that method, we allow to use env-vars
set for the pod based on the cluster-wide-proxy.

Signed-off-by: Moti Asayag <masayag@redhat.com>

### DIFF
--- a/src/main/assisted-installer-controller/assisted_installer_main.go
+++ b/src/main/assisted-installer-controller/assisted_installer_main.go
@@ -2,10 +2,14 @@ package main
 
 import (
 	"log"
+	"net/http"
+	"net/url"
+	"os"
 	"sync"
 	"time"
 
 	"github.com/openshift/assisted-installer/src/k8s_client"
+	"golang.org/x/net/http/httpproxy"
 
 	"github.com/kelseyhightower/envconfig"
 	assistedinstallercontroller "github.com/openshift/assisted-installer/src/assisted_installer_controller"
@@ -17,6 +21,11 @@ import (
 var Options struct {
 	ControllerConfig assistedinstallercontroller.ControllerConfig
 }
+
+var (
+	envProxyOnce          sync.Once
+	envVarsProxyFuncValue func(*url.URL) (*url.URL, error)
+)
 
 func main() {
 	logger := logrus.New()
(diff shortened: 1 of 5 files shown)


src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -18,6 +22,11 @@ var Options struct {
 	ControllerConfig assistedinstallercontroller.ControllerConfig
 }
 
+var (
+	envProxyOnce          sync.Once
```
tsorya, line 26: why we need this const?

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -69,3 +80,26 @@ func main() {
 	logger.Infof("Waiting fo all go routines to finish")
 	wg.Wait()
 }
+
+// ProxyFromEnvVars provides an alternative to http.ProxyFromEnvironment since it is being initialized only
+// once and that happens by k8s before proxy settings was obtained. While this is no issue for k8s, it prevents
+// any out-of-cluster traffic from using the proxy
+func ProxyFromEnvVars(req *http.Request) (*url.URL, error) {
```
tsorya, line 87: not sure we need all 3 one line funcs? can't we move all to utils and call it CreateHttpProxyFunc .... ?All here is proxy oriented and no need to call one without calling others. So not sure what logic we split here?

src/main/assisted-installer-controller/assisted_installer_main.go
masayag: in order to evaluate the proxy configuration only once after they were set. the same approach is used by `http.ProxyFromEnvironment`

src/main/assisted-installer-controller/assisted_installer_main.go
masayag: That's because https://github.com/openshift/assisted-installer/blob/master/src/inventory_client/inventory_client.go#L76 expects to recieve a function as value for Proxy:
```
	transport := requestid.Transport(&http.Transport{
		Proxy: http.ProxyFromEnvironment, // this one is a function
		DialContext: (&net.Dialer{
...
``
it should be possible to inline one of the 3 methods.

PR #45, opened 2020-09-23
MGMT-1176: configure dns (etc/hosts) for controller pod for onprem

pawanpinjarkar:
Assisted service sets a custom DNS name for service base URL for on-prem use. This fix resolves the custom DNS name for the assisted installer controller pod only when running on-prem.

Original PR: https://github.com/openshift/assisted-installer/pull/40

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -243,7 +243,7 @@ func (o *ops) PrepareController() error {
 }
 
 func (o *ops) renderControllerCm() error {
-	var params = map[string]string{
+	var params = map[string]interface{}{
 		"InventoryUrl":         config.GlobalConfig.URL,
 		"ClusterId":            config.GlobalConfig.ClusterID,
 		"SkipCertVerification": strconv.FormatBool(config.GlobalConfig.SkipCertVerification),
@@ -255,7 +255,7 @@ func (o *ops) renderControllerCm() error {
 }
 
 func (o *ops) renderControllerSecret() error {
-	var params = map[string]string{
+	var params = map[string]interface{}{
 		"PullSecretToken": config.GlobalConfig.PullSecretToken,
 	}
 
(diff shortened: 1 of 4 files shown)


src/ops/ops.go
```diff
@@ -281,9 +281,23 @@ func (o *ops) renderDeploymentFiles(srcTemplate string, params map[string]string
 	o.log.Infof("Filling template file %s", srcTemplate)
 	tmpl := template.Must(template.New("assisted-controller").Parse(string(templateData)))
 	var buf bytes.Buffer
-	if err = tmpl.Execute(&buf, params); err != nil {
-		o.log.Errorf("Failed to render controller template: %e", err)
-		return err
+	var content []byte
+	if config.GlobalConfig.ServiceIPs != "" && strings.Contains(srcTemplate, controllerDeployPodTemplate) {
```
carbonin, line 285: I don't think this logic belongs here. This is a generic rendering function, can you move this to `renderControllerPod` as it will only ever apply to the pod case?

carbonin:
Also it would be great to have tests for the pod template generation now that we have different cases.

src/ops/ops.go
pawanpinjarkar: I had thought of doing that but the difference, in this case, is also for the type of params argument sent to `renderDeploymentFiles()` i.e. for onprem, param needs to be of type ` map[string][]string` while originally its ` map[string]string`Are you thinking of adding a separate method just for onprem similar to `renderDeploymentFiles`? I took this approach as then most of the code can be used as is.

carbonin: Can you change `params` to be `map[string]interface{}` in this function?

carbonin: Then I don't think you need to change this function at all, right?

carbonin: Well, except for the parameter type change :sweat_smile:

src/utils/utils.go
```diff
@@ -94,6 +95,14 @@ func SetFileInIgnition(ignitionData []byte, filePath, fileContents string, mode
 			Mode: &mode,
 		},
 	}
+	if config.GlobalConfig.ServiceIPs != "" && filePath == "/etc/hosts" {
```
carbonin, line 98: I thought we didn't need this as I'm removing this in #42

pawanpinjarkar: Ah, right. That's a leftover. Will remove. Thank you.

pawanpinjarkar:
Test cases for ops.go will be added as a separate PR. There are currently no tests for this file and it will need some code restructuring to write the tests.

carbonin (approved):

sacharya:
Closing this as I can't update this PR. See  https://github.com/openshift/assisted-installer/pull/61 instead.

PR #46, opened 2020-09-23
MGMT-1891 Configure CA certificate path for assisted controller

sacharya:
Add CA Cert to ignition file to be added to the host
Mount CA Cert host file as volume to the controller job

### DIFF
--- a/deploy/assisted-installer-controller/assisted-installer-controller-pod.yaml.template
+++ b/deploy/assisted-installer-controller/assisted-installer-controller-pod.yaml.template
@@ -48,6 +48,11 @@ spec:
           envFrom:
             - configMapRef:
                 name: assisted-installer-controller-config
+          {{if .CACertPath}}
+          volumeMounts:
+          - name: service-ca-cert-config
+            mountPath: {{.CACertPath}}
+          {{end}}
       restartPolicy: OnFailure
       nodeSelector:
         node-role.kubernetes.io/master: ""
@@ -56,3 +61,11 @@ spec:
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
         operator: Exists
+      {{if .CACertPath}}
+      volumes:
+      - name: service-ca-cert-config
+        hostPath:
+          path: {{.CACertPath}}
+      {{end}}
+
+
--- a/go.sum
+++ b/go.sum
@@ -78,6 +78,7 @@ github.com/Microsoft/go-winio v0.4.15-0.20190919025122-fc70bd9a86b5/go.mod h1:tT
 github.com/Microsoft/hcsshim v0.8.7/go.mod h1:OHd7sQqRFrYd3RmSgbgji+ctCwkbq2wbEYNSzOYtcBQ=
 github.com/NYTimes/gziphandler v0.0.0-20170623195520-56545f4a5d46/go.mod h1:3wb06e3pkSAbeQ52E9H9iFoQsEEwGN64994WTCIhntQ=
 github.com/NYTimes/gziphandler v1.1.1/go.mod h1:n/CVRwUEOgIxrgPvAQhUUr9oeUtvrhMomdKFjzJNB0c=
+github.com/Nvveen/Gotty v0.0.0-20120604004816-cd527374f1e5 h1:TngWCqHvy9oXAN6lEVMRuU21PR1EtLVZJmdB18Gu3Rw=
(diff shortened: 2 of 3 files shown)


deploy/assisted-installer-controller/assisted-installer-controller-pod.yaml.template
```diff
@@ -13,6 +13,8 @@ spec:
         - name: assisted-installer-controller
           image: {{.ControllerImage}}
           imagePullPolicy: Always
+          securityContext:
+            privileged: true
```
mhrivnak, line 17: What necessitates this?

sacharya: I am placing the cert under the standard /etc/pki/ca-trust/source/anchors/ directory on the host. Looks like hostpathing /etc/pki directory needs privileged access. 
Since we are not running `update-ca-certificates` command, it actually doesn't have to be /etc/pki. So may be I can place the cert under /etc/assisted-service. That would remove the need for privileged access.
And this is running as a job in the newly created OpenShift cluster. Is hostPath available in all OpenShift clusters?

sacharya: Changed the cert path on the node to /etc/assisted-service/service-ca-cert.crt and removed the privileged access.

go.sum
```diff
@@ -198,6 +199,7 @@ github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e h1:Wf6HqHfScWJN9
 github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
 github.com/coreos/go-systemd/v22 v22.0.0 h1:XJIw/+VlJ+87J+doOxznsAWIdmWuViOVhkQamW5YV28=
 github.com/coreos/go-systemd/v22 v22.0.0/go.mod h1:xO0FLkIi5MaZafQlIrOotqXZ90ih+1atmu1JpKERPPk=
+github.com/coreos/ignition v0.35.0 h1:UFodoYq1mOPrbEjtxIsZbThcDyQwAI1owczRDqWmKkQ=
```
empovit, line 202: Why would you need an older ignition version?

carbonin (changes requested):
I don't think you should be editing the ignition here.
The CA cert is shared across all nodes so you can make one edit to the ignition before it is uploaded to s3.
I think the only reason you would need to make a change here would be if your edit was specific to the host, but this is not.

You will still need the pod volume here, but all the ignition editing stuff should move to assisted-service.

deploy/assisted-installer-controller/assisted-installer-controller-pod.yaml.template
```diff
@@ -56,3 +61,11 @@ spec:
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
         operator: Exists
+      {{if .CACertPath}}
+      volumes:
+      - name: service-ca-cert-config
+        hostPath:
+          path: {{.CACertPath}}
```
carbonin, line 68: If this is going to be `/etc/assisted-service/<something>` should the path in the container be different than this one or will that path work somehow?

sacharya: The env variable CA_CERT_PATH is set to /etc/assisted-service/service-ca-cert.crt. And this just hostPaths the volume /etc/assisted-service/service-ca-cert.crt:/etc/assisted-service/service-ca-cert.crt to the container.

carbonin: Yes, I understand that. My question is if `/etc/assisted-service/service-ca-cert.crt` is the correct path to use in the container or should it be placed somewhere else?

sacharya: /etc/pki could be the obvious one. But needs privileged access for the hostpath to be shared, which is really unnecessary. Outside of that, /etc/assisted-service sounds like a reasonable place. Could be any path really.

carbonin: :+1: didn't see this in the env list

carbonin (approved):

PR #48, opened 2020-09-29
MGMT-2286 Continue on WaitAndUpdateNodesStatus error

YuviGold:


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -130,6 +131,7 @@ func (c *controller) ApproveCsrs(done <-chan bool, wg *sync.WaitGroup) {
 	for {
 		select {
 		case <-done:
+			c.log.Infof("Finish approving csrs")
```
romfreiman, line 134: CSRs

PR #50, opened 2020-09-30
Igal/controller logs

tsorya:


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -21,9 +26,11 @@ import (
 
 const (
 	generalWaitTimeoutInt = 30
+	logsSinceInSec        = 30 * 60
```
YuviGold, line 29: Bad name
Use time.Second

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -21,9 +26,11 @@ import (
...
 )
 
 var GeneralWaitTimeout = generalWaitTimeoutInt * time.Second
+var LogsTimeout = 5 * time.Minute
```
YuviGold, line 33: `LogsUploadTimeout`

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -325,3 +333,76 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
...
+	if err != nil {
+		return errors.Wrapf(err, "Failed to upload logs")
+	}
+	return nil
+}
+
+// Uploading logs every 5 minutes
+// We will take logs of assisted controller and upload them to assisted-service
+// by creating tar gz of them.
+func (c *controller) UploadControllerLogs(ctx context.Context, wg *sync.WaitGroup) {
```
YuviGold, line 372: Make a generic function getting job-name as a variable
Fix the logs accordingly.
```Failed to get %s..```
```Didn't find %s...```

File: src/assisted_installer_controller/assisted_installer_controller.go [hunk 4]
```diff
@@ -325,3 +333,76 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
...
+			return
+		case <-ticker.C:
+			if podName == "" {
+				pods, err := c.kc.GetPods(c.Namespace, map[string]string{"job-name": "assisted-installer-controller"})
+				if err != nil {
+					c.log.WithError(err).Warnf("Failed to get controller pod name")
+					continue
+				}
+				if len(pods) < 1 {
+					c.log.Infof("Didn't find myself, something strange had happened")
```
YuviGold, line 395: Is that possible? Can it be changed when running in a loop? Shouldn't be outside of the loop?

src/utils/utils.go
```diff
@@ -174,3 +177,27 @@ func GetHostIpsFromInventory(inventory *models.Inventory) ([]string, error) {
 	}
 	return ips, nil
 }
+
+func AddStringToTarGz(w io.Writer, data string, fileName string) error {
```
YuviGold, line 181: `WriteCompressedFileToPipe`

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -325,3 +333,76 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
...
+	}
+	if podLogs == "" {
+		return errors.Errorf("no logs of %s were found", podName)
+	}
+	pr, pw := io.Pipe()
+	defer pr.Close()
+
+	go func() {
+		defer pw.Close()
+		err = utils.AddStringToTarGz(pw, podLogs, fmt.Sprintf("%s.logs", podName))
```
YuviGold, line 355: I don't understand why do you complicate yourself with go routines/pipe.. This operation is synchronous
1. Creating compressed file -> 2. uploading it

YuviGold (changes requested):

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: why to? it is not duration it is int

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: cause i don't have file.
i am running it as stream. I don't want another object

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: It can't be generic. It is relevant to specific pod. I know that in assisted-installer namespace we have only one.

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: it only means that for some strange reason there is no pod.
just didn't want to fail here on panic while trying to get pod[0]

src/utils/utils.go
tsorya: it is not pipe. it is any writer. i like this name.

tsorya: it describes what it does. writes string to tar gz. can change add to write

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: will be glad for better name

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: to have generic function that does getPods? i have it in client, i still will need to check it's error and to verify i got pods at all.

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: that's why i prefer to right it to buffer and to read it from the same buffer

tsorya: and i don't think it is complicated. It is the way pipe is used. Even standard way

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -325,3 +334,73 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
...
+		return errors.Wrapf(err, "Failed to get logs of pod %s", podName)
+	}
+	pr, pw := io.Pipe()
+	defer pr.Close()
+
+	go func() {
+		defer pw.Close()
+		err = utils.WriteToTarGz(pw, podLogs, int64(podLogs.Len()), fmt.Sprintf("%s.logs", podName))
+		if err != nil {
+			c.log.WithError(err).Warnf("Failed to upload")
```
YuviGold, line 355: Failed to write tar gz

YuviGold (approved):
LGTM

PR #51, opened 2020-09-30
Use gotestsum tool for testing and publish junit on Jenkins

YuviGold:


eranco74:
@YuviGold we no longer have PR checks?

YuviGold:
> @YuviGold we no longer have PR checks?

We started PR checks as a pilot only on assisted-service :)

PR #54, opened 2020-10-04
MGMT-2349 Wrong retry request with invalid content length

eranco74:
Use rehttp for assisted-installer client retries


eranco74:
I'll appreciate it suggestions on how to verify the retry 
Currently the only validation we have that verify it works is setting the expect in a goroutine that sleep for 2 seconds (nice one @YuviGold :wink: )

YuviGold:
> I'll appreciate it suggestions on how to verify the retry
> Currently the only validation we have that verify it works is setting the expect in a goroutine that sleep for 2 seconds (nice one @YuviGold  )

We need to find out what the current test scenarios didn't cover.

eranco74:
> > I'll appreciate it suggestions on how to verify the retry
> > Currently the only validation we have that verify it works is setting the expect in a goroutine that sleep for 2 seconds (nice one @YuviGold  )
> 
> We need to find out what the current test scenarios didn't cover.

It didn't cover the actual request & budy the client is sending.
We just validate the assisted-service client mock was called.

YuviGold:
> > > I'll appreciate it suggestions on how to verify the retry
> > > Currently the only validation we have that verify it works is setting the expect in a goroutine that sleep for 2 seconds (nice one @YuviGold  )
> > 
> > 
> > We need to find out what the current test scenarios didn't cover.
> 
> It didn't cover the actual request & budy the client is sending.
> We just validate the assisted-service client mock was called.

So let's add this check to this PR

filanov:
@yuvalgalanti @eranco74 if you don't have subsystem them maybe creating a dummy service will help?
something like https://github.com/openshift/assisted-service/blob/master/pkg/auth/auth_handler_test.go#L151

YuviGold (approved):

YuviGold (approved):

PR #58, opened 2020-10-07
OCPBUGSM-18103 waiting for cluster version to be available

tsorya:
refactoring controller code and adding 2 hours timeout for post installed operations

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -8,6 +8,8 @@ import (
 	"sync"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
+
 	"github.com/openshift/assisted-installer/src/common"
 	"github.com/openshift/assisted-installer/src/utils"
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -267,64 +299,72 @@ func (c controller) unmarshalStatusAnnotation(content []byte) (*metal3v1alpha1.B
...
-				c.log.Infof("Found running console pod")
-				return
-			}
-			c.log.Infof("Console pod is in status %s. Continue waiting for it", pod.Status.Phase)
+		c.log.Infof("Console pod is in status %s. Continue waiting for it", pod.Status.Phase)
+	}
+	return false
+}
+
+func (c controller) validateClusterVersion() bool {
```
eranco74, line 351: Make sure this works and come back with OperatorAvailable==True within the timeout.
Note that this might effect our success rate, any idea if the openshift installer check this as well?

tsorya: sure it is, thats why i added this

PR #62, opened 2020-10-12
MGMT-2417 waiting for controller pod to start running

tsorya:


src/common/common.go
```diff
@@ -42,3 +44,20 @@ func SetConfiguringStatusForHosts(client inventory_client.InventoryClient, inven
...
 	}
 }
+
+func IsPodInStatus(k8Client k8s_client.K8SClient, podNamePrefix string, namespace string, labelMatch map[string]string, status v1.PodPhase, log logrus.FieldLogger) bool {
+	pods, err := k8Client.GetPods(namespace, labelMatch)
+	if err != nil {
+		log.WithError(err).Warnf("Failed to get pod with prefix %s in namespace %s", podNamePrefix, namespace)
+		return false
+	}
+	for _, pod := range pods {
```
eranco74, line 54: You expect more than 1?

src/installer/installer.go
```diff
@@ -276,6 +284,30 @@ func (i *installer) waitForBootkube(ctx context.Context) {
...
+		return err
+	}
+	kc, err := i.kcBuilder(KubeconfigPath, i.log)
+	if err != nil {
+		i.log.WithError(err).Errorf("Failed to create kc client from %s", KubeconfigPath)
+		return err
+	}
+	controllerTimeout := time.Duration(i.Config.InstallationTimeout) * time.Minute
+	predicate := func() bool {
+		return common.IsPodInStatus(kc, assistedControllerPrefix, "assisted-installer",
```
eranco74, line 301: Move assisted-installer to const

src/installer/installer.go
```diff
@@ -276,6 +284,30 @@ func (i *installer) waitForBootkube(ctx context.Context) {
...
+	if err != nil {
+		i.log.WithError(err).Error("Failed to reload resolv.conf")
+		return err
+	}
+	kc, err := i.kcBuilder(KubeconfigPath, i.log)
+	if err != nil {
+		i.log.WithError(err).Errorf("Failed to create kc client from %s", KubeconfigPath)
+		return err
+	}
+	controllerTimeout := time.Duration(i.Config.InstallationTimeout) * time.Minute
```
eranco74, line 299: timeout seems too long,
How long do you think we should wait?

src/ops/ops.go
```diff
@@ -385,3 +386,26 @@ func (o *ops) UploadInstallationLogs(isBootstrap bool) (string, error) {
 	}
 	return o.ExecPrivilegeCommand(o.logWriter, command, args...)
 }
+
+func (o *ops) ReloadHostFile(filepath string) error {
```
eranco74, line 390: Add a comment about why this is required

src/common/common.go
```diff
@@ -42,3 +44,20 @@ func SetConfiguringStatusForHosts(client inventory_client.InventoryClient, inven
 		}
 	}
 }
+
+func IsPodInStatus(k8Client k8s_client.K8SClient, podNamePrefix string, namespace string, labelMatch map[string]string, status v1.PodPhase, log logrus.FieldLogger) bool {
+	pods, err := k8Client.GetPods(namespace, labelMatch)
+	if err != nil {
+		log.WithError(err).Warnf("Failed to get pod with prefix %s in namespace %s", podNamePrefix, namespace)
```
eranco74, line 51: This will spam the log with errors no?

src/common/common.go
tsorya: it is common , yes it is possible for sure

src/installer/installer.go
tsorya: i think we will fail on timeout in assisted-service earlier. Didn't thought it is really big issue. If everything is ok we will have it running before bootkube finishes

src/common/common.go
tsorya: possible but u don't want to see that u don't have a pod when u are trying to reach it?

tsorya: and for sure it must not happened

src/installer/installer.go
tsorya: but if u think we need a smaller one, lets change just through the number

src/common/common.go
eranco74: OK, consider logging all the pods you find with this selector just so we will see what is going on.

src/installer/installer.go
eranco74: If we expect it to run before bootkube finish, (and we call this function once bootkube is done) let's give it 5 minutes.

tsorya: sure, will change

src/common/common.go
tsorya: didn't wanted to spawn too much logs

eranco74 (commented):
/lgtm
Please move timeout and interval to consts next pr

PR #65, opened 2020-10-15
Igal/proxy

tsorya:


src/config/config.go
```diff
@@ -58,6 +60,9 @@ func ProcessArgs() {
 	flag.StringVar(&ret.ServiceIPs, "service-ips", "", "All IPs of assisted service node")
 	h := flag.Bool("help", false, "Help message")
 	flag.Parse()
+	if ret.NoProxy != "" {
```
masayag, line 63: Isn't it redundant if the no-proxy values are updated by the assisted-service which means when the installer-agent starts - it will run with no-proxy set properly with the new values being added by https://github.com/openshift/assisted-service/pull/567/files#diff-e5c144e5eab23eac9b5947751c077100cbafd0c6d2f61059920666f6513c8ec1R99 ?

tsorya: nice, question. I will check it. Sounds like you are right

tsorya: checked, we must have this set. The problem is that installer runs with env from agent and we can't set this no_proxy for agent cause domain can be changed. 
I can set it as env

tsorya: but this way is easier

masayag (approved):

PR #68, opened 2020-10-19
MGMT-2456 Stop timing out installation, timeout logic should run only…

eranco74:
… in the assisted-service

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -12,12 +12,10 @@ import (
 	"github.com/openshift/assisted-service/models"
 	"github.com/thoas/go-funk"
 
-	"github.com/pkg/errors"
-	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


src/installer/installer.go
```diff
@@ -320,42 +315,43 @@ func (i *installer) waitForController() error {
...
 		i.log.Infof("Found %d ready master nodes", len(readyMasters))
 		if len(readyMasters) >= minMasterNodes {
 			i.log.Infof("Waiting for master nodes - Done")
-			cancel()
+			return true
+		}
+		return false
+	}
+
+	for {
```
tsorya, line 343: maybe lets add waitForPredicateWithContext? i will need to another function for sure?

eranco74: In another PR

PR #69, opened 2020-10-19
MGMT-2477 Send controller logs from installer before exit after contr…

tsorya:
…oller is running

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -304,8 +304,20 @@ func (i *installer) waitForController() error {
 		return err
 	}
 	predicate := func() bool {
-		return common.IsPodInStatus(kc, assistedControllerPrefix, assistedControllerNamespace,
+		controllerPod := common.GetPodInStatus(kc, assistedControllerPrefix, assistedControllerNamespace,
 			map[string]string{"job-name": assistedControllerPrefix}, v1.PodRunning, i.log)
+		if controllerPod != nil {
+			// uploading logs here cause we will finish installing bootstrap in couple of seconds
+			// just didn't wanted to start handling this logic in other place
+			// controller must be running for at least couple of minutes before this code so it will cover 90% of the cases
+			// if we will see that it is not enough, we will change the place.
+			err = common.UploadPodLogs(kc, i.inventoryClient, i.ClusterID, controllerPod.Name, assistedControllerNamespace, common.ControllerLogsSecondsAgo, i.log)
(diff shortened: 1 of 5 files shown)


src/installer/installer.go
```diff
@@ -304,8 +304,16 @@ func (i *installer) waitForController() error {
...
 	}
 	predicate := func() bool {
-		return common.IsPodInStatus(kc, assistedControllerPrefix, assistedControllerNamespace,
+		controllerPod := common.GetPodInStatus(kc, assistedControllerPrefix, assistedControllerNamespace,
 			map[string]string{"job-name": assistedControllerPrefix}, v1.PodRunning, i.log)
+		if controllerPod != nil {
+			err = common.UploadPodLogs(kc, i.inventoryClient, i.ClusterID, controllerPod.Name, assistedControllerNamespace, common.ControllerLogsSecondsAgo, i.log)
+			// if failed to upload logs, log why and continue
+			if err != nil {
+				i.log.WithError(err).Warnf("Failed to upload controller logs")
```
eranco74, line 313: Note that in case of failure it will not continue in the loop but return true in the next line

src/installer/installer.go
```diff
@@ -304,8 +304,16 @@ func (i *installer) waitForController() error {
 		return err
 	}
 	predicate := func() bool {
-		return common.IsPodInStatus(kc, assistedControllerPrefix, assistedControllerNamespace,
+		controllerPod := common.GetPodInStatus(kc, assistedControllerPrefix, assistedControllerNamespace,
 			map[string]string{"job-name": assistedControllerPrefix}, v1.PodRunning, i.log)
+		if controllerPod != nil {
+			err = common.UploadPodLogs(kc, i.inventoryClient, i.ClusterID, controllerPod.Name, assistedControllerNamespace, common.ControllerLogsSecondsAgo, i.log)
```
eranco74, line 310: Unsure sending the logs here is the best option,
I think logs should be sent late as possible, right before restart.

tsorya: this is only controller logs, just to see what is going on in it before exit. Reboot will be set in ~ 5 seconds

src/installer/installer.go
tsorya: yap, done on purpose. Sort of best effort. Don't want to break installation on this case

eranco74 (commented):
/lgtm

PR #71, opened 2020-10-20
MGMT-2272: set the consumerRef in the BMH object

asalkeld:
This needs to be done here so that we can figure out the correct machine
name. As the names are auto generated and not know ahead of time.

This has been tested by starting a cluster with 2 nodes, and it does set the machine cosumerRef correctly.


romfreiman:
@eranco74 @yevgeny-shnaidman why we're doing it from the installer and not from the cloud?

yevgeny-shnaidman:
@romfreiman because the worker Machine CRs are not created by the openshift installer, but by the MAO when the workers are added to the cluster, and we need to update BMH CRs to point to the workers CRs

romfreiman:
and we have to execute it as a command per worker?

yevgeny-shnaidman:
what do you mean by "as command"? it will be executed by assisted-controller the same way as we update BMHs now, in loop

asalkeld:
@yevgeny-shnaidman @romfreiman ptal

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -247,6 +247,168 @@ func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
...
+
+		allUpdated := c.updateWorkerLinkage(bmhs, nodes, machines)
+		if allUpdated {
+			log.Infof("Updated all the worker links, finished successfully")
+			return
+		}
+	}
+}
+
+func (c controller) updateWorkerLinkage(bmhList metal3v1alpha1.BareMetalHostList, nodeList *v1.NodeList, machineList *mapiv1beta1.MachineList) bool {
```
yevgeny-shnaidman, line 291: a question regarding the flow: why do we need to find the node? Why not simply pick BMH without ConsumerRef, and then pick an available machine?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -247,6 +247,168 @@ func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
...
+		if a.Type == v1.NodeInternalIP {
+			internalAddress = a
+		}
+	}
+	if internalAddress.Address == "" {
+		return nil, fmt.Errorf("can not find internal address of node %v", node.Name)
+	}
+
+	var firstUnassociatedWorker *mapiv1beta1.Machine
+	for i, machine := range machineList.Items {
```
yevgeny-shnaidman, line 389: i think you should keep a list of  all the machines that were already mapped or remove a mapped machine from the list. Otherwise, in case Actuator has not succeeded to update Machine status ip address before your loop moves to the next BMH, you can pick the same machine.

src/assisted_installer_controller/assisted_installer_controller.go
asalkeld: OK, i'l give that ago. I was trying to set the internalIP and I needed the node. but now that i am not doing that, i can remove the node stuff.

src/utils/utils.go
```diff
@@ -134,6 +134,15 @@ func FindAndRemoveElementFromStringList(s []string, r string) []string {
 	return s
 }
 
+func ElementInStringList(list []string, word string) bool {
```
tsorya, line 137: i think we don't need this one, just use funk.Contains

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -246,6 +246,98 @@ func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
 	}
 }
 
+func (c controller) UpdateWorkerLinkage(wg *sync.WaitGroup) {
```
tsorya, line 249: can't we unite it with UpdateBMHs ? looks like most of the logic is the same

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -246,6 +246,98 @@ func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
...
+	return allUpdated
+}
+
+func (c controller) findUnassignedWorkerMachine(machineList *mapiv1beta1.MachineList, assigned []string) (*mapiv1beta1.Machine, error) {
+	for i, machine := range machineList.Items {
+		role, ok := machine.Labels["machine.openshift.io/cluster-api-machine-role"]
+		if !ok || role != "worker" {
+			continue
+		}
+		if !utils.ElementInStringList(assigned, machine.Name) {
```
tsorya, line 334: funk.Contains will do the trick

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: will be nice to have test for new functions

ronniel1 (commented):
@asalkeld we don't use vendoring 
Please don't add them to git

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -246,6 +246,98 @@ func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
...
+			Namespace:  machine.Namespace,
+			Name:       machine.Name,
+		}
+
+		log.Infof("updating bmh consumerRef")
+		err = c.kc.UpdateBMH(&bmh)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to remove status annotation from BMH %s", bmh.Name)
+		} else {
+			assignedMachines = append(assignedMachines, machine.Name)
```
yevgeny-shnaidman, line 322: I think that you can simply delete an Item from MachineList , and not to managed an additional list (assignedMachines).

eranco74:
@yevgeny-shnaidman , @asalkeld h will this work for day2 nodes?
The controller exit once the installation is completed.

eranco74:
/hold
@asalkeld please remove the vendors

yevgeny-shnaidman:
@eranco74 for day2 on OCP we will have a different image of controller deployed,one that is pod, not a job, and will run only part of the functionality that we need

asalkeld:
@eranco74 ptal

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -238,48 +238,91 @@ func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
...
 	allUpdated := true
+
 	for i := range bmhList.Items {
 		bmh := bmhList.Items[i]
 		c.log.Infof("Checking bmh %s", bmh.Name)
 		annotations := bmh.GetAnnotations()
 		content := []byte(annotations[metal3v1alpha1.StatusAnnotation])
-		if annotations[metal3v1alpha1.StatusAnnotation] == "" {
+
+		if annotations[metal3v1alpha1.StatusAnnotation] == "" && bmh.Spec.ConsumerRef != nil {
```
eranco74, line 286: Seems that the only way for this method to return true (allUpdated) is if all bmh will get true in this if.
Does that make sense?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -238,48 +238,91 @@ func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
...
 }
 
-func (c controller) updateBMHStatus(bmhList metal3v1alpha1.BareMetalHostList) bool {
+func (c controller) unallocatedMachines(bmhList metal3v1alpha1.BareMetalHostList) (*mapiv1beta1.MachineList, error) {
+	machineList, err := c.kc.ListMachines()
+	if err != nil {
+		return nil, err
+	}
+	// filter out machines that are not an allocatable worker
+	for m := 0; m < len(machineList.Items); m++ {
```
eranco74, line 261: https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating

src/assisted_installer_controller/assisted_installer_controller.go
asalkeld: Yes, I believe it is correct.
It scans through all bmh and makes sure that 
1. the status annotation has been moved
2. the consumerRef has been set

"1." is as before and "2." will already be true for masters, and we are setting for workers.
The function gets continuously called so even if we set them all in one run, it will be called once more for a check and then it will return true.

eranco74:
/lgtm
please add the relevant tests

PR #72, opened 2020-10-21
Don't assume docker in the makefile

asalkeld:
This copies the way the dev scripts handle docker/podman


eranco74 (commented):
/lgtm
/approve

PR #75, opened 2020-10-25
MGMT-2654 Disable wait for clusterversion operator availability

tsorya:


eranco74:
Why not simple revert?
Why do we need all these comments of dead code?

eranco74 (commented):
/lgtm

PR #79, opened 2020-10-30
MGMT-2481: Make errors from ExecCommand() less verbose by default

asalkeld:


YuviGold:
WDYT? @filanov

src/ops/ops.go
```diff
@@ -84,6 +83,33 @@ func (o *ops) ExecPrivilegeCommand(liveLogger io.Writer, command string, args ..
...
+}
+
+func (e *ExecCommandError) Error() string {
+	lastOutput := e.Output
+	if len(e.Output) > 200 {
+		lastOutput = "... " + e.Output[len(e.Output)-200:]
+	}
+
+	firstArgs := e.Args
+	if len(e.Args) > 3 {
```
YuviGold, line 102: I think that we always want to see all the arguments

src/ops/ops_test.go
```diff
@@ -0,0 +1,66 @@
...
+			},
+			wantError: `failed executing nsenter [-t 1 -m], Error exit status 255, LastOutput "... or: Timeout was reached
+F1019 19:04:35.869592 1 start.go:147] Failed to initialize single run daemon: error reading osImageURL from rpm-ostree: error running rpm-ostree status --json: : exit status 1)"`,
+			wantDetailedError: `failed executing nsenter [-t 1 -m -i -- podman run --net host --volume /:/rootfs:rw --volume /usr/bin/rpm-ostree:/usr/bin/rpm-ostree --privileged --entrypoint /usr/bin/machine-config-daemon quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:dc1a34f55c712b2b9c5e5a14dd85e67cbdae11fd147046ac2fef9eaf179ab221 start --node-name localhost --root-mount /rootfs --once-from /opt/install-dir/bootstrap.ign --skip-reboot], env vars [HOME=/home/userZ], error exit status 255, waitStatus 255, Output "Trying to pull quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:dc1a34f55c712b2b9c5e5a14dd85e67cbdae11fd147046ac2fef9eaf179ab221...
```
YuviGold, line 39: This error is too huge and kinda confusing
I would prefer an unreal simpler string, because we don't check actual data here, only the parsing.
e.g.
```
word word word word word word.....
====
Everything from here on should be removed
```

src/ops/ops.go
```diff
@@ -108,15 +134,20 @@ func (o *ops) ExecCommand(liveLogger io.Writer, command string, args ...string)
...
 			if status, ok := exitErr.Sys().(syscall.WaitStatus); ok {
-				o.log.Debugf("failed executing %s %v, env vars %v, out: %s, with status %d",
-					command, args, output, cmd.Env, status.ExitStatus())
-
-				return output, fmt.Errorf("failed executing %s %v, env vars %v, error %s. Output %s", command, args, cmd.Env, exitErr, output)
+				execErr.WaitStatus = status.ExitStatus()
 			}
 		}
-		return output, fmt.Errorf("failed executing %s %v, env vars %v, error %s. Output %s", command, args, cmd.Env, err, output)
+		o.log.Debugf(execErr.DetailedError())
```
filanov, line 149: Maybe it should be info?

filanov (commented):
In general i agree that event should not be long but it's only in case user have a way to see or get the full error.
Will it help use to get a json output? will it be clearer? 
@eranco74 ?

src/ops/ops_test.go
asalkeld: `wantDetailedError` is the full error that only goes into the logs
`wantError` is the error that goes into the event

src/ops/ops.go
```diff
@@ -83,6 +82,28 @@ func (o *ops) ExecPrivilegeCommand(liveLogger io.Writer, command string, args ..
 	return o.ExecCommand(liveLogger, commandBase, arguments...)
 }
 
+type ExecCommandError struct {
+	Command    string
+	Args       []string
+	Env        []string
+	ExitErr    error
+	Output     string
+	WaitStatus int
```
eranco74, line 91: What is WaitStatus? should be ExitStatus no?

YuviGold (approved):

eranco74 (commented):
> In general i agree that event should not be long but it's only in case user have a way to see or get the full error.
> Will it help use to get a json output? will it be clearer?
> @eranco74 ?

We discussed it in the [ticket](https://issues.redhat.com/browse/MGMT-2481).
We decided to go with: " some text that says what high-level command failed+ the interesting part of the output (why it failed) "

PR #80, opened 2020-11-01
MGMT-2678: Adding bootstrap public SSH key to masters ignition.

ybettan:
It will allows bootstrap SSH to masters in order to collect
`installer-gather.sh` logs in case of cluster installation failure.

Signed-off-by: Yoni Bettan <ybettan@redhat.com>

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -176,6 +176,16 @@ func (i *installer) startBootstrap() error {
 		return err
 	}
 
+	err = i.generateSshKeyPair()
+	if err != nil {
+		return err
+	}
+
+	err = i.ops.CreateOpenshiftSshManifest(assistedInstallerSshManifest, sshManifestTmpl, sshPubKeyPath)
+	if err != nil {
+		return err
+	}
+
 	// reload systemd configurations from filesystem and regenerate dependency trees
 	err = i.ops.SystemctlAction("daemon-reload")
 	if err != nil {
@@ -229,6 +239,19 @@ func (i *installer) extractIgnitionToFS(ignitionPath string) (err error) {
 	}
 	i.log.Errorf("Failed to extract ignition to disk, giving up")
 	return err
+}
+
+func (i *installer) generateSshKeyPair() error {
+	i.log.Info("Generating new SSH key pair")
+	if err := i.ops.Mkdir(sshDir); err != nil {
+		i.log.WithError(err).Error("Failed to create SSH dir")
+		return err
+	}
+	if _, err := i.ops.ExecPrivilegeCommand(utils.NewLogWriter(i.log), "ssh-keygen", "-q", "-f", sshKeyPath, "-N", ""); err != nil {
+		i.log.WithError(err).Error("Failed to generate SSH key pair")
+		return err
+	}
+	return nil
 }
 
 func (i *installer) getFileFromService(filename string) (string, error) {
--- a/src/installer/installer_test.go
+++ b/src/installer/installer_test.go
@@ -52,8 +52,8 @@ var _ = Describe("installer HostRoleMaster role", func() {
 	generalWaitTimeout = 100 * time.Minute
 	device := "/dev/vda"
 	l.SetOutput(ioutil.Discard)
-	mkdirSuccess := func() {
-		mockops.EXPECT().Mkdir(InstallDir).Return(nil).Times(1)
+	mkdirSuccess := func(filepath string) {
+		mockops.EXPECT().Mkdir(filepath).Return(nil).Times(1)
 	}
 	downloadFileSuccess := func(fileName string) {
 		mockbmclient.EXPECT().DownloadFile(gomock.Any(), fileName, filepath.Join(InstallDir, fileName)).Return(nil).Times(1)
@@ -117,6 +117,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 	}
 
 	Context("Bootstrap role", func() {
+
 		conf := config.Config{Role: string(models.HostRoleBootstrap),
 			ClusterID:        "cluster-id",
 			HostID:           "host-id",
@@ -179,13 +180,22 @@ var _ = Describe("installer HostRoleMaster role", func() {
 		extractSecretFromIgnitionSuccess := func() {
 			mockops.EXPECT().ExtractFromIgnition(filepath.Join(InstallDir, bootstrapIgn), dockerConfigFile).Return(nil).Times(1)
 		}
+		generateSshKeyPairSuccess := func() {
+			mkdirSuccess(sshDir)
+			mockops.EXPECT().ExecPrivilegeCommand(gomock.Any(), "ssh-keygen", "-q", "-f", sshKeyPath, "-N", "").Return("OK", nil).Times(1)
+		}
+		createOpenshiftSshManifestSuccess := func() {
+			mockops.EXPECT().CreateOpenshiftSshManifest(assistedInstallerSshManifest, sshManifestTmpl, sshPubKeyPath).Return(nil).Times(1)
+		}
 
 		bootstrapSetup := func() {
 			cleanInstallDevice()
-			mkdirSuccess()
+			mkdirSuccess(InstallDir)
 			downloadFileSuccess(bootstrapIgn)
 			extractSecretFromIgnitionSuccess()
-			extractIgnitionToFS("Sוuccess", nil)
+			extractIgnitionToFS("Success", nil)
+			generateSshKeyPairSuccess()
+			createOpenshiftSshManifestSuccess()
 			daemonReload(nil)
 		}
 
@@ -233,6 +243,26 @@ var _ = Describe("installer HostRoleMaster role", func() {
 			ret := installerObj.InstallNode()
 			Expect(ret).Should(Equal(err))
 		})
+		It("bootstrap role creating SSH manifest failed", func() {
+			updateProgressSuccess([][]string{{string(models.HostStageStartingInstallation), conf.Role},
+				{string(models.HostStageWaitingForControlPlane)},
+				{string(models.HostStageInstalling), string(models.HostRoleMaster)},
+				{string(models.HostStageWritingImageToDisk)},
+			})
+			cleanInstallDevice()
+			mkdirSuccess(InstallDir)
+			downloadFileSuccess(bootstrapIgn)
+			extractSecretFromIgnitionSuccess()
+			extractIgnitionToFS("Success", nil)
+			generateSshKeyPairSuccess()
+			err := fmt.Errorf("generate SSH keys failed")
+			mockops.EXPECT().CreateOpenshiftSshManifest(assistedInstallerSshManifest, sshManifestTmpl, sshPubKeyPath).Return(err).Times(1)
+			//HostRoleMaster flow:
+			downloadHostIgnitionSuccess(hostId, "master-host-id.ign")
+			writeToDiskSuccess()
+			ret := installerObj.InstallNode()
+			Expect(ret).To(HaveOccurred())
+		})
 		It("bootstrap role extract ignition retry", func() {
 			updateProgressSuccess([][]string{{string(models.HostStageStartingInstallation), conf.Role},
 				{string(models.HostStageWaitingForControlPlane)},
@@ -266,7 +296,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				{string(models.HostStageWaitingForControlPlane)},
 			})
 			cleanInstallDevice()
-			mkdirSuccess()
+			mkdirSuccess(InstallDir)
 			downloadFileSuccess(bootstrapIgn)
 			downloadHostIgnitionSuccess(hostId, "master-host-id.ign")
 			writeToDiskSuccess()
@@ -378,7 +408,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				{string(models.HostStageRebooting)},
 			})
 			cleanInstallDevice()
-			mkdirSuccess()
+			mkdirSuccess(InstallDir)
 			downloadHostIgnitionSuccess(hostId, "master-host-id.ign")
 			writeToDiskSuccess()
 			uploadLogsSuccess(false)
@@ -424,7 +454,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				{string(models.HostStageInstalling), conf.Role},
 			})
 			cleanInstallDevice()
-			mkdirSuccess()
+			mkdirSuccess(InstallDir)
 			err := fmt.Errorf("failed to fetch file")
 			mockbmclient.EXPECT().DownloadHostIgnition(gomock.Any(), hostId, filepath.Join(InstallDir, "master-host-id.ign")).Return(err).Times(1)
 			ret := installerObj.InstallNode()
@@ -436,7 +466,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				{string(models.HostStageWritingImageToDisk)},
 			})
 			cleanInstallDevice()
-			mkdirSuccess()
+			mkdirSuccess(InstallDir)
 			downloadHostIgnitionSuccess(hostId, "master-host-id.ign")
 			err := fmt.Errorf("failed to write image to disk")
 			mockops.EXPECT().WriteImageToDisk(filepath.Join(InstallDir, "master-host-id.ign"), device, mockbmclient, nil).Return(err).Times(3)
@@ -450,7 +480,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				{string(models.HostStageRebooting)},
 			})
 			cleanInstallDevice()
-			mkdirSuccess()
+			mkdirSuccess(InstallDir)
 			downloadHostIgnitionSuccess(hostId, "master-host-id.ign")
 			uploadLogsSuccess(false)
 			writeToDiskSuccess()
@@ -478,7 +508,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				{string(models.HostStageRebooting)},
 			})
 			cleanInstallDevice()
-			mkdirSuccess()
+			mkdirSuccess(InstallDir)
 			downloadHostIgnitionSuccess(hostId, "worker-host-id.ign")
 			mockops.EXPECT().WriteImageToDisk(filepath.Join(InstallDir, "worker-host-id.ign"), device, mockbmclient, nil).Return(nil).Times(1)
 			// failure must do nothing
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -4,14 +4,15 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
+	"strconv"
+	"text/template"
+
+	"io/ioutil"
 	"os/exec"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"syscall"
-	"text/template"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -39,6 +40,7 @@ type Ops interface {
 	GetMCSLogs() (string, error)
 	UploadInstallationLogs(isBootstrap bool) (string, error)
 	ReloadHostFile(filepath string) error
+	CreateOpenshiftSshManifest(filePath, template, sshPubKeyPath string) error
 }
 
 const (
@@ -433,6 +435,30 @@ func (o *ops) ReloadHostFile(filepath string) error {
 	_, err = f.WriteString(output)
 	if err != nil {
 		o.log.Errorf("Failed to write host %s data to local", filepath)
+		return err
+	}
+	return nil
+}
+
+func (o *ops) CreateOpenshiftSshManifest(filePath, tmpl, sshPubKeyPath string) error {
+	o.log.Info("Create an openshift manifets for SSH public key")
+	sshPublicKey, err := o.ExecPrivilegeCommand(o.logWriter, "cat", sshPubKeyPath)
+	if err != nil {
+		o.log.WithError(err).Errorf("Failed to read SSH pub key from %s", sshPubKeyPath)
+		return err
+	}
+	f, err := os.Create(filePath)
+	if err != nil {
+		o.log.WithError(err).Errorf("Failed to create %s", filePath)
+		return err
+	}
+	defer f.Close()
+	t := template.Must(template.New("openshift SSH manifest").Parse(tmpl))
+	sshConfig := struct {
+		SshPubKey string
+	}{sshPublicKey}
+	if err := t.Execute(f, sshConfig); err != nil {
+		o.log.WithError(err).Error("Failed to execute template")
 		return err
 	}
 	return nil
--- a/go.mod
+++ b/go.mod
@@ -21,11 +21,13 @@ require (
 	github.com/openshift/client-go v0.0.0-20200827190008-3062137373b5
 	github.com/openshift/machine-api-operator v0.2.1-0.20201002104344-6abfb5440597
 	github.com/pkg/errors v0.9.1
+	github.com/prometheus/common v0.10.0
 	github.com/sirupsen/logrus v1.7.0
 	github.com/thoas/go-funk v0.6.0
 	github.com/vincent-petithory/dataurl v0.0.0-20191104211930-d1553a71de50
 	golang.org/x/net v0.0.0-20201002202402-0a1ea396d57c
 	golang.org/x/sync v0.0.0-20201008141435-b3e1573b7520
+	gopkg.in/yaml.v2 v2.3.0
 	k8s.io/api v0.19.2
 	k8s.io/apimachinery v0.19.2
 	k8s.io/client-go v12.0.0+incompatible
--- a/go.sum
+++ b/go.sum
@@ -111,6 +111,7 @@ github.com/ajeddeloh/go-json v0.0.0-20170920214419-6a2fe990e083/go.mod h1:otnto4
 github.com/ajeddeloh/go-json v0.0.0-20200220154158-5ae607161559 h1:4SPQljF/GJ8Q+QlCWMWxRBepub4DresnOm4eI2ebFGc=
 github.com/ajeddeloh/go-json v0.0.0-20200220154158-5ae607161559/go.mod h1:otnto4/Icqn88WCcM4bhIJNSgsh9VLBuspyyCfvof9c=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
+github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 h1:JYp7IbQjafoB+tBA3gMyHYHrpOtNuDiK/uB5uXxq5wM=
 github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
 github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
@@ -1681,6 +1682,7 @@ google.golang.org/protobuf v1.23.1-0.20200526195155-81db48ad09cc/go.mod h1:EGpAD
 google.golang.org/protobuf v1.24.0 h1:UhZDfRO8JRQru4/+LlLE0BRKGF8L+PICnvYZmx/fEGA=
 google.golang.org/protobuf v1.24.0/go.mod h1:r/3tXBNzIEhYS9I1OUVjXDlt8tc493IdKGjtUeSXeh4=
 gopkg.in/airbrake/gobrake.v2 v2.0.9/go.mod h1:/h5ZAUhDkGaJfjzjKLSjv6zCL6O0LLBxU4K+aSYdM/U=
+gopkg.in/alecthomas/kingpin.v2 v2.2.6 h1:jMFz6MfLP0/4fUyZle81rXUoxOBFi19VUFKVDOQfozc=
 gopkg.in/alecthomas/kingpin.v2 v2.2.6/go.mod h1:FMv+mEhP44yOT+4EoQTLFTRgOQ1FBLkstjWtayDeSgw=
 gopkg.in/asn1-ber.v1 v1.0.0-20181015200546-f715ec2f112d/go.mod h1:cuepJuh7vyXfUyUwEgHQXw849cJrilpS5NeIjOWESAw=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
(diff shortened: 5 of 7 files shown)


src/types/machine_config.go
```diff
@@ -0,0 +1,46 @@
...
+
+type Spec struct {
+	Config          Config  `yaml:"config"`
+	Fips            bool    `yaml:"fips"`
+	KernelArguments *string `yaml:"kernelArguments"`
+	KernelType      *string `yaml:"kernelType"`
+	OsImageURL      *string `yaml:"osImageURL"`
+}
+
+type MachineConfig struct {
```
ErezAlster, line 41: I think that you can import it from bootkube instead of generating it by yourself

src/installer/installer.go
```diff
@@ -183,6 +186,20 @@ func (i *installer) startBootstrap() error {
...
 
+	var mc types.MachineConfig
+	yamlFile := "/opt/openshift/openshift/99_openshift-machineconfig_99-master-ssh.yaml"
+	fieldName := "Spec.Config.Passwd.Users[0].SshAuthorizedKeys"
+	newSshKey, err := ioutil.ReadFile("/etc/ssh/ssh_host_rsa_key.pub")
+	if err != nil {
+		i.log.Error(err)
+		return err
+	}
+	err = i.patchMastersIgnition(yamlFile, fieldName, string(newSshKey), "append-list", &mc)
```
romfreiman, line 197: what will happen if the user didn specify ssh key at all - it's not a mandatory field

src/types/machine_config.go
eranco74: Bootkube start the cluster-bootstrap-operator that just apply these yamls./
This yaml is generated by the openshift-installerm which isn't a nice dependency to have.

ybettan: I will need to check that.

src/installer/installer.go
ybettan: Umm that require some changes, I need to check if that change the yaml structure or just have an empty list (in the second case no changes will be needed)

src/installer/installer.go
```diff
@@ -234,6 +268,29 @@ func (i *installer) extractIgnitionToFS(ignitionPath string) (err error) {
 	return err
 }
 
+func (i *installer) patchMastersIgnition(yamlFile, fieldName, newData, op string, yamlStruct interface{}) error {
```
rwsu, line 271: I think this is sufficiently complex to warrant a test.

ybettan: Agree, will do it (this is why it is still [WIP])
:)

src/types/machine_config.go
carbonin: Should be able to get this from here https://github.com/openshift/machine-config-operator/blob/master/pkg/apis/machineconfiguration.openshift.io/v1/types.go

carbonin:
We have code to edit ignition files before they're uploaded to S3 on the service.
We also have ways to provide additional manifests from the service.

Why are we doing any of this on the installer?

carbonin:
To be more specific, I think you should add the ssh key file to the installer directory somewhere around [here](https://github.com/openshift/assisted-service/blob/be1bc532930e1d07bcf4a5f24936e2ed96776e80/internal/ignition/ignition.go#L113-L136) where we're already adding the user-provided manifests.

I suggested the same approach to @ori-amizur for MGMT-1868 to create the dhcp lease renewal config, I'd imagine the approach would be almost identical.

ybettan:
https://github.com/openshift/assisted-installer/pull/80#issuecomment-723088802
@carbonin 

The reason is related to the owner of the key.
The service can't generate the key pair because then there will be no way to send the private key to the bootstrap host.

The optimal solution will be to generate the key pair (or use existing one) in the bootstrap host, send the public key to service, let the service insert it to the second ignition of masters (the one that contains the IP of bootstrap).

According to @eranco74, this approach will stall the installation process and the gain from it don't worth it.

src/installer/installer.go
eranco74: I guess we can create our own manifest "99_openshift-0assisted-installer-ssh.yaml" and place it under /opt/openshift/openshift/

eranco74:
The issue is that we decide which node will be the bootstrap node once the installation is triggered.
So in order to avoid moving the decision to an earlier stage (will lead to more edge cases we will need to handle)
We also don't want to wait for a command that will fetch the bootstrap public key and add it as a manifest prior to ignition generation (will cost us 1+ minutes).
There is an option for getting all the hosts public key as part of their inventory and using it once we know who is the bootstrap.

I believe this effort started before we had a way to provide additional manifests and that's why we didn't go with it.
Now since we need this functionality ASAP, I think we should get this in, add the test coverage and then get it to work with what @carbonin suggested (Just so we avoid doing these things in the installer) 
 

I guess we could get the bootstrap

src/installer/installer.go
ybettan: This may be a problem, the manifest contain an `ignition version` for example, so if the version change (in `openshift/installer`) it will break.

Another option is to insert a dummy shh key in case the user didn't supplied one and to override it with the bootstrap key in `assisted-installer`.

ybettan: Updated PR.

src/types/machine_config.go
ybettan: Will do it it a different task, https://issues.redhat.com/browse/MGMT-2847

deploy/manifests/99_openshift-machineconfig_99-additional-master-ssh.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  creationTimestamp: null
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: 99-additional-master-ssh
```
eranco74, line 7: Consider renaming this to:
```suggestion
  name: 99-assisted-installer-master-ssh
```

src/installer/installer.go
```diff
@@ -176,6 +178,42 @@ func (i *installer) startBootstrap() error {
...
+	if err != nil {
+		i.log.Error(err)
+		return err
+	}
+	_, err = i.ops.ExecPrivilegeCommand(utils.NewLogWriter(i.log), "chmod", "400", "/root/.ssh/id_rsa")
+	if err != nil {
+		i.log.Error(err)
+		return err
+	}
+
```
eranco74, line 216: Move this to a method

src/installer/installer.go
```diff
@@ -176,6 +178,42 @@ func (i *installer) startBootstrap() error {
...
+	i.log.Infof("Patching masters ignition with bootstrap public SSH key")
+	var mc types.MachineConfig
+	yamlFile := "/opt/openshift/openshift/99_openshift-machineconfig_99-additional-master-ssh.yaml"
+	fieldName := "Spec.Config.Passwd.Users[0].SshAuthorizedKeys"
+	newSshKey, err := i.ops.ReadFile("/etc/ssh/ssh_host_rsa_key.pub")
+	if err != nil {
+		i.log.Error(err)
+		return err
+	}
+	_, err = i.ops.ExecCommand(utils.NewLogWriter(i.log), "cp", "/99_openshift-machineconfig_99-additional-master-ssh.yaml", "/opt/openshift/openshift/")
```
eranco74, line 190: Please use the var (yamlFile) here

src/installer/installer.go
```diff
@@ -176,6 +178,42 @@ func (i *installer) startBootstrap() error {
 		return err
 	}
 
+	i.log.Infof("Patching masters ignition with bootstrap public SSH key")
+	var mc types.MachineConfig
+	yamlFile := "/opt/openshift/openshift/99_openshift-machineconfig_99-additional-master-ssh.yaml"
```
eranco74, line 183: Rename to:
```suggestion
	assistedInstallerSshKeyManifest := "/opt/openshift/openshift/99_openshift-machineconfig_99-additional-master-ssh.yaml"
```

src/installer/installer.go
```diff
@@ -176,6 +178,42 @@ func (i *installer) startBootstrap() error {
...
+	fieldName := "Spec.Config.Passwd.Users[0].SshAuthorizedKeys"
+	newSshKey, err := i.ops.ReadFile("/etc/ssh/ssh_host_rsa_key.pub")
```
eranco74, line 185: sshPublicKey

src/installer/installer.go
```diff
@@ -176,6 +178,42 @@ func (i *installer) startBootstrap() error {
...
+	if err != nil {
+		i.log.Error(err)
+		return err
+	}
+	_, err = i.ops.ExecCommand(utils.NewLogWriter(i.log), "cp", "/99_openshift-machineconfig_99-additional-master-ssh.yaml", "/opt/openshift/openshift/")
+	if err != nil {
+		i.log.Error(err)
+		return err
+	}
+	err = i.patchMastersIgnition(yamlFile, fieldName, string(newSshKey), "put", &mc)
```
eranco74, line 195: This method isn't patching the master ignition it's updating the manifest.
Please rename

src/installer/installer.go
tsorya, line 183: consts

src/installer/installer.go
```diff
@@ -176,6 +178,42 @@ func (i *installer) startBootstrap() error {
 		return err
 	}
 
+	i.log.Infof("Patching masters ignition with bootstrap public SSH key")
+	var mc types.MachineConfig
+	yamlFile := "/opt/openshift/openshift/99_openshift-machineconfig_99-additional-master-ssh.yaml"
+	fieldName := "Spec.Config.Passwd.Users[0].SshAuthorizedKeys"
```
tsorya, line 184: same

src/installer/installer.go
tsorya, line 185: same with /etc/ssh/ssh_host_rsa_key.pub

src/installer/installer.go
```diff
@@ -176,6 +178,42 @@ func (i *installer) startBootstrap() error {
...
+	err = i.ops.Mkdir("/root/.ssh")
+	if err != nil {
+		i.log.Error(err)
+		return err
+	}
+	_, err = i.ops.ExecPrivilegeCommand(utils.NewLogWriter(i.log), "cp", "/etc/ssh/ssh_host_rsa_key", "/root/.ssh/id_rsa")
+	if err != nil {
+		i.log.Error(err)
+		return err
+	}
```
eranco74, line 210: I think that if you generate your own key pair this won't be required.
* by default it will get created under ~/.ssh with the proper name.
`chmod` won't be required as well

src/installer/installer.go
tsorya, line 190: why exec for copy file and not golang?

src/installer/installer.go
```diff
@@ -227,6 +265,29 @@ func (i *installer) extractIgnitionToFS(ignitionPath string) (err error) {
 	return err
 }
 
+func (i *installer) patchMastersIgnition(yamlFile, fieldName, newData, op string, yamlStruct interface{}) error {
```
eranco74, line 268: Not updating ignition, please rename

src/installer/installer.go
```diff
@@ -227,6 +265,29 @@ func (i *installer) extractIgnitionToFS(ignitionPath string) (err error) {
...
+	err := i.ops.ReadYaml(yamlFile, yamlStruct)
+	if err != nil {
+		i.log.WithError(err).Error("Failed to read yaml")
+		return err
+	}
+	data := utils.GetFieldFromFieldName(yamlStruct, fieldName)
+	switch op {
```
eranco74, line 275: Since we support a single operation (put) and we have a single use case for this.
Also keeping the desire to remove this in the future I wouldn't bother with passing the "put"
I think you can just rename the method to updateManifest or updateYaml.

src/installer/installer_test.go
```diff
@@ -179,13 +205,30 @@ var _ = Describe("installer HostRoleMaster role", func() {
 		extractSecretFromIgnitionSuccess := func() {
 			mockops.EXPECT().ExtractFromIgnition(filepath.Join(InstallDir, bootstrapIgn), dockerConfigFile).Return(nil).Times(1)
 		}
+		readFileSuccess := func(filename string, retBytes []byte, retErr error) {
+			mockops.EXPECT().ReadFile(filename).Return(retBytes, retErr)
+		}
+		patchMastersIgnitionSuccess := func() {
+			readFileSuccess("/etc/ssh/ssh_host_rsa_key.pub", []byte("ssh-rsa dummyKEY="), nil)
+			mockops.EXPECT().ExecCommand(gomock.Any(), "cp", "/99_openshift-machineconfig_99-additional-master-ssh.yaml", "/opt/openshift/openshift/").Return("OK", nil).Times(1)
```
eranco74, line 213: Use const for this

src/installer/installer_test.go
```diff
@@ -179,13 +205,30 @@ var _ = Describe("installer HostRoleMaster role", func() {
 		extractSecretFromIgnitionSuccess := func() {
 			mockops.EXPECT().ExtractFromIgnition(filepath.Join(InstallDir, bootstrapIgn), dockerConfigFile).Return(nil).Times(1)
 		}
+		readFileSuccess := func(filename string, retBytes []byte, retErr error) {
```
eranco74, line 208: Add tests for bad flow

src/types/machine_config.go
carbonin: Why? This should be pretty easy I think, right?

src/utils/utils.go
```diff
@@ -225,3 +227,26 @@ func SetNoProxyEnv(noProxy string) {
 	os.Setenv("NO_PROXY", noProxy)
 	os.Setenv("no_proxy", noProxy)
 }
+
+func parseField(s string) (string, bool, int) {
```
carbonin, line 231: This all seems rather error-prone. To @eranco74 's point in https://github.com/openshift/assisted-installer/pull/80/files#r519886807, can we just index into the struct directly for now rather than adding all this parsing code?

We can generalize when we have at least one other use case.

deploy/manifests/99_openshift-machineconfig_99-additional-master-ssh.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  config:
+    ignition:
+      version: 3.1.0
+    passwd:
+      users:
+      - name: core
+        sshAuthorizedKeys:
+        - REPLACE_SSH_KEY
```
carbonin, line 16: Any reason to not use an actual [go template](https://golang.org/pkg/text/template/)?

deploy/manifests/99_openshift-machineconfig_99-additional-master-ssh.yaml
```diff
@@ -0,0 +1,20 @@
...
+  fips: false
```
carbonin, line 17: Why include this and the other fields below? Can we remove these so we're only setting the ssh key?

deploy/manifests/99_openshift-machineconfig_99-additional-master-ssh.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  creationTimestamp: null
```
carbonin, line 4: This can be removed I think.

src/ops/ops.go
```diff
@@ -76,6 +83,7 @@ func NewOps(logger *logrus.Logger, proxySet bool) Ops {
 
 // ExecPrivilegeCommand execute a command in the host environment via nsenter
 func (o *ops) ExecPrivilegeCommand(liveLogger io.Writer, command string, args ...string) (string, error) {
+	o.log.Infof("Ececuting privilege comman: %s %s", command, args)
```
carbonin, line 86:
```suggestion
	o.log.Infof("Executing privileged command: %s %s", command, args)
```

A few typos

src/ops/ops.go
```diff
@@ -419,3 +427,35 @@ func (o *ops) ReloadHostFile(filepath string) error {
...
 	return nil
 }
+
+func (o *ops) ReadFile(filename string) ([]byte, error) {
+	return ioutil.ReadFile(filename)
+}
+
+func (o *ops) WriteFile(filename string, data []byte, perm os.FileMode) error {
+	return ioutil.WriteFile(filename, data, perm)
+}
```
carbonin, line 437: Is this really the only way to be able to test this?
It seems a shame to add unnecessary methods like this because of a test framework limitation.

src/installer/installer.go
ybettan: done

src/installer/installer.go
ybettan: done

src/installer/installer.go
ybettan: done

deploy/manifests/99_openshift-machineconfig_99-additional-master-ssh.yaml
ybettan: done

src/installer/installer.go
ybettan: done

src/installer/installer.go
ybettan: done

src/types/machine_config.go
ybettan: Because we wanted to insert this task yesterday at 22:00, obviously that didn't happened so I will do it in the same PR :)

src/installer/installer.go
ybettan: done

src/installer/installer.go
ybettan: done

src/installer/installer.go
ybettan: done

deploy/manifests/99_openshift-machineconfig_99-additional-master-ssh.yaml
ybettan: How you will process it? cluster isn't up yet (`oc process` won't work)

src/ops/ops.go
ybettan: Thanks

src/types/machine_config.go
ybettan: obsolete

src/installer/installer.go
ybettan: obsolete

src/installer/installer.go
ybettan: done

src/installer/installer_test.go
ybettan: obsolete

src/utils/utils.go
ybettan: removed

deploy/manifests/99_openshift-machineconfig_99-additional-master-ssh.yaml
ybettan: At first I was updating a current manifest so it wasn't an option, later we decided to create a new manifest for it and now it is a good idea.

Updated to use template.

src/ops/ops.go
ybettan: Obsolete

src/installer/installer.go
```diff
@@ -26,14 +26,37 @@ import (
...
+      version: 3.1.0
+    passwd:
+      users:
+      - name: core
+        sshAuthorizedKeys:
+        - {{.}}
+  fips: false
+  kernelArguments: null
+  kernelType:
+  osImageURL:
```
carbonin, line 58: I still don't understand why any of these keys are here. Do these have anything to do with setting an ssh key?

src/ops/ops.go
```diff
@@ -76,6 +78,7 @@ func NewOps(logger *logrus.Logger, proxySet bool) Ops {
 
 // ExecPrivilegeCommand execute a command in the host environment via nsenter
 func (o *ops) ExecPrivilegeCommand(liveLogger io.Writer, command string, args ...string) (string, error) {
+	o.log.Infof("Ececuting privilege command: %s %s", command, args)
```
carbonin, line 81:
```suggestion
	o.log.Infof("Executing privileged command: %s %s", command, args)
```

src/installer/installer.go
```diff
@@ -227,6 +260,25 @@ func (i *installer) extractIgnitionToFS(ignitionPath string) (err error) {
 	return err
 }
 
+func (i *installer) generateSshKeyPair() error {
+	i.log.Info("Generating new SSH key pair")
+	if _, err := i.ops.ExecCommand(utils.NewLogWriter(i.log), "ssh-keygen", "-q", "-f", "/root/.ssh/id_rsa", "-N", "''"); err != nil {
```
carbonin, line 265: Don't want to block the PR on this, but it feels like there should be a "native" way to generate a ssh keypair in go without using exec.

Then we wouldn't have to write and re-read the file as well.

src/installer/installer.go
ybettan: Didn't reach to it yet :)

src/installer/installer.go
```diff
@@ -227,6 +260,25 @@ func (i *installer) extractIgnitionToFS(ignitionPath string) (err error) {
...
+		i.log.Error(err)
+		return err
+	}
+	return nil
+}
+
+//FIXME:ybettan: do I need this function? it is directly calling ops.CreateFileFromTemplate
```
carbonin, line 272: I would put this directly in the caller. It's just creating extra lines of error checking.

src/ops/ops.go
ybettan: Just stayed consistent with the function name.

src/installer/installer.go
ybettan: There is but then something else will be needed:
I will need to write the output to `/root/.ssh` so I don't think we will gain anything from it.

src/ops/ops.go
carbonin: Okay, but "Executing" is still spelled incorrectly :wink:

ybettan: lol, missed that, fixing, I was lazy to run spell-check.

src/installer/installer_test.go
ybettan: done

src/installer/installer.go
ybettan: Agree

src/installer/installer.go
ybettan: done

deploy/manifests/99_openshift-machineconfig_99-additional-master-ssh.yaml
ybettan: done

deploy/manifests/99_openshift-machineconfig_99-additional-master-ssh.yaml
ybettan: We can, removing.

src/installer/installer.go
```diff
@@ -26,14 +26,32 @@ import (
...
+	InstallDir                   = "/opt/install-dir"
+	KubeconfigPathLoopBack       = "/opt/openshift/auth/kubeconfig-loopback"
+	KubeconfigPath               = "/opt/openshift/auth/kubeconfig"
+	minMasterNodes               = 2
+	dockerConfigFile             = "/root/.docker/config.json"
+	assistedControllerPrefix     = "assisted-installer-controller"
+	assistedControllerNamespace  = "assisted-installer"
+	extractRetryCount            = 3
+	assistedInstallerSshManifest = "/opt/openshift/openshift/99_openshift-machineconfig_99-assisted-installer-master-ssh.yaml"
+	sshManifestTmpl              = `
```
tsorya, line 38: can we have it as template file maybe? or at least move to another file.

src/ops/ops.go
```diff
@@ -437,3 +440,24 @@ func (o *ops) ReloadHostFile(filepath string) error {
 	}
 	return nil
 }
+
+func (o *ops) CreateOpenshiftSshManifest(filePath, tmpl string) error {
+	o.log.Info("Create an openshift manifets for SSH public key")
+	sshPublicKey, err := ioutil.ReadFile("/root/.ssh/id_rsa.pub")
```
tsorya, line 446: consts?

src/installer/installer.go
ybettan: This change will later move to `assisted-service`, so no reason to have an extra file for it which will need to be added to the Dockerfile which will lead to downstream-image-change.

Also according to @carbonin, it is short enough to be in code.

This is no different than what we are doing in `assisted-service` which BTW is a much longer file :)
https://github.com/ybettan/assisted-service/blob/be82d501cee20799a4e0d4d95527878886ca6a92/internal/bminventory/inventory.go#L134
https://github.com/ybettan/assisted-service/blob/be82d501cee20799a4e0d4d95527878886ca6a92/internal/bminventory/inventory.go#L199

tsorya: but template sounds good for me

src/ops/ops.go
ybettan: This is the only place it is used, I think it is more  clear this way.

src/ops/ops.go
```diff
@@ -437,3 +440,24 @@ func (o *ops) ReloadHostFile(filepath string) error {
...
+	if err != nil {
+		o.log.Error(err)
```
tsorya, line 448: o.log.WithError(err).Error("failed ....")

src/ops/ops.go
```diff
@@ -437,3 +440,24 @@ func (o *ops) ReloadHostFile(filepath string) error {
...
+		return err
+	}
+	f, err := os.Create(filePath)
+	if err != nil {
+		o.log.Error(err)
```
tsorya, line 453: same

src/ops/ops.go
```diff
@@ -437,3 +440,24 @@ func (o *ops) ReloadHostFile(filepath string) error {
...
+		return err
+	}
+	defer f.Close()
+	t := template.Must(template.New("openshift SSH manifest").Parse(tmpl))
+	if err := t.Execute(f, string(sshPublicKey)); err != nil {
+		o.log.Error(err)
```
tsorya, line 459: same

src/installer/installer.go
```diff
@@ -231,6 +259,15 @@ func (i *installer) extractIgnitionToFS(ignitionPath string) (err error) {
 	return err
 }
 
+func (i *installer) generateSshKeyPair() error {
+	i.log.Info("Generating new SSH key pair")
+	if _, err := i.ops.ExecPrivilegeCommand(utils.NewLogWriter(i.log), "ssh-keygen", "-q", "-f", "/root/.ssh/id_rsa", "-N", ""); err != nil {
+		i.log.Error(err)
```
tsorya, line 265: with error

src/installer/installer.go
```diff
@@ -231,6 +259,15 @@ func (i *installer) extractIgnitionToFS(ignitionPath string) (err error) {
 	return err
 }
 
+func (i *installer) generateSshKeyPair() error {
+	i.log.Info("Generating new SSH key pair")
+	if _, err := i.ops.ExecPrivilegeCommand(utils.NewLogWriter(i.log), "ssh-keygen", "-q", "-f", "/root/.ssh/id_rsa", "-N", ""); err != nil {
```
eranco74, line 264: Why use ExecPrivilegeCommand if `/root/ssh` is mounted?

src/ops/ops.go
```diff
@@ -77,6 +79,7 @@ func NewOps(logger *logrus.Logger, proxySet bool) Ops {
 
 // ExecPrivilegeCommand execute a command in the host environment via nsenter
 func (o *ops) ExecPrivilegeCommand(liveLogger io.Writer, command string, args ...string) (string, error) {
+	o.log.Infof("Executing privilege command: %s %s", command, args)
```
eranco74, line 82: Please remove this.
It  will spam the log

src/installer/installer.go
eranco74, line 264: "/root/.ssh/id_rsa" should be const

src/installer/installer_test.go
```diff
@@ -179,13 +180,21 @@ var _ = Describe("installer HostRoleMaster role", func() {
 		extractSecretFromIgnitionSuccess := func() {
 			mockops.EXPECT().ExtractFromIgnition(filepath.Join(InstallDir, bootstrapIgn), dockerConfigFile).Return(nil).Times(1)
 		}
+		generateSshKeyPairSuccess := func() {
+			mockops.EXPECT().ExecPrivilegeCommand(gomock.Any(), "ssh-keygen", "-q", "-f", "/root/.ssh/id_rsa", "-N", "").Return("OK", nil).Times(1)
```
eranco74, line 184: "/root/.ssh/id_rsa" shouldbe const

src/ops/ops.go
eranco74, line 446: "/root/.ssh/id_rsa" shouldbe const

src/installer/installer.go
tsorya: so just create new go file with it.

src/installer/installer.go
```diff
@@ -26,14 +26,32 @@ import (
...
+  name: 99-assisted-installer-master-ssh
+spec:
+  config:
+    ignition:
+      version: 3.1.0
+    passwd:
+      users:
+      - name: core
+        sshAuthorizedKeys:
+        - {{.}}
```
eranco74, line 53:
```suggestion
        - {{.sshPublicKey}}
```

src/installer/installer.go
tsorya: this file is already become long, so why to have all this context here?

src/installer/installer.go
ybettan: Because `ssh-keygen` doesn't exist on the container.

src/ops/ops.go
ybettan: done

src/installer/installer.go
ybettan: done

src/installer/installer_test.go
ybettan: done

src/ops/ops.go
ybettan: done

src/ops/ops.go
ybettan: done

src/ops/ops.go
ybettan: done

src/ops/ops.go
ybettan: done

src/installer/installer.go
ybettan: done

src/ops/ops.go
ybettan: done

src/installer/installer.go
ybettan: done.

src/installer/installer.go
```diff
@@ -231,6 +241,19 @@ func (i *installer) extractIgnitionToFS(ignitionPath string) (err error) {
...
 }
 
+func (i *installer) generateSshKeyPair() error {
+	i.log.Info("Generating new SSH key pair")
+	if err := i.ops.Mkdir(sshDir); err != nil {
+		i.log.Errorf("Failed to create SSH dir: %s", err)
+		return err
+	}
+	if _, err := i.ops.ExecPrivilegeCommand(utils.NewLogWriter(i.log), "ssh-keygen", "-q", "-f", sshKeyPath, "-N", ""); err != nil {
+		i.log.WithError(err).Error("Failed to generate SSH key pair")
```
eranco74, line 251: Be consistent 
o.log.Errorf / i.log.WithError(err).Error

eranco74 (commented):
Looks good.
remove the "/hold" when you want to merge it

src/installer/installer.go
ybettan: I missed that one when I replaced all `log.Errorf` with `log.WithError(err).Error()`, thanks.

PR #86, opened 2020-11-05
MGMT-2629 - No longer printing pull secrets

omertuc:


ronniel1 (commented):
Very impressive solution @omertuc 
Maybe worth putting this in a common place for all the services to use.
I think that assisted-service/pkg directory is used by all other repos.
@filanov what do you think?

Make sure it has extensive UT.
Also worth checking what happens if there are nested structs.

omertuc:
> Make sure it has extensive UT.
Will add some tests

> Also worth checking what happens if there are nested structs.
It works for nested structs (if the field is another struct it recurses) but I'm not sure how it'll behave with pointers to structs, I'll check

filanov:
> Very impressive solution @omertuc
> Maybe worth putting this in a common place for all the services to use.
> I think that assisted-service/pkg directory is used by all other repos.
> @filanov what do you think?
> 
> Make sure it has extensive UT.
> Also worth checking what happens if there are nested structs.

Agree with @ronniel1

omertuc:
Okay so I will move this PR to the assisted-service repo

filanov:
@omertuc the usage is good, the common part should move into assited-service repo.

omertuc:
> @omertuc the usage is good, the common part should move into assited-service repo.

Right. I'll reopen after the other PR is merged and I modify this PR accordingly

omertuc:
assisted-service PR https://github.com/openshift/assisted-service/pull/653

PR #94, opened 2020-11-09
MGMT-2629 PullSecretToken marked secret

omertuc:
Recreated https://github.com/openshift/assisted-installer/pull/86 because it wouldn't let me reopen

Same as before, but the dump function moved to assisted-service

Had to run 

```go get github.com/openshift/assisted-service@bb2f4a5dbbd8357716edeacc0ffbc540fa1e1757```

to get the relevant version of assisted-service, I hope it's alright.


filanov (approved):

filanov:
There is a debug log that print the pull secret https://github.com/openshift/assisted-installer/blob/master/src/ops/ops.go#L147
The log level is info for this  service right @eranco74 ?
Regardless if it was info i'm not really sure how to handle it.

PR #78, opened 2020-10-28
MGMT-2694 Add additional args to coreos-installer command

carbonin:
This allows for customizing things like kernel args as well as bringing networking files forward from the previous stage.

[MGMT-2694](https://issues.redhat.com/browse/MGMT-2694)

PR #95, opened 2020-11-10
MGMT-1263: allow overriding the rhcos image url

asalkeld:
This add RhcosOverrideImage to the global config so that a user can override the built-in rhcos image.

Note: [this PR](https://github.com/openshift/assisted-installer/pull/78) could also be used for this purpose

### DIFF
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -17,6 +17,7 @@ type Config struct {
 	URL                  string
 	Verbose              bool
 	OpenshiftVersion     string
+	RhcosOverrideImage   string
 	Hostname             string
 	ControllerImage      string
 	AgentImage           string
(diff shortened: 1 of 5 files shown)


carbonin:
@asalkeld I think it would be better to put this with the rest of the args in the service. Do you have a service-side PR I can take a look at?

asalkeld:
> @asalkeld I think it would be better to put this with the rest of the args in the service. Do you have a service-side PR I can take a look at?

I haven't made a server side PR yet. but should be ease to add the option on top of your PR https://github.com/openshift/assisted-service/pull/626

I'll close this one

carbonin:
Ah, right. I forgot that one wasn't merged yet :facepalm:

PR #96, opened 2020-11-10
MGMT-2465: assisted-controller image for day2

danielerez:
Added 'assisted-installer-controller-ocp' image for supporting the controller as deployment in ocp cluster:
- main: invokes periodically approve CSRs and update nodes statuses.
- Makefile: added targets for deploy, build and push.
- Added deploy resources to deploy/assisted-installer-controller-ocp.
- Added a new docker file for the image.


danielerez:
@yevgeny-shnaidman

src/main/assisted-installer-controller-ocp/main.go
```diff
@@ -0,0 +1,78 @@
...
+		kc,
+	)
+
+	logger.Infof("Controller deployed on OCP cluster")
+
+	for {
+		var wg sync.WaitGroup
+		ctxApprove, cancelApprove := context.WithCancel(context.Background())
+		go assistedController.ApproveCsrs(ctxApprove, &wg)
+		wg.Add(1)
```
yevgeny-shnaidman, line 64: this will surely work, but i would suggest that you create a specific goroutine function per ApproveCsr and nodeState, and the loop will be inside each of those functions. This way approveCSR won't have to wait until waitAndUpdate finished in order to be run again

danielerez: Moved to two separate goroutines

Makefile
```diff
@@ -2,9 +2,12 @@ CONTAINER_COMMAND = $(shell if [ -x "$(shell which docker)" ];then echo "docker"
 INSTALLER := $(or ${INSTALLER},quay.io/ocpmetal/assisted-installer:stable)
 GIT_REVISION := $(shell git rev-parse HEAD)
 CONTROLLER :=  $(or ${CONTROLLER}, quay.io/ocpmetal/assisted-installer-controller:stable)
+CONTROLLER_OCP :=  $(or ${CONTROLLER_OCP}, quay.io/ocpmetal/assisted-installer-controller-ocp:stable)
+CONTROLLER_OCP_IMAGE := $(or ${CONTROLLER_OCP_IMAGE},quay.io/ocpmetal/assisted-installer-controller-ocp:latest)
```
yevgeny-shnaidman, line 6: why do we need both of them? why not just CONTROLLER_OCP

danielerez: copy-paste error :) will remove one.

src/main/assisted-installer-controller-ocp/main.go
```diff
@@ -0,0 +1,86 @@
...
+	go approveCsrs(assistedController, &wg, logger)
+	wg.Add(1)
+	go waitAndUpdateNodesStatus(assistedController, &wg, logger)
+	wg.Add(1)
+	wg.Wait()
+}
+
+func waitAndUpdateNodesStatus(controller assistedinstallercontroller.Controller, wg *sync.WaitGroup, logger *logrus.Logger) {
+	defer wg.Done()
+	ctx := context.Background()
```
yevgeny-shnaidman, line 70: why do we need ctx and check if it Done? Nobody is going to cancel it anyway

src/main/assisted-installer-controller-ocp/main.go
```diff
@@ -0,0 +1,86 @@
...
+	ticker := time.NewTicker(assistedinstallercontroller.GeneralWaitInterval)
+	for {
+		select {
```
yevgeny-shnaidman, line 73: i think this function maybe implemented simpler:
for {
    controller.WaitAndUpdateNodesStatus()
    time.Sleep
}

src/main/assisted-installer-controller-ocp/main.go
danielerez: added just for consistency, but can remove it.

src/main/assisted-installer-controller-ocp/main.go
danielerez: Actually started with sleep approach but changed to a ticker to resemble the ApproveCsrs flow: https://github.com/openshift/assisted-installer/blob/master/src/assisted_installer_controller/assisted_installer_controller.go#L141
I can change again and see if looks clearer.

danielerez: done

src/main/assisted-installer-controller-ocp/main.go
danielerez: done

Makefile
danielerez: done

src/main/assisted-installer-controller-ocp/main.go
```diff
@@ -0,0 +1,79 @@
...
+
+	var wg sync.WaitGroup
+	go approveCsrs(assistedController, &wg, logger)
+	wg.Add(1)
+	go waitAndUpdateNodesStatus(assistedController, &wg, logger)
+	wg.Add(1)
+	wg.Wait()
+}
+
+func waitAndUpdateNodesStatus(controller assistedinstallercontroller.Controller, wg *sync.WaitGroup, logger *logrus.Logger) {
```
yevgeny-shnaidman, line 68: please remove wg, and just call wg.Done outside of the loop

danielerez: removed the wg. no need for Done actually as it should keep running.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -50,6 +50,7 @@ type ControllerConfig struct {
 
 type Controller interface {
 	WaitAndUpdateNodesStatus()
+	ApproveCsrs(ctx context.Context, wg *sync.WaitGroup)
```
yevgeny-shnaidman, line 53: one last comment an that's it:)
let's remove it for now, i don't want the changes in day1 code at all, the NewController returns object and not interface

danielerez: sure, removed from the interface.

src/main/assisted-installer-controller-ocp/main.go
eranco74: Why is this required (approving CSR), why not let cluster-machine-approver handle it?

Makefile
eranco74: Why do we need a new image?
Why not use the current controller image with a different entry point?
@gamli75 FYI :point_up:

yevgeny-shnaidman: 2 reasons:
1) we don't want to interfere with the day1 in anyway, so a new image will not affect anything, including DS buillds
2) currently we are delivering milestone1, which is mainly allowing to run on test-infra. the assisted-installer code won't go into production until we move both of them (service and installer) to be run as operators. This is a major task ( i am not even sure who will do it), and i am not even sure if it will use the same repo for it

src/main/assisted-installer-controller-ocp/main.go
yevgeny-shnaidman: because currently we don't provide BMHs in day2. Once we wil start using BMHs (probably 4.7),we can skip approve CSR

eranco74 (commented):
/lgtm
/approve

.github/workflows/build.yml
```diff
@@ -55,6 +55,17 @@ jobs:
         dockerfile: Dockerfile.assisted-installer-controller
         buildargs: GIT_REVISION, QUAY_TAG_EXPIRATION
         tags: "PR_${{ env.GIT_REVISION }}"
+    - name: Publish Controller OCP to Registry
```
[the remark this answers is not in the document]
danielerez: yeah, removed in: https://github.com/openshift/assisted-installer/pull/110

Makefile
gamli75: This change will require a new destination project in the midstream - @djzager can you review this PR? let's talk about the changes that needed for the DS.

yevgeny-shnaidman: @gamli75 this new image needs not to be in DS, currently it will be used only in assisted-test-infra

PR #98, opened 2020-11-12
MGMT-2634 Remove host-name flag

carbonin:
This is not used as of 93495e81245278863102d155b15b4d0782641ff4 and is no longer being sent from the service as of openshift/assisted-service#670


carbonin:
/hold

until openshift/assisted-service#670 is merged

YuviGold (approved):

PR #100, opened 2020-11-12
NO-ISSUE Add installer args to test cases

carbonin:
In response to https://github.com/openshift/assisted-installer/pull/78#discussion_r522160320


eranco74 (commented):
/lgtm
Wondering, is there a way for the assisted-installer to get invalid installer-args?

carbonin:
> is there a way for the assisted-installer to get invalid installer-args?

Yeah, definitely. The validation we have in the service isn't going to be able to handle every case.

PR #101, opened 2020-11-14
MGMT-2411: Fixed error format returned from assisted client

RazRegev:
Currently due to a bug in assisted-service API, the Error() function
prints memory addresses instead of values.
Fixing the bug in assisted-service side will break backward
compatibility, therefore the fix should be made here.

Before logging the error, try to convert it to one of assisted
service Error or InfraError errors, and build the message from
the payload.

See an example of the bad error format:
Oct 09 04:43:36 localhost installer[182675]: time="2020-10-09T04:43:36Z" level=error msg="Failed to fetch file (worker.ign) from server. err: [GET /clusters/{cluster_id}/downloads/files][409] downloadClusterFilesConflict  &{Code:0xc00038aa70 Href:0xc00038aa80 ID:0xc000541e04 Kind:0xc00038aa90 Reason:0xc00038aaa0}"

Oct 09 04:43:36 localhost installer[182675]: time="2020-10-09T04:43:36Z" level=info msg="Updating node installation stage: Failed - [GET /clusters/{cluster_id}/downloads/files][409] downloadClusterFilesConflict  &{Code:0xc00038aa70 Href:0xc00038aa80 ID:0xc000541e04 Kind:0xc00038aa90 Reason:0xc00038aaa0}"

### DIFF
--- a/src/utils/error_utils.go
+++ b/src/utils/error_utils.go
@@ -1,0 +1,62 @@
+package utils
+
+import (
+	"fmt"
+
+	"github.com/go-openapi/swag"
+	"github.com/openshift/assisted-service/models"
+)
+
+type AssistedServiceErrorAPI interface {
+	Error() string
+	GetPayload() *models.Error
+}
+
+type AssistedServiceError struct {
+	Payload *models.Error
+}
+
+func (err AssistedServiceError) Error() string {
+	return fmt.Sprintf(
+		"AssistedServiceError Code: %s Href: %s ID: %d Kind: %s Reason: %s",
+		swag.StringValue(err.Payload.Code),
+		swag.StringValue(err.Payload.Href),
+		swag.Int32Value(err.Payload.ID),
+		swag.StringValue(err.Payload.Kind),
+		swag.StringValue(err.Payload.Reason))
+}
+
+func (err AssistedServiceError) GetPayload() *models.Error {
+	return err.Payload
+}
+
+type AssistedServiceInfraErrorAPI interface {
+	Error() string
+	GetPayload() *models.InfraError
+}
+
+type AssistedServiceInfraError struct {
+	Payload *models.InfraError
+}
+
+func (err AssistedServiceInfraError) Error() string {
+	return fmt.Sprintf(
+		"AssistedServiceInfraError Code: %d Message: %s",
+		swag.Int32Value(err.Payload.Code),
+		swag.StringValue(err.Payload.Message))
+}
+
+func (err AssistedServiceInfraError) GetPayload() *models.InfraError {
+	return err.Payload
+}
+
+func GetAssistedError(err error) error {
+	switch err := err.(type) {
+	case AssistedServiceErrorAPI:
+		return AssistedServiceError{err.GetPayload()}
+	case AssistedServiceInfraErrorAPI:
+		return AssistedServiceInfraError{err.GetPayload()}
+	default:
+		return err
+	}
+}
--- a/src/inventory_client/inventory_client.go
+++ b/src/inventory_client/inventory_client.go
@@ -176,7 +176,7 @@ func (c *inventoryClient) DownloadFile(ctx context.Context, filename string, des
 		fo.Close()
 	}()
 	_, err = c.ai.Installer.DownloadClusterFiles(ctx, c.createDownloadParams(filename), fo)
-	return err
+	return utils.GetAssistedError(err)
 }
 
 func (c *inventoryClient) DownloadHostIgnition(ctx context.Context, hostID string, dest string) error {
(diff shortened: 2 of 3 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -78,7 +78,7 @@ func (c *controller) WaitAndUpdateNodesStatus() {
 		time.Sleep(GeneralWaitInterval)
 		assistedInstallerNodesMap, err := c.ic.GetHosts(ignoreStatuses)
 		if err != nil {
-			c.log.WithError(err).Error("Failed to get node map from inventory")
+			c.log.WithError(utils.WrappedAssistedServiceErrorMessage(err, "")).Error("Failed to get node map from inventory")
```
eranco74, line 81: Consider moving the usage of this wrapper into ic ( I guess it's InventoryClient) this way it is used just in that module and the rest of the code won't be aware of the error behavior.

src/utils/error_utils.go
```diff
@@ -0,0 +1,45 @@
...
+	Error() string
+	GetPayload() *models.Error
+}
+
+type AssistedServiceInfraError interface {
+	Error() string
+	GetPayload() *models.InfraError
+}
+
+func BuildMessageFromAssistedServiceError(err error) string {
```
filanov, line 21: I think that this function should return an error then you will not need the next function `WrappedAssistedServiceErrorMessage` because you will be able to use the errors package to wrap an error.  
Then the name should probably change as well, `GetAssistedError` or something like this.

src/utils/error_utils.go
```diff
@@ -0,0 +1,45 @@
...
+	case AssistedServiceError:
+		payload := err.GetPayload()
+		return fmt.Sprintf(
+			"AssistedServiceError Code: %s Href: %s ID: %d Kind: %s Reason: %s",
+			swag.StringValue(payload.Code),
+			swag.StringValue(payload.Href),
+			swag.Int32Value(payload.ID),
+			swag.StringValue(payload.Kind),
+			swag.StringValue(payload.Reason))
+	case AssistedServiceInfraError:
```
eranco74, line 32: Curious. what is AssistedServiceInfraError?

filanov (commented):
Please add tests.

src/utils/error_utils.go
```diff
@@ -0,0 +1,45 @@
...
+	GetPayload() *models.InfraError
+}
+
+func BuildMessageFromAssistedServiceError(err error) string {
+	switch err := err.(type) {
+	case AssistedServiceError:
+		payload := err.GetPayload()
+		return fmt.Sprintf(
+			"AssistedServiceError Code: %s Href: %s ID: %d Kind: %s Reason: %s",
+			swag.StringValue(payload.Code),
```
eranco74, line 27: Can we get the request ID from the payload?

src/utils/error_utils.go
```diff
@@ -0,0 +1,45 @@
...
+			swag.StringValue(payload.Href),
+			swag.Int32Value(payload.ID),
```
eranco74, line 29: Isn't this the request ID?

src/utils/error_utils.go
filanov: A hack for Auth/Authz the infra return `runtime.APIError` we have a way to fix it but didn't handled it yet.

src/assisted_installer_controller/assisted_installer_controller.go
filanov: +1

src/utils/error_utils.go
filanov: As far as i know request id is only in the context.

filanov (commented):
I was thinking that you don't need the objects `AssistedServiceError` and `AssistedServiceInfraError` 
You can just write something like 
```
package main

import (
	"github.com/pkg/errors"

	"github.com/go-openapi/swag"
	"github.com/openshift/assisted-service/models"
)

type AssistedServiceErrorAPI interface {
	Error() string
	GetPayload() *models.Error
}

type AssistedServiceInfraErrorAPI interface {
	Error() string
	GetPayload() *models.InfraError
}

func assistedErrorToError(err AssistedServiceErrorAPI) error {
	payload := err.GetPayload()
	return errors.Errorf(
		"AssistedServiceError Code: %s Href: %s ID: %d Kind: %s Reason: %s",
		swag.StringValue(payload.Code),
		swag.StringValue(payload.Href),
		swag.Int32Value(payload.ID),
		swag.StringValue(payload.Kind),
		swag.StringValue(payload.Reason))
}

func infraErrorToError(err AssistedServiceInfraErrorAPI) error {
	payload := err.GetPayload()
	return errors.Errorf(
		"AssistedServiceInfraError Code: %d Message: %s",
		swag.Int32Value(payload.Code),
		swag.StringValue(payload.Message))
}

func GetAssistedError(err error) error {
	switch err := err.(type) {
	case AssistedServiceErrorAPI:
		return assistedErrorToError(err)
	case AssistedServiceInfraErrorAPI:
		return infraErrorToError(err)
	default:
		return err
	}

```

filanov:
@ronniel1 maybe it should be as part of assisted-service? it looks like something that can be used by the agent as well.

RazRegev:
> I was thinking that you don't need the objects `AssistedServiceError` and `AssistedServiceInfraError`
> You can just write something like
> 
> ```
> package main
> 
> import (
> 	"github.com/pkg/errors"
> 
> 	"github.com/go-openapi/swag"
> 	"github.com/openshift/assisted-service/models"
> )
> 
> type AssistedServiceErrorAPI interface {
> 	Error() string
> 	GetPayload() *models.Error
> }
> 
> type AssistedServiceInfraErrorAPI interface {
> 	Error() string
> 	GetPayload() *models.InfraError
> }
> 
> func assistedErrorToError(err AssistedServiceErrorAPI) error {
> 	payload := err.GetPayload()
> 	return errors.Errorf(
> 		"AssistedServiceError Code: %s Href: %s ID: %d Kind: %s Reason: %s",
> 		swag.StringValue(payload.Code),
> 		swag.StringValue(payload.Href),
> 		swag.Int32Value(payload.ID),
> 		swag.StringValue(payload.Kind),
> 		swag.StringValue(payload.Reason))
> }
> 
> func infraErrorToError(err AssistedServiceInfraErrorAPI) error {
> 	payload := err.GetPayload()
> 	return errors.Errorf(
> 		"AssistedServiceInfraError Code: %d Message: %s",
> 		swag.Int32Value(payload.Code),
> 		swag.StringValue(payload.Message))
> }
> 
> func GetAssistedError(err error) error {
> 	switch err := err.(type) {
> 	case AssistedServiceErrorAPI:
> 		return assistedErrorToError(err)
> 	case AssistedServiceInfraErrorAPI:
> 		return infraErrorToError(err)
> 	default:
> 		return err
> 	}
> ```

right, assuming we won't need the original payload I'll fix this accordinally.

ronniel1:
> @ronniel1 maybe it should be as part of assisted-service? it looks like something that can be used by the agent as well.

@RazRegev , @filanov I think that makes sense

RazRegev:
duplicates https://github.com/openshift/assisted-installer/pull/116

PR #102, opened 2020-11-15
MGMT-2409: Added request id to each InventoryClient request

RazRegev:
In order to be ease the debug process of requests and responses
sent from assisted-installer to assisted-service and backwards a
request id will be added to each request. Loggers will print any
errors returned from assisted-service with this request id (same
as it works in assisted-installer-agent).

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -76,15 +76,17 @@ func (c *controller) WaitAndUpdateNodesStatus() {
 		models.HostStatusError, models.HostStatusInstalled}
 	for {
 		time.Sleep(GeneralWaitInterval)
-		assistedInstallerNodesMap, err := c.ic.GetHosts(ignoreStatuses)
+		ctx := utils.GenerateRequestContext()
+		log := utils.RequestIDLogger(ctx, c.log)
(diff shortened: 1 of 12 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -76,15 +76,17 @@ func (c *controller) WaitAndUpdateNodesStatus() {
 		models.HostStatusError, models.HostStatusInstalled}
 	for {
 		time.Sleep(GeneralWaitInterval)
-		assistedInstallerNodesMap, err := c.ic.GetHosts(ignoreStatuses)
+		ctx := utils.GenerateRequestContext()
+		log := utils.RequestIDLogger(ctx, c.log)
+		assistedInstallerNodesMap, err := c.ic.GetHosts(ctx, log, ignoreStatuses)
```
eranco74, line 81: What?
Is this actually required (generating a request ID and setting a logger)?
Can't we get the request ID (generated by the service) from the response?

RazRegev: yes, like the agent, we need to generate ID here and send it to assisted-service. 
if we won't send an id assisted-service will not generate one.

eranco74 (commented):
/approve
/lgtm

PR #109, opened 2020-11-16
MGMT-2403 Add status flag to the post-reboot processes

slaviered:
* Part 2 of add must gather logs
  * Add indication when all the hosts are in error state ([MGMT-2711](https://issues.redhat.com/browse/MGMT-2711))

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
@@ -49,7 +50,11 @@ type ControllerConfig struct {
 }
 
 type Controller interface {
-	WaitAndUpdateNodesStatus()
+	WaitAndUpdateNodesStatus(status *ControllerStatus)
+}
+
+type ControllerStatus struct {
+	errCounter uint32
 }
 
 type controller struct {
@@ -70,22 +75,41 @@ func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inv
 	}
 }
 
-func (c *controller) WaitAndUpdateNodesStatus() {
+func (status *ControllerStatus) Error() {
+	atomic.AddUint32(&status.errCounter, 1)
+}
+
+func (status *ControllerStatus) HasError() bool {
+	return atomic.LoadUint32(&status.errCounter) > 0
+}
+
+func (c *controller) WaitAndUpdateNodesStatus(status *ControllerStatus) {
 	c.log.Infof("Waiting till all nodes will join and update status to assisted installer")
-	ignoreStatuses := []string{models.HostStatusDisabled,
-		models.HostStatusError, models.HostStatusInstalled}
+	ignoreStatuses := []string{models.HostStatusDisabled, models.HostStatusInstalled}
+	var hostsInError int
 	for {
 		time.Sleep(GeneralWaitInterval)
 		ctx := utils.GenerateRequestContext()
 		log := utils.RequestIDLogger(ctx, c.log)
+
 		assistedInstallerNodesMap, err := c.ic.GetHosts(ctx, log, ignoreStatuses)
 		if err != nil {
 			log.WithError(err).Error("Failed to get node map from inventory")
 			continue
 		}
+		errNodesMap := common.FilterHostsByStatus(assistedInstallerNodesMap, []string{models.HostStatusError})
+		hostsInError = len(errNodesMap)
+
+		//if all hosts are in error, mark the failure and finish
+		if hostsInError > 0 && hostsInError == len(assistedInstallerNodesMap) {
+			status.Error()
+			break
+		}
+		//if all hosts are successfully installed, finish
 		if len(assistedInstallerNodesMap) == 0 {
 			break
 		}
+		//otherwise, update the progress status and keep waiting
 		log.Infof("Searching for host to change status, number to find %d", len(assistedInstallerNodesMap))
 		nodes, err := c.kc.ListNodes()
 		if err != nil {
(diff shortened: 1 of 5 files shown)


src/inventory_client/inventory_client.go
```diff
@@ -219,14 +219,20 @@ func (c *inventoryClient) GetCluster(ctx context.Context) (*models.Cluster, erro
...
-	return c.GetHosts(ctx, log, []string{models.HostStatusDisabled})
+	hosts, errorhosts, err := c.GetHosts(ctx, log, []string{models.HostStatusDisabled})
+	for k, v := range errorhosts {
+		hosts[k] = v
+	}
+	return hosts, err
 }
 
-func (c *inventoryClient) GetHosts(ctx context.Context, log logrus.FieldLogger, skippedStatuses []string) (map[string]HostData, error) {
+func (c *inventoryClient) GetHosts(ctx context.Context, log logrus.FieldLogger, skippedStatuses []string) (map[string]HostData, map[string]HostData, error) {
```
tsorya, line 229: i don't like the idea to return error hosts as separate var

tsorya: i think u can create func that gets list of error hosts and return list of them. Or lets create function that do what u want specifically. I think it is strange behavior for GetHosts

slaviered:
But then you'll end up with 2 REST calls for the same thing which is
inefficient. I can change the return type to return a single map of maps
indexed by status (as done in assisted-service inner utilities)

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -253,6 +270,8 @@ func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
 		if allUpdated {
 			c.log.Infof("Updated all the BMH CRs, finished successfully")
 			return
+		} else {
```
tsorya, line 273: i don't think this is the right behavior, we don't need set error here

tsorya: i think it will retry here, u can add timeout maybe but we don't need to set error if we don't exit the loop

src/inventory_client/inventory_client.go
tsorya: or map in our case

slaviered: DONE

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: DONE

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -70,22 +75,41 @@ func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inv
...
 		assistedInstallerNodesMap, err := c.ic.GetHosts(ctx, log, ignoreStatuses)
 		if err != nil {
 			log.WithError(err).Error("Failed to get node map from inventory")
 			continue
 		}
+		errNodesMap := c.ic.FilterHostsByStatus(assistedInstallerNodesMap, models.HostStatusError)
+		hostsInError = len(errNodesMap)
+
+		//if all hosts are in error, mark the failure and finish
+		if hostsInError > 0 && hostsInError == len(assistedInstallerNodesMap) {
```
tsorya, line 104: i think it is not the right behavior, what if only 1 worker is in error?

slaviered: so actually we should ask here "if all master hosts are in error"? basically it was meant to catch the situation of timeout where all hosts are moved to "error" state

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -178,7 +201,7 @@ func isCsrApproved(csr *certificatesv1beta1.CertificateSigningRequest) bool {
 	return false
 }
 
-func (c controller) PostInstallConfigs(wg *sync.WaitGroup) {
+func (c controller) PostInstallConfigs(wg *sync.WaitGroup, status *ControllerStatus) {
```
tsorya, line 204: I think we miss logic here. If we have cluster in error there is not point to continue to wait for finalizing. We can just exit it

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -70,22 +75,41 @@ func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inv
...
 		time.Sleep(GeneralWaitInterval)
 		ctx := utils.GenerateRequestContext()
 		log := utils.RequestIDLogger(ctx, c.log)
+
 		assistedInstallerNodesMap, err := c.ic.GetHosts(ctx, log, ignoreStatuses)
 		if err != nil {
 			log.WithError(err).Error("Failed to get node map from inventory")
 			continue
 		}
+		errNodesMap := c.ic.FilterHostsByStatus(assistedInstallerNodesMap, models.HostStatusError)
```
tsorya, line 100: I think will be very helpful to be able to filter by status as a list. What do u think?

slaviered: I don't know because we haven't got a real use case for it, but it can be done. I'll fix

src/inventory_client/inventory_client.go
```diff
@@ -222,6 +223,16 @@ func (c *inventoryClient) GetEnabledHostsNamesHosts(ctx context.Context, log log
 	return c.GetHosts(ctx, log, []string{models.HostStatusDisabled})
 }
 
+func (c *inventoryClient) FilterHostsByStatus(hosts map[string]HostData, status string) map[string]HostData {
```
tsorya, line 226: lets move it to assisted-installer/src/common/common.go

tsorya: at least i think so. I think that functions in inventory client is meant to use inventory client no?

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: yeah, got it

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: This edge case (of all hosts in error) usually occur when we timeout for waiting for control plane. I believe this exceeds all other timeouts?

slaviered:
@tsorya fixed code review comments and tested it with full e2e tests of test-infra (~8 hours)

PR #111, opened 2020-11-17
Refactor Ops

slaviered:
* Separate command executor from Ops interfce to allow mocking in unit tests
   * This is Part 3 of [MGMT-2403](https://issues.redhat.com/browse/MGMT-2403)


src/installer/installer.go
```diff
@@ -209,13 +209,7 @@ func (i *installer) extractIgnitionToFS(ignitionPath string) (err error) {
...
 	i.log.Infof("Extracting ignition to disk using %s mcoImage", mcoImage)
 	for j := 0; j < extractRetryCount; j++ {
-		_, err = i.ops.ExecPrivilegeCommand(utils.NewLogWriter(i.log), "podman", "run", "--net", "host",
-			"--volume", "/:/rootfs:rw",
-			"--volume", "/usr/bin/rpm-ostree:/usr/bin/rpm-ostree",
-			"--privileged",
-			"--entrypoint", "/usr/bin/machine-config-daemon",
-			mcoImage,
-			"start", "--node-name", "localhost", "--root-mount", "/rootfs", "--once-from", ignitionPath, "--skip-reboot")
+		err = i.ops.ExtractIgnitionWithMCO(mcoImage, ignitionPath)
```
carbonin, line 212: Maybe `ExtractIgnitionFromMCO`?

src/ops/ops.go
```diff
@@ -56,6 +54,7 @@ type ops struct {
 	log       *logrus.Logger
 	logWriter *utils.LogWriter
 	cmdEnv    []string
+	Runner    opsexec.Excecutor
```
carbonin, line 57: Any reason not to use a pointer here?

src/ops/ops.go
```diff
@@ -3,33 +3,31 @@ package ops
...
 	"strings"
-	"syscall"
 	"text/template"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	"github.com/openshift/assisted-installer/src/config"
 	"github.com/openshift/assisted-installer/src/inventory_client"
+	opsexec "github.com/openshift/assisted-installer/src/ops/exec"
```
carbonin, line 18: Maybe `ops_exec` here?

src/ops/ops.go
carbonin: If the goal is to "hide" the details of how we are actually running these commands, can this not be exported?
And maybe we use either "runner" or "executor" for both the type and the struct field to reduce the number of names we use for the same thing.

carbonin (commented):
Just a few comments about naming.
Maybe also add a test that would not have been possible (or very difficult) to write before this change? Looks like `GetVGByPV` would be a good candidate for some tests.

slaviered:
closing it for now. will not merge it with the main feature

PR #47, opened 2020-09-28
Eran/remove etcd patch

eranco74:

PR #115, opened 2020-11-23
MGMT-2454 Remove patch for unsupported etcd flag for versions >= 4.7

carbonin:
Going forward we will use a marker file written to the bootstrap
node. This is implemented in the service in  openshift/assisted-service#672
and the etcd operator in openshift/cluster-etcd-operator#449

We also need to consider what version of openshift will contain the
etcd operator enhancement we require before merging this.

Fixes [MGMT-2454](https://issues.redhat.com/browse/MGMT-2454)
Alternative to https://github.com/openshift/assisted-installer/pull/47

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -48,7 +48,8 @@ type ControllerConfig struct {
 	PullSecretToken      string `envconfig:"PULL_SECRET_TOKEN" required:"true"`
 	SkipCertVerification bool   `envconfig:"SKIP_CERT_VERIFICATION" required:"false" default:"false"`
 	CACertPath           string `envconfig:"CA_CERT_PATH" required:"false" default:""`
-	Namespace            string `enconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
+	Namespace            string `envconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
+	OpenshiftVersion     string `envconfig:"OPENSHIFT_VERSION" required:"true"`
 }
 type Controller interface {
 	WaitAndUpdateNodesStatus(status *ControllerStatus)
@@ -236,9 +237,17 @@ func (c controller) postInstallConfigs() error {
 		return errors.Errorf("Timeout while waiting router ca data")
 	}
 
-	err = utils.WaitForPredicate(WaitTimeout, GeneralWaitInterval, c.unpatchEtcd)
+	unpatch, err := utils.EtcdPatchRequired(c.ControllerConfig.OpenshiftVersion)
 	if err != nil {
-		return errors.Errorf("Timeout while trying to unpatch etcd")
+		return err
+	}
+	if unpatch {
+		err = utils.WaitForPredicate(WaitTimeout, GeneralWaitInterval, c.unpatchEtcd)
+		if err != nil {
+			return errors.Errorf("Timeout while trying to unpatch etcd")
+		}
+	} else {
+		c.log.Infof("Skipping etcd unpatch for cluster version %s", c.ControllerConfig.OpenshiftVersion)
 	}
 
 	err = utils.WaitForPredicate(WaitTimeout, GeneralWaitInterval, c.validateConsolePod)
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -275,9 +275,18 @@ func (i *installer) downloadHostIgnition() (string, error) {
 func (i *installer) waitForControlPlane(ctx context.Context, kc k8s_client.K8SClient) error {
 	i.waitForMasterNodes(ctx, minMasterNodes, kc)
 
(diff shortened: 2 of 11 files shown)


carbonin:
Made this conditional on the version we're installing. Will unhold when I test with both 4.6 and 4.7

src/utils/utils.go
```diff
@@ -230,3 +230,16 @@ func GenerateRequestContext() context.Context {
...
+	clusterVersion, err := version.NewVersion(openshiftVersion)
+	if err != nil {
+		return false, err
+	}
+	v47, err := version.NewVersion("4.7")
+	if err != nil {
+		return false, err
+	}
+
+	return clusterVersion.LessThan(v47), nil
```
romfreiman, line 244: 4.6.8 is less that 4.7? I assume go-version support it, right?

src/utils/utils.go
```diff
@@ -230,3 +230,16 @@ func GenerateRequestContext() context.Context {
 func RequestIDLogger(ctx context.Context, log *logrus.Logger) logrus.FieldLogger {
 	return requestid.RequestIDLogger(log, requestid.FromContext(ctx))
 }
+
+func EtcdPatchRequired(openshiftVersion string) (bool, error) {
```
romfreiman, line 234: please log! How would u know what pass did the code take - it would be very hard to debug

src/installer/installer.go
```diff
@@ -275,10 +275,17 @@ func (i *installer) downloadHostIgnition() (string, error) {
 func (i *installer) waitForControlPlane(ctx context.Context, kc k8s_client.K8SClient) error {
 	i.waitForMasterNodes(ctx, minMasterNodes, kc)
 
-	if err := kc.PatchEtcd(); err != nil {
+	patch, err := utils.EtcdPatchRequired(i.Config.OpenshiftVersion)
+	if err != nil {
 		i.log.Error(err)
 		return err
 	}
+	if patch {
```
romfreiman, line 283: log.....

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -236,9 +237,15 @@ func (c controller) postInstallConfigs() error {
 		return errors.Errorf("Timeout while waiting router ca data")
 	}
 
-	err = utils.WaitForPredicate(WaitTimeout, GeneralWaitInterval, c.unpatchEtcd)
+	unpatch, err := utils.EtcdPatchRequired(c.ControllerConfig.OpenshiftVersion)
```
romfreiman, line 240: log....

romfreiman:
Please add a UT with 4.6.6 and 4.7.5 (just to verify patch releases are working).

src/utils/utils.go
carbonin: The actual [k8s client functions log when they patch and unpatch](https://github.com/openshift/assisted-installer/blob/8a36dac584f7aeec89db4f4df68bd7c6d5c45308/src/k8s_client/k8s_client.go#L150-L170).
I can put an additional log in the non-patch case though if you think that's necessary.

src/utils/utils.go
carbonin: Yup, added some more tests for these kinds of cases.

src/utils/utils.go
romfreiman: if u'll not do patch/unpatch - then there will be no log. It's better to have it explicit. Than u can search in ELK for example

romfreiman:
Btw, lets not affect the current flow - meaning if this code will be deployed to the cloud now - will it work? Or it will require marking somewhere that it installs 4.6. If it wild require extra changes, then - not good 👎

carbonin:
> meaning if this code will be deployed to the cloud now - will it work?

I thought it would, but I'm hitting some issues running test infra. I'll know for sure tomorrow.

carbonin:
Verified that this doesn't break 4.6, but waiting on the 4.7 MCO issue to continue testing that.

src/utils/utils.go
carbonin: Logged in the callers.

carbonin:
Also just validated on 4.7 with nightly release `4.7.0-0.nightly-2020-12-10-202001`

Validated that we didn't path or unpatch as well.
```
# grep -Rn "Skipping etcd" *
logs_host_f8384dab-5526-4418-a7e1-13e23bc331ce/installer.logs:406:Dec 16 19:53:33 test-infra-cluster-assisted-installer-master-1 installer[3042]: time="2020-12-16T19:53:33Z" level=info msg="Skipping etcd patch for cluster version 4.7"
# oc logs -n assisted-installer  assisted-installer-controller-wfc8f | grep "Skipping etcd unpatch"
time="2020-12-16T20:08:53Z" level=info msg="Skipping etcd unpatch for cluster version 4.7"
```

/unhold

Any other concerns @romfreiman @YuviGold @eranco74 ?

PR #117, opened 2020-11-25
MGMT-2403 Add must-gather logs

slaviered:


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -430,41 +439,96 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
...
-// write tar.gz to pipe in a routine
-// upload tar.gz from pipe to assisted service.
-// close read and write pipes
-func (c controller) uploadPodLogs(podName string, namespace string, sinceSeconds int64) error {
+func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSeconds int64, isMustGatherEnabled bool) error {
+	var errz = make([]error, 0)
+	var tarentries = make([]utils.TarEntry, 0)
+	ctx := utils.GenerateRequestContext()
+
+	var returnVal = func() error {
```
tsorya, line 447: why we need this logic? Why just not to add log line to every error? I think it is too much for just printing log line

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -430,41 +439,96 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
...
+	}
+
+	if isMustGatherEnabled {
+		if tarfile, err := c.collectMustGatherLogs(ctx); err == nil {
+			if entry, tarerr := utils.NewTarEntryFromFile(tarfile); tarerr == nil {
+				tarentries = append(tarentries, *entry)
+			} else {
+				errz = append(errz, errors.Wrapf(tarerr, "Failed to open must-gather logs tar"))
+			}
+		} else {
```
tsorya, line 471: u print log for error inside collectMustGatherLogs, not sure why need one more log line? lets have a little less if's here

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -430,41 +439,96 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
...
+	if podLogs, err := c.kc.GetPodLogsAsBuffer(namespace, podName, sinceSeconds); err == nil {
+		tarentries = append(tarentries,
+			*utils.NewTarEntry(podLogs, nil, int64(podLogs.Len()), fmt.Sprintf("%s.logs", podName)))
+	} else {
+		errz = append(errz, errors.Wrapf(err, "Failed to get logs of pod %s", podName))
+	}
+
+	if isMustGatherEnabled {
+		if tarfile, err := c.collectMustGatherLogs(ctx); err == nil {
+			if entry, tarerr := utils.NewTarEntryFromFile(tarfile); tarerr == nil {
```
tsorya, line 466: maybe providing log to NewTarEntryFromFile we can remove else from here?

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: its not for log lines at all. I want to be resilient for errors and collect the most logs I can. so if collection of must-gather fails I want to still be able to collect controller logs and vice versa. This is why I accumulate the errors instead of aborting the entire process at the first error we encounter - and then flatten them to a single error at the end.

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: it help us to see where we are in the process... I can remove it but I recommend to keep it for debugging purposes

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: The else clause is for error aggregation and not for the purpose of log collection

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -46,9 +48,16 @@ type ControllerConfig struct {
 	PullSecretToken      string `envconfig:"PULL_SECRET_TOKEN" required:"true"`
 	SkipCertVerification bool   `envconfig:"SKIP_CERT_VERIFICATION" required:"false" default:"false"`
 	CACertPath           string `envconfig:"CA_CERT_PATH" required:"false" default:""`
+	MustGatherMode       string `envconfig:"MUST_GATHER_MODE" required:"false" default:"Disabled"`
```
tsorya, line 51: we are going to provide it in configmap?

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: I think it is overcomplicated. Why u need to accumulate errors here? U do nothing with them.

tsorya: and u only return one at the end. The only thing u do is printing log in line 451 and nothing else

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: as i said , i think we don't need error aggregation. Removing it will make this func much more readable

tsorya: and providing log to utils will allow us to remove this else.

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: but we have it already inside collectMustGatherLogs ? why to have 2 same logs?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -430,41 +439,96 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
...
+
+func (c controller) collectMustGatherLogs(ctx context.Context) (string, error) {
+	tempDir, ferr := ioutil.TempDir("/tmp", "controller-must-gather-logs-")
+	if ferr != nil {
+		c.log.Errorf("Failed to create temp directory for must-gather-logs %v\n", ferr)
+		return "", ferr
+	}
+
+	//donwload kubeconfig file
+	kubeconfigPath := path.Join(tempDir, "kubeconfig-noingress")
```
tsorya, line 511: make consts from kubeconfig-noingress

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -430,41 +439,96 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
...
-		return errors.Wrapf(err, "Failed to upload logs")
+		errz = append(errz, errors.Wrapf(err, "Failed to upload logs to assisted-service"))
 	}
-	return nil
+
+	return returnVal()
+}
+
+func (c controller) collectMustGatherLogs(ctx context.Context) (string, error) {
+	tempDir, ferr := ioutil.TempDir("/tmp", "controller-must-gather-logs-")
```
tsorya, line 504: controller-must-gather-logs can be consts too and just print it in logs

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -70,7 +70,7 @@ func main() {
 	wg.Add(1)
 	go assistedController.UpdateBMHs(&wg)
 	wg.Add(1)
-	go assistedController.UploadControllerLogs(ctxLogs, &wgLogs)
+	go assistedController.UploadControllerLogs(ctxLogs, &wgLogs, &status)
```
tsorya, line 73: Maybe now it is UploadLogs ?

src/ops/ops.go
```diff
@@ -22,10 +23,19 @@ import (
...
+	Out io.Writer
+	Env []string
+	Dir string
+}
+
 //go:generate mockgen -source=ops.go -package=ops -destination=mock_ops.go
 type Ops interface {
 	ExecPrivilegeCommand(liveLogger io.Writer, command string, args ...string) (string, error)
 	ExecCommand(liveLogger io.Writer, command string, args ...string) (string, error)
+	ExecPrivilegeCommandWithContext(cmdctx *CmdContext, command string, args ...string) (string, error)
```
tsorya, line 36: I think we need to move all to use context and not to add new functions

slaviered: I have a separate PR to handle this. It a huge change and needed to be committed separately

src/ops/ops.go
```diff
@@ -463,3 +465,44 @@ func (o *ops) CreateOpenshiftSshManifest(filePath, tmpl, sshPubKeyPath string) e
 	}
 	return nil
 }
+
+func (o *ops) GetMustGatherLogs(workDir string, kubeconfigPath string) (string, error) {
+	//invoke oc adm must-gather command in the working directory
+	o.log.Infof("collecting must-gather logs into %s", workDir)
+	command := fmt.Sprintf("cd %s && oc --kubeconfig=%s adm must-gather", workDir, kubeconfigPath)
```
tsorya, line 472: can we have must-gather as consts?

src/ops/ops.go
```diff
@@ -463,3 +465,44 @@ func (o *ops) CreateOpenshiftSshManifest(filePath, tmpl, sshPubKeyPath string) e
...
+	output, err := o.ExecCommand(o.logWriter, "bash", "-c", command)
+	if err != nil {
+		return "", err
+	}
+	o.log.Info(output)
+
+	//find the directory of logs which is the output of the command
+	//this is a temp directory so we have to find it by its prefix
+	files, err := ioutil.ReadDir(workDir)
```
tsorya, line 481: can't we have this directory as not tmp? We will run this code one in the container life so not sure why it worth to have it as tmp with unique name

src/ops/ops.go
slaviered: it's less clear as const. this is self explanatory to the reader and it's actually being used just once (here). all the other places are comments or part of directory names

src/ops/ops.go
tsorya: maximum we can delete if exists

slaviered: I use tmp directory for safety reasons. all the files related to the process are located there and not in the home directory and than it can be released. the embedded directory is related to how the command works. I truly think its for the best

tsorya: but we have 4 times must-gather in different places

tsorya: 7 times

src/ops/ops.go
tsorya: but we have 7 times must-gather in different places

src/ops/ops.go
tsorya: not sure it worth it. U can have just hard coded folder name and u will not need all search operation

slaviered: The search code is not for my directory but for the directory generated by the oc must-gather command. They generate directory name with rand suffix so I am obliged to search for it by its fixed prefix anyhow, no matter what my parent directory looks like (temp or not)

src/ops/ops.go
romfreiman, line 472: @slaviered @tsorya u can use dest dir, no need to cd:
https://github.com/openshift/oc/blob/8ad72f131e0c55c93b9ceac7180f9096095f75a8/pkg/cli/admin/mustgather/mustgather.go#L60

PR #121, opened 2020-11-29
Igal/sno

tsorya:
Will add tests after review

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -205,7 +205,6 @@ func (i *installer) runBootstrap(ctx context.Context) error {
 
 func (i *installer) startBootstrap() error {
 	i.log.Infof("Running bootstrap")
-	servicesToStart := []string{"bootkube.service", "approve-csr.service", "progress.service"}
 	ignitionFileName := "bootstrap.ign"
 	ignitionPath, err := i.getFileFromService(ignitionFileName)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


src/installer/installer.go
```diff
@@ -71,41 +70,52 @@ func (i *installer) InstallNode() error {
...
-		isBootstrap = true
+
+	if isBootstrap {
 		errs.Go(func() error {
 			return i.runBootstrap(ctx)
 		})
 		i.Config.Role = string(models.HostRoleMaster)
 	}
 
+	if err := i.installNode(isSingleNodeBootstrap); err != nil {
```
eranco74, line 91: Just don't call installNode if `isSingleNodeBootstrap == true`
```suggestion
	if isSingleNodeBootstrap {
            i.log.Infof("Bootstrap in single node mode, skipping installNode")
	} else {
	    if err := i.installNode(); err != nil {
```

src/installer/installer.go
```diff
@@ -115,19 +125,47 @@ func (i *installer) InstallNode() error {
...
+	return nil
+}
+
+func (i *installer) finalizingInstallation(isBootstrap bool, isSingleNodeBootstrap bool) error {
+
+	stage := models.HostStageRebooting
+	operation := i.ops.Reboot
+	if isSingleNodeBootstrap {
+		stage = models.HostStageDone
+		operation = i.ops.Shutdown
```
eranco74, line 158: I think it's safer to have no-op operation in case of single node

tsorya: possible, thought it will be better

PR #128, opened 2020-12-06
Use quay.io/app-sre/golang for build container

YuviGold:
To avoid docker hub limitations


YuviGold:
Images weren't changed, just their base image to be from quay.io registry

/remove-label downstream-change-needed

Dockerfile.assisted-installer-controller-ocp
```diff
@@ -1,4 +1,4 @@
-FROM bash
+FROM quay.io/app-sre/centos:8
```
eranco74, line 1: @yevgeny-shnaidman  seems that this should be golang:1.14.3 too.

eranco74 (commented):
/lgtm

PR #133, opened 2020-12-07
MGMT-3179 OVNKubernetes Patch for IPv6

ori-amizur:
This change id for OVNKubernetes network type only
When starting network with OVNKubernetes netwkork type, the operator requires
that all replicas of controlPlane, which are actually the masters, will be up.
Since one of the Assisted Installer masters is actually the bootstrap when the network is started,
this network type cannot be started.
This patch changes temporarily the number of replicas to 2, to let the network start,
and then returns it to 3.

/cc @eranco74 
/cc @tsorya


eranco74:
/lgtm
As long we agree this is a temporary patch until the OVNKubernetes will add support for starting with 2 nodes.

eranco74 (commented):
/lgtm

PR #134, opened 2020-12-09
OCPBUGSM-21420 - fix race condition in day2 on ocp for controller che…

yevgeny-shnaidman:
…ck ignition pull


filanov:
How is that related to a race condition and how does it fix it?

yevgeny-shnaidman:
> How is that related to a race condition and how does it fix it?

Described in the ticket:
Description of problem:

day2 on ocp node may fail to execute install command due to the following race condition:

1) node is booted with discovery agent

2) node executes check api vip validations, which requires it to try to pull ignition from the MCS on OCP cluster

3) if the validation was successfull, then an entry with the ip of the node was created in the MCS log

4) node reaches "known" state in the Assisted Service

5) user issues install command

6) Assisted Service moves node into the installing state

7) assisted-controller polls MCS logs, finds node's entry and sends UpdateProgress to Configuring

8) assisted services updates node state to Configuring

9) node issues NextStep command

10) since node in assisted-service is in Configuring state, no commands are sent to the agent and node fails to install

 

Solution:

polling the logs should be based not only on IP but also on the User-Agent. In case the agent is ignition, and not api vip validation , it will be logged in the MCS log

filanov:
what part of the commit or the code is explaining it? if i do git show i will never understand why it was done and why it solve some race.

filanov:
IMO this explanation is exactly what need to be added to commit message.

PR #135, opened 2020-12-10
MGMT-2662 Handle the case when MetalProvisioning exists

asalkeld:
In the current situation (no provisioning resource)
- don't change anything

In the new situation of having a provisioning resource
- set host.Spec.ExternallyProvisioned=true
- don't move the metal3v1alpha1.StatusAnnotation
- remove the metal3v1alpha1.PausedAnnotation (set on masters by the installer)
- continue setting the controllerRef

Provisioning is enabled by https://github.com/openshift/assisted-service/pull/709 when we have 4.7 and above.
This can be merged now as it is only used when the PR above is merged.

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -275,18 +275,9 @@ func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
 	defer wg.Done()
 	for {
 		time.Sleep(GeneralWaitInterval)
-		exists, err := c.kc.IsMetalProvisioningExists()
-		if err != nil {
-			continue
-		}
-		if err == nil && exists {
-			c.log.Infof("Provisioning CR exists, no need to update BMHs")
-			return
-		}
-
 		bmhs, err := c.kc.ListBMHs()
 		if err != nil {
-			c.log.WithError(err).Errorf("Failed to BMH hosts")
+			c.log.WithError(err).Errorf("Failed to list BMH hosts")
 			continue
 		}
 
@@ -294,13 +285,13 @@ func (c controller) UpdateBMHs(wg *sync.WaitGroup) {
 
 		machines, err := c.unallocatedMachines(bmhs)
 		if err != nil {
-			c.log.WithError(err).Errorf("Failed to get machines")
+			c.log.WithError(err).Errorf("Failed to find unallocated machines")
 			continue
 		}
 
 		c.log.Infof("Number of unallocated Machines is %d", len(machines.Items))
 
-		allUpdated := c.updateBMHs(bmhs, machines)
+		allUpdated := c.updateBMHs(&bmhs, machines)
 		if allUpdated {
 			c.log.Infof("Updated all the BMH CRs, finished successfully")
 			return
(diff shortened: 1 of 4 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -305,43 +296,75 @@ func (c controller) unallocatedMachines(bmhList metal3v1alpha1.BareMetalHostList
...
+		return false
+	}
+
 	for i := range bmhList.Items {
 		bmh := bmhList.Items[i]
 		c.log.Infof("Checking bmh %s", bmh.Name)
-		annotations := bmh.GetAnnotations()
-		content := []byte(annotations[metal3v1alpha1.StatusAnnotation])
+		needsUpdate := !bmh.Spec.ExternallyProvisioned
+		bmh.Spec.ExternallyProvisioned = true
```
zaneb, line 353: It might pay to check that the provisioning state is None or Unmanaged before forcing this on.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -305,43 +296,75 @@ func (c controller) unallocatedMachines(bmhList metal3v1alpha1.BareMetalHostList
...
 }
 
-func (c controller) updateBMHs(bmhList metal3v1alpha1.BareMetalHostList, machineList *mapiv1beta1.MachineList) bool {
+func (c controller) updateBMHAnnotations(bmh *metal3v1alpha1.BareMetalHost, provisioningExists bool) (bool, error) {
+	annotations := bmh.GetAnnotations()
+	content := []byte(annotations[metal3v1alpha1.StatusAnnotation])
+
+	if provisioningExists {
+		// when baremetal operator is enabled, we need to remove the pausedAnnotation
+		// to indicated that the statusAnnotation is available. This is only on
```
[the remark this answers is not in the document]
asalkeld: it will allways exist

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -305,43 +296,84 @@ func (c controller) unallocatedMachines(bmhList metal3v1alpha1.BareMetalHostList
...
+		t := metav1.Now()
+		bmh.Status.LastUpdated = &t
+	}
+	err = c.kc.UpdateBMHStatus(bmh)
+	if err != nil {
+		c.log.WithError(err).Errorf("Failed to update status of BMH %s", bmh.Name)
+		return false, err
+	}
+	delete(annotations, metal3v1alpha1.StatusAnnotation)
+	return true, nil
```
zaneb, line 338: Updating the Status changes the version of the resource, so an immediate write to change the annotations will always fail (it'll work the next time with a fresh copy of the resource though). This appears to be the case here, although it pre-dates this patch.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -305,43 +296,84 @@ func (c controller) unallocatedMachines(bmhList metal3v1alpha1.BareMetalHostList
...
-			if err != nil {
-				c.log.WithError(err).Errorf("Failed to update status of BMH %s", bmh.Name)
+		needsUpdate := false
+		if provisioningExists {
+			if bmh.Status.Provisioning.State == metal3v1alpha1.StateNone || bmh.Status.Provisioning.State == metal3v1alpha1.StateUnmanaged {
+				needsUpdate = !bmh.Spec.ExternallyProvisioned
+				bmh.Spec.ExternallyProvisioned = true
+			} else {
+				c.log.Infof("bmh %s Provisioning.State=%s", bmh.Name, bmh.Status.Provisioning.State)
+				allUpdated = false
```
zaneb, line 360: I don't think you want to return false in this case - if the Host isn't in one of those states it's never going to be (the state machine doesn't move back in that direction), so we'll be stuck in an infinite loop.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -334,41 +325,97 @@ func (c controller) unallocatedMachines(bmhList metal3v1alpha1.BareMetalHostList
...
-			continue
-		}
-		allUpdated = false
-		if annotations[metal3v1alpha1.StatusAnnotation] != "" {
-			c.log.Infof("Updating annotations for bmh %s", bmh.Name)
-			objStatus, err := c.unmarshalStatusAnnotation(content)
-			if err != nil {
-				c.log.WithError(err).Errorf("Failed to unmarshal status annotation of %s", bmh.Name)
+		needsUpdate := false
+		if provisioningExists {
```
tsorya, line 379: it looks like we have same "if provisioningExists" can't we move this part to the one that is on line 388 ? or vice versa.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -334,41 +325,97 @@ func (c controller) unallocatedMachines(bmhList metal3v1alpha1.BareMetalHostList
...
+
+func (c controller) updateBMHs(bmhList *metal3v1alpha1.BareMetalHostList, machineList *mapiv1beta1.MachineList) bool {
 	allUpdated := true
 
+	provisioningExists, err := c.kc.IsMetalProvisioningExists()
+	if err != nil {
+		return false
+	}
+
 	for i := range bmhList.Items {
```
tsorya, line 374: very large loop with many conditions.
Can you split it to functions ?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -334,41 +325,97 @@ func (c controller) unallocatedMachines(bmhList metal3v1alpha1.BareMetalHostList
...
 			if err != nil {
-				c.log.WithError(err).Errorf("Failed to update status of BMH %s", bmh.Name)
+				c.log.WithError(err).Errorf("Failed to copy BMH Annotations to Status %s", bmh.Name)
+				allUpdated = false
 				continue
 			}
-			delete(annotations, metal3v1alpha1.StatusAnnotation)
+			if statusUpdated {
+				// refresh the BMH after the StatusUpdate to get the new generation
+				bmhPtr, err := c.kc.GetBMH(bmh.Name)
```
eranco74, line 408: Add logging

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -334,41 +325,97 @@ func (c controller) unallocatedMachines(bmhList metal3v1alpha1.BareMetalHostList
...
+			// to indicated that the statusAnnotation is available. This is only on
+			// the master nodes.
+			if _, ok := annotations[metal3v1alpha1.PausedAnnotation]; ok {
+				c.log.Infof("Removing pausedAnnotation from BMH host %s", bmh.Name)
+				delete(annotations, metal3v1alpha1.PausedAnnotation)
+				needsUpdate = true
 			}
-			err = c.kc.UpdateBMHStatus(&bmh)
+		} else {
+			statusUpdated, err := c.copyBMHAnnotationsToStatus(&bmh, provisioningExists)
```
eranco74, line 400: Pass the annotations to copyBMHAnnotationsToStatus instead of getting it again in line 329

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -334,41 +325,97 @@ func (c controller) unallocatedMachines(bmhList metal3v1alpha1.BareMetalHostList
...
+func (c controller) unmarshalStatusAnnotation(content []byte) (*metal3v1alpha1.BareMetalHostStatus, error) {
+	bmhStatus := &metal3v1alpha1.BareMetalHostStatus{}
+	err := json.Unmarshal(content, bmhStatus)
+	if err != nil {
+		return nil, err
+	}
+	return bmhStatus, nil
+}
+
+func (c controller) updateBMHs(bmhList *metal3v1alpha1.BareMetalHostList, machineList *mapiv1beta1.MachineList) bool {
```
eranco74, line 366: Consider extracting some of the logic to a separate method

PR #99, opened 2020-11-12
MGMT-2573 Use the mco image passed from assisted-service

sacharya:

PR #136, opened 2020-12-13
MGMT-1739 Remove MCO hardcoded hashmap

YuviGold:
Complete [MGMT-2573](https://github.com/openshift/assisted-installer/pull/99) and
remove old code, while MCO image is already received as a parameter
(extracted on assisted-service)


YuviGold:
This PR is required in order to support the extraction of the OpenShift version on https://github.com/openshift/assisted-test-infra/pull/328
Otherwise, there is a failure:
```
Failed - openshift version 4.7 is not supported
```

PR #137, opened 2020-12-13
NO-ISSUE: Adding the error to logs when extracting ignition to disk.

ybettan:
Signed-off-by: Yoni Bettan <ybettan@redhat.com>


YuviGold (approved):
LGTM

PR #147, opened 2020-12-17
MGMT-3298: enable MCD to use 'systemd start rpm-ostreed' while running as a container

osherdp:
Combination of ``--pid=host`` and ``--privilege`` should be enough for it to work


romfreiman:
please open a blocker jira ticket and add it to the commit message

eranco74 (commented):
/lgtm

PR #153, opened 2020-12-24
MGMT-3063 Add Env variable to enable/disable CVO wait 

nmagnezi:
To be followed up with a proper config flag.

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -53,6 +53,7 @@ type ControllerConfig struct {
 	Namespace            string `envconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
 	OpenshiftVersion     string `envconfig:"OPENSHIFT_VERSION" required:"true"`
 	HighAvailabilityMode string `envconfig:"HIGH_AVAILABILITY_MODE" required:"false" default:"Full"`
+	CheckClusterVersion  bool   `envconfig:"CHECK_CLUSTER_VERSION" required:"false" default:"false"`
 }
 type Controller interface {
 	WaitAndUpdateNodesStatus(status *ControllerStatus)
@@ -234,10 +235,13 @@ func (c controller) PostInstallConfigs(wg *sync.WaitGroup, status *ControllerSta
 }
 
 func (c controller) postInstallConfigs() error {
+	var err error
 
-	err := c.waitForClusterVersion()
-	if err != nil {
-		return err
+	if c.CheckClusterVersion {
+		err = c.waitForClusterVersion()
+		if err != nil {
+			return err
+		}
 	}
 
 	err = utils.WaitForPredicate(WaitTimeout, GeneralWaitInterval, c.addRouterCAToClusterCA)
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -45,15 +45,15 @@ var WaitTimeout = 2 * time.Hour
 // as a first step it will wait till nodes are added to cluster and update their status to Done
 
 type ControllerConfig struct {
-	ClusterID            string `envconfig:"CLUSTER_ID" required:"true" `
-	URL                  string `envconfig:"INVENTORY_URL" required:"true"`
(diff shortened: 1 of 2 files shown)


nmagnezi:
/hold
testing

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -314,9 +314,10 @@ var _ = Describe("installer HostRoleMaster role", func() {
 
 	Context("validating AddRouterCAToClusterCA", func() {
 		conf := ControllerConfig{
-			ClusterID:        "cluster-id",
-			URL:              "https://assisted-service.com:80",
-			OpenshiftVersion: "4.7",
+			ClusterID:           "cluster-id",
+			URL:                 "https://assisted-service.com:80",
+			OpenshiftVersion:    "4.7",
+			CheckClusterVersion: true,
```
ronniel1, line 320: Shouldn't we also check a case where this is false?

YuviGold:
@yevgeny-shnaidman Does it replace the code in test-infra?

yevgeny-shnaidman:
> @yevgeny-shnaidman Does it replace the code in test-infra?

more like changes it: now instead of waiting to use wait-for-cvo flag after installation, we need to pass it as part of the assisted-service deployment

src/assisted_installer_controller/assisted_installer_controller_test.go
nmagnezi: Yes, we do.
Done.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -52,6 +52,7 @@ type ControllerConfig struct {
 	CACertPath           string `envconfig:"CA_CERT_PATH" required:"false" default:""`
 	Namespace            string `envconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
 	OpenshiftVersion     string `envconfig:"OPENSHIFT_VERSION" required:"true"`
+	CheckClusterVersion  bool   `envconfig:"CHECK_CLUSTER_VERSION" required:"false" default:"false"`
```
gamli75, line 55: I think the name is a little bit confusing and not clear - maybe it should be called: WAIT_FOR_CLUSTER_VERSION

gamli75 (commented):
Maybe we should add it to openshift template so we will be able to change it without changing the code - @ronniel1 WDYT?

src/assisted_installer_controller/assisted_installer_controller.go
nmagnezi: np problem, I'll change that.

nmagnezi:
> Maybe we should add it to openshift template so we will be able to change it without changing the code - @ronniel1 WDYT?

@gamli75 the plan to have this env variable in assisted-service configMap, forward it to assisted-installer to render the controller ConfigMap with the same value. 
Will happen in follow-up PRs.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -45,13 +45,14 @@ var WaitTimeout = 2 * time.Hour
...
-	Namespace            string `envconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
-	OpenshiftVersion     string `envconfig:"OPENSHIFT_VERSION" required:"true"`
+	ClusterID             string `envconfig:"CLUSTER_ID" required:"true" `
+	URL                   string `envconfig:"INVENTORY_URL" required:"true"`
+	PullSecretToken       string `envconfig:"PULL_SECRET_TOKEN" required:"true"`
+	SkipCertVerification  bool   `envconfig:"SKIP_CERT_VERIFICATION" required:"false" default:"false"`
+	CACertPath            string `envconfig:"CA_CERT_PATH" required:"false" default:""`
+	Namespace             string `envconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
+	OpenshiftVersion      string `envconfig:"OPENSHIFT_VERSION" required:"true"`
+	waitForClusterVersion bool   `envconfig:"WAIT_FOR_CLUSTER_VERSION" required:"false" default:"false"`
```
tsorya, line 55: upper case

nmagnezi: yup

PR #155, opened 2020-12-28
Mgmt 2971 Add single node installation flow to assisted-installer 

eranco74:


src/installer/installer.go
```diff
@@ -127,6 +130,20 @@ func (i *installer) InstallNode() error {
 	return nil
 }
 
+func (i *installer) writeImageToDisk(err error, ignitionPath string) error {
+	i.UpdateHostInstallProgress(models.HostStageWritingImageToDisk, "")
+	err = utils.Retry(3, time.Second, i.log, func() error {
```
osherdp, line 135: What is ``time.Second``?
Can we use keyword arguments here?

src/installer/installer.go
```diff
@@ -127,6 +130,20 @@ func (i *installer) InstallNode() error {
...
+		return i.ops.WriteImageToDisk(ignitionPath, i.Device, i.inventoryClient, i.Config.InstallerArgs)
+	})
+	if err != nil {
+		i.log.Errorf("Failed to write image to disk %s", err)
+		return err
+	} else {
```
osherdp, line 141: Delete else clause

osherdp (changes requested):

PR #157, opened 2020-12-28
MGMT-3401 - collect oc log immediately on error

slaviered:


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -600,12 +601,28 @@ func (c *controller) UploadLogs(ctx context.Context, wg *sync.WaitGroup, status
 				}
 				podName = pods[0].Name
 			}
-			err := common.UploadPodLogs(c.kc, c.ic, c.ClusterID, podName, c.Namespace, controllerLogsSecondsAgo, c.log)
+			var err error
+			if errorDetected != status.HasError() {
+				//once error is detected in the flow, upload the full capacity of debugging logs
+				//this is required only once, upon the detection of a fatal error
+				errorDetected = status.HasError()
+				_ = c.uploadSummaryLogs(podName, c.Namespace, controllerLogsSecondsAgo, errorDetected)
```
YuviGold, line 609: Please check the return error and log if necessary. Errors should never pass silently

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -600,12 +601,28 @@ func (c *controller) UploadLogs(ctx context.Context, wg *sync.WaitGroup, status
 				}
 				podName = pods[0].Name
 			}
-			err := common.UploadPodLogs(c.kc, c.ic, c.ClusterID, podName, c.Namespace, controllerLogsSecondsAgo, c.log)
+			var err error
+			if errorDetected != status.HasError() {
+				//once error is detected in the flow, upload the full capacity of debugging logs
```
YuviGold, line 606: Add whitespace between the slashes and the comment string

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -600,12 +601,28 @@ func (c *controller) UploadLogs(ctx context.Context, wg *sync.WaitGroup, status
...
+				//this is required only once, upon the detection of a fatal error
+				errorDetected = status.HasError()
+				_ = c.uploadSummaryLogs(podName, c.Namespace, controllerLogsSecondsAgo, errorDetected)
+				c.log.Infof("Uploading controller and cluster logs on error ...")
+				return
+			} else if !errorDetected {
+				//on normal flow, keep updating the controller log output every 5 minutes
+				err = common.UploadPodLogs(c.kc, c.ic, c.ClusterID, podName, c.Namespace, controllerLogsSecondsAgo, c.log)
+			} else {
+				//when error state persists, do nothing.
```
YuviGold, line 616: Add log

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -577,6 +577,7 @@ func (c *controller) UploadLogs(ctx context.Context, wg *sync.WaitGroup, status
 	c.log.Infof("Start sending logs")
 	podName := ""
 	ticker := time.NewTicker(LogsUploadPeriod)
+	var errorDetected bool = false
```
YuviGold, line 580: `errorDetected` is very generic. Maybe `isControllerErrorDetected`?

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: it must in this case, because we have nothing to do with it. The controller is about to die, the logs will never be uploaded again so there is no point to log anything here. Inside the flow there are ample logs

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: Done

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: It was meant to be generic because at this point I do not know what is the source of the error. And it is not a controller error but an error in the cluster state. Therefore I think the short version is better and was designed to be generic.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -600,12 +600,30 @@ func (c *controller) UploadLogs(ctx context.Context, wg *sync.WaitGroup, status
...
+				errorDetected = status.HasError()
+				c.log.Infof("Start uploading controller and cluster logs on error ...")
+				_ = c.uploadSummaryLogs(podName, c.Namespace, controllerLogsSecondsAgo, errorDetected)
+				c.log.Infof("Done uploading controller and cluster logs on error")
+			} else if !errorDetected {
+				//on normal flow, keep updating the controller log output every 5 minutes
+				c.log.Infof("Start uploading controller logs (intermediate snapshot) ...")
+				err = common.UploadPodLogs(c.kc, c.ic, c.ClusterID, podName, c.Namespace, controllerLogsSecondsAgo, c.log)
+				c.log.Infof("Done uploading controller logs")
+			} else {
```
tsorya, line 617: why just not to return?not sure i understand why we need this else

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -600,12 +600,30 @@ func (c *controller) UploadLogs(ctx context.Context, wg *sync.WaitGroup, status
 				}
 				podName = pods[0].Name
 			}
-			err := common.UploadPodLogs(c.kc, c.ic, c.ClusterID, podName, c.Namespace, controllerLogsSecondsAgo, c.log)
+			var err error
+			c.log.Info("Collecting Logs ....")
+			if errorDetected != status.HasError() {
```
tsorya, line 605: sorry not sure i understand the logic. 
I thought we will have something more clear. 
if  status.HasError() {
  upload all logs and exit
}

tsorya: i think even better to just set ctx to Done and exit as expected without adding all this logic.
if status.HasError() {
  set done and let regular flow and exit
}

slaviered: It's possible and certainly ends up in more concise code, but it means that we will not have other logs from the routines that still waits. Is this OK?

slaviered: Also, In case of an error the context will be deleted twice (once from within UploadLogs and the second one from main). The second attempt will panic

src/assisted_installer_controller/assisted_installer_controller.go
YuviGold: So maybe let's change it to `clusterErrorDetected`? keeping it as an "error" can mean there was an error even in this function. I would like it to refer to where the error is coming from, even if it can only eliminate some trivial places.

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: Logic changed. This remark was resolved as part of the new flow

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: logic has changed. errorDetected var has been removed

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: Fixed

PR #159, opened 2021-01-04
MGMT-3297 - support static ips when rebooting nodes with RHCOS image

yevgeny-shnaidman:


src/installer/installer.go
```diff
@@ -588,3 +600,20 @@ func (i *installer) createSingleNodeMasterIgnition() (string, error) {
...
+	i.log.Infof("Start checking localhostname")
+	hostname, err := i.ops.GetHostname()
+	if err != nil {
+		i.log.Errorf("Failed to get hostname from kernel, err %s", err)
+		return err
+	}
+	if hostname != "localhost" {
+		i.log.Infof("hostname is not localhost, no need to do anything")
+		return nil
+	}
```
empovit, line 614: Can it be empty? In some weird corner case

src/ops/ops.go
```diff
@@ -177,6 +179,7 @@ func (o *ops) WriteImageToDisk(ignitionPath string, device string, progressRepor
 
 func installerArgs(ignitionPath string, device string, extra []string) []string {
 	allArgs := []string{"install", "--insecure", "-i", ignitionPath}
+	// allArgs := []string{"install", "--insecure", "-i", ignitionPath, "--copy-network"}
```
empovit, line 182: Delete please

src/ops/ops.go
```diff
@@ -503,3 +506,14 @@ func (o *ops) GetMustGatherLogs(workDir string, kubeconfigPath string) (string,
 	}
 	return path.Join(workDir, tarName), nil
 }
+
+func (o *ops) CreateRandomHostname(hostname string) error {
+	command := fmt.Sprintf("echo %s > /etc/hostname", hostname)
```
empovit, line 511: If it's supposed to run under sudo with a non-root user, then please make sure you test it. `echo` is known not to work with sudo. `sudo bash -c echo ... > file" didn't work for me either. A good way to echo to a file that requires sudo is `echo "text" | sudo tee --append /path/to/file`

src/installer/installer.go
yevgeny-shnaidman: no,we are using os.Hostname function which eventually goes through all the possible OS permutations to get the hostname: name, proc, etc', and in case there was no error during any of the system calls, the hostname should be populated

src/ops/ops.go
yevgeny-shnaidman: done

src/ops/ops.go
yevgeny-shnaidman: echo by itself indeed does not work , but running it via nsenter with bin/bas -c works just fine. Checked it manually. Also , it it was not working, then installation would have stuck

PR #160, opened 2021-01-05
MGMT-3519 Change control plane replicas patch to do the patch only when control plane replicas is 3

ori-amizur:
- This fixes it for IPv6 single node

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -35,6 +35,7 @@ const (
 	ovnKubernetes                = "OVNKubernetes"
 	networkTypeTimeoutSeconds    = 300
 	singleNodeMasterIgnitionPath = "/opt/openshift/master.ign"
+	numMasterNodes               = 3
 )
 
 var generalWaitTimeout = 30 * time.Second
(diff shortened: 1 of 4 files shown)


src/installer/installer.go
```diff
@@ -307,19 +307,28 @@ func (i *installer) waitForControlPlane(ctx context.Context, kc k8s_client.K8SCl
...
+		if origControlPlaneReplicas, err = kc.GetControlPlaneReplicas(); err != nil {
+			i.log.WithError(err).Error("Failed to get control plane replicas")
 			return err
 		}
+		if origControlPlaneReplicas == 3 {
+			if err = kc.PatchControlPlaneReplicas(); err != nil {
+				i.log.WithError(err).Error("Failed to patch control plane replicas")
+				return err
+			}
+		}
```
eranco74, line 321: Extract to a separate function

src/installer/installer.go
```diff
@@ -307,19 +307,28 @@ func (i *installer) waitForControlPlane(ctx context.Context, kc k8s_client.K8SCl
...
 	}
+	var origControlPlaneReplicas int
 	if nt == ovnKubernetes {
-		if err = kc.PatchControlPlaneReplicas(); err != nil {
-			i.log.WithError(err).Error("Failed to patch control plane replicas")
+		if origControlPlaneReplicas, err = kc.GetControlPlaneReplicas(); err != nil {
+			i.log.WithError(err).Error("Failed to get control plane replicas")
 			return err
 		}
+		if origControlPlaneReplicas == 3 {
```
eranco74, line 316: Use a const for 3.
Also, add a comment on why it's required.

ori-amizur: OK

src/installer/installer.go
ori-amizur: OK

eranco74 (commented):
/lgtm

PR #161, opened 2021-01-05
Allow installing single node when hosts role is bootstrap

eranco74:


eranco74:
This change should allow us to avoid this ugly hack
https://github.com/eranco74/assisted-service/commit/72b77aca6f8161b1735c27cc97e66588eb903b41#diff-e5c144e5eab23eac9b5947751c077100cbafd0c6d2f61059920666f6513c8ec1R101

PR #162, opened 2021-01-05
MGMT-2403 retry mechanism for oc must-gather logs

slaviered:
Report back if oc-must gater command has failed and retry
for an hour to fetch the logs before aborting completely

This will give an opportunity for the newly formed cluster
to stabilize and evetually serve the gather pod

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -39,6 +39,7 @@ var GeneralWaitInterval = generalWaitTimeoutInt * time.Second
 var generalProgressUpdateInt = 60 * time.Second
 var LogsUploadPeriod = 5 * time.Minute
 var WaitTimeout = 70 * time.Minute
+var NumRetrySendingLogs = 6
 
 // assisted installer controller is added to control installation process after  bootstrap pivot
 // assisted installer will deploy it on installation process
@@ -505,6 +506,7 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
  **/
 func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSeconds int64, isMustGatherEnabled bool) error {
 	var tarentries = make([]utils.TarEntry, 0)
+	var ok bool = true
 	ctx := utils.GenerateRequestContext()
 
 	if isMustGatherEnabled {
@@ -513,6 +515,8 @@ func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSec
 			if entry, tarerr := utils.NewTarEntryFromFile(tarfile); tarerr == nil {
 				tarentries = append(tarentries, *entry)
 			}
+		} else {
+			ok = false
 		}
 	}
 
@@ -520,6 +524,8 @@ func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSec
 	if podLogs, err := c.kc.GetPodLogsAsBuffer(namespace, podName, sinceSeconds); err == nil {
 		tarentries = append(tarentries,
 			*utils.NewTarEntry(podLogs, nil, int64(podLogs.Len()), fmt.Sprintf("%s.logs", podName)))
+	} else {
+		ok = false
 	}
 
 	if len(tarentries) == 0 {
@@ -543,9 +549,16 @@ func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSec
 	err := c.ic.UploadLogs(ctx, c.ClusterID, models.LogsTypeController, pr)
 	if err != nil {
 		utils.RequestIDLogger(ctx, c.log).WithError(err).Error("Failed to upload logs")
+		return err
 	}
 
-	return err
+	if !ok {
+		msg := "Some Logs were not collected in summary"
+		c.log.Errorf(msg)
+		return errors.New(msg)
+	}
+
+	return nil
 }
 
 func (c controller) collectMustGatherLogs(ctx context.Context) (string, error) {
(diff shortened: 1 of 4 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -585,9 +598,19 @@ func (c *controller) UploadLogs(ctx context.Context, cancellog context.CancelFun
...
 		select {
 		case <-ctx.Done():
+			var retry = 0
+			var err error
 			if podName != "" {
 				c.log.Infof("Upload final controller and cluster logs before exit")
-				_ = c.uploadSummaryLogs(podName, c.Namespace, controllerLogsSecondsAgo, status.HasError())
+				for {
+					err = c.uploadSummaryLogs(podName, c.Namespace, controllerLogsSecondsAgo, status.HasError())
+					if err == nil || retry >= NumRetrySendingLogs {
```
tsorya, line 607: no need, retry till it is running, why matters ?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -543,9 +549,16 @@ func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSec
 	err := c.ic.UploadLogs(ctx, c.ClusterID, models.LogsTypeController, pr)
 	if err != nil {
 		utils.RequestIDLogger(ctx, c.log).WithError(err).Error("Failed to upload logs")
+		return err
 	}
 
-	return err
+	if !ok {
+		msg := "Some Logs were not collected in summary"
```
tsorya, line 556: return error and print logs in UploadLogs , or log all the errors here

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -505,6 +506,7 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
  **/
 func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSeconds int64, isMustGatherEnabled bool) error {
 	var tarentries = make([]utils.TarEntry, 0)
+	var ok bool = true
```
tsorya, line 509: maybe better to change if lines :
tarfile, err := c.collectMustGatherLogs(ctx)
if err == nil {
			if entry, tarerr := utils.NewTarEntryFromFile(tarfile); tarerr == nil {
				tarentries = append(tarentries, *entry)
			}
		}
and then to return err and will be no need to have this bool?

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: why not to use WaitForPredicate ? and just create local function as 
func() bool {
   err = c.uploadSummaryLogs(podName, c.Namespace, controllerLogsSecondsAgo, status.HasError())
   if err != nil {
   log.smth
}
return err == nil
}

slaviered: 10 seconds is too little. Even when it fails, the command takes a long time. I choose 5 minutes because it's the existing interval for collecting controller logs and it fits well with the times I see for the cluster to get ready

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: No. I wanted the flow to be forgiving making the most of the logs collection and not stopping if one of the component fails. However, if one of them fails, I'd like to retry and therefore at the end, I return error to activate the retry mechanism

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: I will use the predicate but as for the infinite retry I don't think it is good idea. we do want the controller to die at some point and not block forever. There is no guarantee that the oc makes it at all in some edge cases

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: this is what i meant, i just think this var is not relevant , just take err and return it at the end

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -604,24 +623,21 @@ func (c *controller) UploadLogs(ctx context.Context, cancellog context.CancelFun
...
 			}
-			var err error
-			c.log.Info("Collecting Logs ....")
+
 			if status.HasError() {
-				//once error is detected in the flow, upload the full capacity of debugging logs
-				//and abort
-				c.log.Infof("Error detected... closing logs...")
+				c.log.Infof("Error detected. Closing logs and aborting...")
+				ticker.Stop()
```
tsorya, line 629: why we need ticker stop?

slaviered:
there are multiple errors possible , so I can not use err (it might be
overridden by all the calls that returned err) this is why I needed another
flag that once is raised, not get overridden

slaviered:
For good measures to make sure it will not get selected again in the outer
select statement

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: For good measures to make sure it will not get selected again in the outer
select statement

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: agree with u

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -587,7 +600,13 @@ func (c *controller) UploadLogs(ctx context.Context, cancellog context.CancelFun
 		case <-ctx.Done():
 			if podName != "" {
 				c.log.Infof("Upload final controller and cluster logs before exit")
-				_ = c.uploadSummaryLogs(podName, c.Namespace, controllerLogsSecondsAgo, status.HasError())
+				_ = utils.WaitForPredicate(WaitTimeout, LogsUploadPeriod, func() bool {
```
tsorya, line 603: looks much prettier as for me :)

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: no need. till you are in case it will not be selected

PR #169, opened 2021-01-12
NO-ISSUE: Work to get images built in prow

djzager:
See https://github.com/openshift/release/pull/14728

### DIFF
--- a/Dockerfile.assisted-installer
+++ b/Dockerfile.assisted-installer
@@ -1,16 +1,19 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.15 AS builder
 ENV GOFLAGS=-mod=mod
 WORKDIR /go/src/github.com/openshift/assisted-installer
+
+# Bring in the go dependencies before anything else so we can take
+# advantage of caching these layers in future builds.
+COPY go.mod go.mod
+COPY go.sum go.sum
+RUN go mod download
+
 COPY . .
 RUN make installer
 
 FROM registry.svc.ci.openshift.org/ocp/4.7:base
-ARG GIT_REVISION
-ARG QUAY_TAG_EXPIRATION
 
 COPY --from=builder /go/src/github.com/openshift/assisted-installer/build/installer /usr/bin/installer
 COPY --from=builder /go/src/github.com/openshift/assisted-installer/deploy/assisted-installer-controller /assisted-installer-controller
 
-LABEL "git_revision"=${GIT_REVISION}
-LABEL "quay.expires-after"=${QUAY_TAG_EXPIRATION}
 ENTRYPOINT ["/usr/bin/installer"]
--- a/Dockerfile.assisted-installer
+++ b/Dockerfile.assisted-installer
@@ -1,8 +1,16 @@
-FROM quay.io/centos/centos:centos8
-copy build/installer /
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
+ENV GOFLAGS=-mod=mod
+WORKDIR /go/src/github.com/openshift/assisted-installer
+COPY . .
+RUN make installer
+
+FROM registry.svc.ci.openshift.org/ocp/4.7:base
 ARG GIT_REVISION
-LABEL "git_revision"=${GIT_REVISION}
 ARG QUAY_TAG_EXPIRATION
+
+COPY --from=builder /go/src/github.com/openshift/assisted-installer/build/installer /usr/bin/installer
+COPY --from=builder /go/src/github.com/openshift/assisted-installer/deploy/assisted-installer-controller /assisted-installer-controller
+
+LABEL "git_revision"=${GIT_REVISION}
 LABEL "quay.expires-after"=${QUAY_TAG_EXPIRATION}
-add deploy/assisted-installer-controller /assisted-installer-controller/deploy
-ENTRYPOINT ["/installer"]
+ENTRYPOINT ["/usr/bin/installer"]
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ NAMESPACE := $(or ${NAMESPACE},assisted-installer)
 GIT_REVISION := $(shell git rev-parse HEAD)
 PUBLISH_TAG := $(or ${GIT_REVISION})
 
-all: image image_controller image_controller_ocp unit-test
+all: build-images unit-test
 
 lint:
 	golangci-lint run -v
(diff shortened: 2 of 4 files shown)


Dockerfile.assisted-installer
```diff
@@ -1,8 +1,16 @@
-FROM quay.io/centos/centos:centos8
-copy build/installer /
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
+ENV GOFLAGS=-mod=mod
+WORKDIR /go/src/github.com/openshift/assisted-installer
+COPY . .
+RUN make installer
+
+FROM registry.svc.ci.openshift.org/ocp/4.7:base
```
YuviGold, line 7: What's wrong with `quay.io/centos/centos:centos8` ? :)

Dockerfile.assisted-installer
```diff
@@ -1,8 +1,16 @@
-FROM quay.io/centos/centos:centos8
-copy build/installer /
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
```
YuviGold, line 1: I see we're using mostly `registry.svc.ci.openshift.org/openshift/release:golang-1.15` on other places. Is there a difference?

On [assisted-service](https://github.com/openshift/assisted-service/blob/33090c305040b9202c7f84b4be9489568e79c84b/Dockerfile.assisted-service#L18-L24) for example. But then we empty `GOFLAGS`

Dockerfile.assisted-installer
```diff
@@ -1,8 +1,16 @@
...
+ENV GOFLAGS=-mod=mod
+WORKDIR /go/src/github.com/openshift/assisted-installer
+COPY . .
```
YuviGold, line 4: Maybe add only the required files (`src`, `go.mod`, `go.sum`) to reduce the image size and make caching work better. Alternatively, add a `.dockerignore` file

Dockerfile.assisted-installer
```diff
@@ -1,8 +1,16 @@
...
+
+FROM registry.svc.ci.openshift.org/ocp/4.7:base
 ARG GIT_REVISION
-LABEL "git_revision"=${GIT_REVISION}
 ARG QUAY_TAG_EXPIRATION
+
+COPY --from=builder /go/src/github.com/openshift/assisted-installer/build/installer /usr/bin/installer
+COPY --from=builder /go/src/github.com/openshift/assisted-installer/deploy/assisted-installer-controller /assisted-installer-controller
+
+LABEL "git_revision"=${GIT_REVISION}
```
YuviGold, line 14: iirc, Prow doesn't work well with Dockerfile arguments.
You can Remove the labels and pass them as part of the build process, without arguments.

For example on [assisted-service](https://github.com/openshift/assisted-service/blob/7aa360c9ac4ad66c1afddaf20b18a484a1070268/Makefile#L31)

Dockerfile.assisted-installer
```diff
@@ -1,8 +1,16 @@
...
+COPY . .
+RUN make installer
+
+FROM registry.svc.ci.openshift.org/ocp/4.7:base
 ARG GIT_REVISION
-LABEL "git_revision"=${GIT_REVISION}
 ARG QUAY_TAG_EXPIRATION
+
+COPY --from=builder /go/src/github.com/openshift/assisted-installer/build/installer /usr/bin/installer
+COPY --from=builder /go/src/github.com/openshift/assisted-installer/deploy/assisted-installer-controller /assisted-installer-controller
```
YuviGold, line 12: I think that the controller isn't needed, it is part of the 2nd image. no?

Makefile
```diff
@@ -10,7 +10,7 @@ NAMESPACE := $(or ${NAMESPACE},assisted-installer)
 GIT_REVISION := $(shell git rev-parse HEAD)
 PUBLISH_TAG := $(or ${GIT_REVISION})
 
-all: image image_controller image_controller_ocp unit-test
```
YuviGold, line 13: Please run `lint` and `format`

Makefile
```diff
@@ -10,7 +10,7 @@ NAMESPACE := $(or ${NAMESPACE},assisted-installer)
...
+all: build-images unit-test
 
 lint:
```
YuviGold, line 15: I think `lint` can include `format-check`. I don't see anyone calling it directly. Most of the time you would like one target to verify everything.

Makefile
```diff
@@ -29,34 +29,35 @@ unit-test: $(REPORTS)
...
 
-push_controller: image_controller
-	$(CONTAINER_COMMAND) push $(CONTROLLER)
-
-image_controller_ocp: build/controller_ocp
+controller-ocp-image:
 	GIT_REVISION=${GIT_REVISION} $(CONTAINER_COMMAND) build --build-arg GIT_REVISION -f Dockerfile.assisted-installer-controller-ocp . -t $(CONTROLLER_OCP)
 
-push_controller_ocp: image_controller_ocp
+push: installer-image
```
YuviGold, line 54: Let's keep the same naming convention. `push-installer`

YuviGold (commented):
Looks good! Some comments from previous experience when building assisted-service in Prow https://github.com/openshift/assisted-service/pull/86

Dockerfile.assisted-installer
djzager: I think this was a copy/paste error on my part. Will go back to `registry.svc.ci.openshift.org/openshift/release:golang-1.15` or have an answer as to why ;)

Dockerfile.assisted-installer
djzager: these `${PROJECT_ROOT}/deploy/assisted-installer-controller` are the deployment manifests for the controller right?

YuviGold: You are probably right. From the name, I thought it's an executable.

Makefile
djzager: We settled this, I believe, in our discussion offline and the openshift/release PR. I went ahead and added both `lint` and `format-check` to the `all` target so, by default, everything gets checked.

Dockerfile.assisted-installer
djzager: My reason for choosing this was to be more in-line with what we are doing downstream https://pkgs.devel.redhat.com/cgit/containers/assisted-installer/tree/Dockerfile?h=rhaos-4.6-rhel-8 
https://pkgs.devel.redhat.com/cgit/containers/assisted-installer-agent/tree/Dockerfile?h=rhaos-4.6-rhel-8

I know I have at least heard these debates rage on about which image to use. Any specific reason to avoid ocp 4.7 base? Should I pull in others to get feedback.

djzager:
/hold cancel

We may want to provide the necessary credentials to the jenkins server to be able to build images from registry.svc.ci.openshift.org.

YuviGold:
/retitle MGMT-3316 Work to get images built in prow

PR #172, opened 2021-01-13
MGMT-3612 Assisted-insatller should start the patch.service in case o…

tsorya:
…f single node


src/installer/installer.go
```diff
@@ -232,6 +233,8 @@ func (i *installer) startBootstrap() error {
 		if err != nil {
 			return err
 		}
+	} else {
+		servicesToStart = append(servicesToStart, "patch.service")
```
eranco74, line 237: Add a TODO for removing this once SNO is supported

eranco74 (commented):
/lgtm

PR #178, opened 2021-01-17
Bugfix set controller deployment files on /assisted-installer-controller/deploy

YuviGold:
Fixes https://github.com/openshift/assisted-installer/pull/169 changing controller deployment files to the correct directory.


YuviGold:
/retitle MGMT-3316 Bugfix set controller deployment files on /assisted-installer-controller/deploy

PR #190, opened 2021-01-24
MGTM-3751 retry to send controller logs from installer and recheck co…

tsorya:
…ntroller state

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -42,6 +42,10 @@ const (
 )
 
 var generalWaitTimeout = 30 * time.Second
+var waitForControllerPodInterval = 5 * time.Second
+
+// retry sending logs for 5 minutes maximum
+var retryCounterUploadControllerLogs = (5 * time.Minute).Milliseconds() / waitForControllerPodInterval.Milliseconds()
 
 // Installer will run the install operations on the node
 type Installer interface {
@@ -465,6 +469,7 @@ func (i *installer) waitForController() error {
 		controllerPod := common.GetPodInStatus(kc, assistedControllerPrefix, assistedControllerNamespace,
 			map[string]string{"job-name": assistedControllerPrefix}, v1.PodRunning, i.log)
 		if controllerPod != nil {
+
 			// uploading logs here cause we will finish installing bootstrap in couple of seconds
 			// just didn't wanted to start handling this logic in other place
 			// controller must be running for at least couple of minutes before this code so it will cover 90% of the cases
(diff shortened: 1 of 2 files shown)


src/installer/installer.go
```diff
@@ -465,23 +469,29 @@ func (i *installer) waitForController() error {
...
 			// if we will see that it is not enough, we will change the place.
 			err = common.UploadPodLogs(kc, i.inventoryClient, i.ClusterID, controllerPod.Name, assistedControllerNamespace, common.ControllerLogsSecondsAgo, i.log)
 			// if failed to upload logs, log why and continue
 			if err != nil {
-				i.log.WithError(err).Warnf("Failed to upload controller logs")
+				if retryCounterUploadControllerLogs > 0 {
+					retryCounterUploadControllerLogs--
+					controllerPod = nil
+				}
+				i.log.WithError(err).Warnf("Failed to upload controller logs, left %d retries", retryCounterUploadControllerLogs)
```
eranco74, line 484:
```suggestion
				i.log.WithError(err).Warnf("Failed to upload controller logs, %d retries left", retryCounterUploadControllerLogs)
```

src/installer/installer.go
```diff
@@ -465,23 +469,29 @@ func (i *installer) waitForController() error {
 		controllerPod := common.GetPodInStatus(kc, assistedControllerPrefix, assistedControllerNamespace,
 			map[string]string{"job-name": assistedControllerPrefix}, v1.PodRunning, i.log)
 		if controllerPod != nil {
+
 			// uploading logs here cause we will finish installing bootstrap in couple of seconds
 			// just didn't wanted to start handling this logic in other place
 			// controller must be running for at least couple of minutes before this code so it will cover 90% of the cases
 			// if we will see that it is not enough, we will change the place.
 			err = common.UploadPodLogs(kc, i.inventoryClient, i.ClusterID, controllerPod.Name, assistedControllerNamespace, common.ControllerLogsSecondsAgo, i.log)
```
eranco74, line 477: I think this method needs some refactoring as well.
The log says that we fail to upload while we suspect that we failed to get the log from kube-apiserver.

tsorya: this method returns error, in the case we saw:
"Failed to get logs of pod assisted-installer-controller-p7pjv: Get \"https://192.168.50.122:10250/containerLogs/assisted-installer/assisted-installer-controller-p7pjv/assisted-installer-controller?sinceSeconds=3600\": open /etc/kubernetes/secrets/kube-apiserver-to-kubelet-client.crt: no such file or directory"

tsorya: than we just wrap it with one more message that is relevant for our specific case. We can though just print the error that returns from UploadPodLogs

PR #197, opened 2021-01-26
MGMT-3792 Move assisted-installer wait for joined master to look on h…

tsorya:
…ostname and not systemUUID

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -581,11 +581,11 @@ func (i *installer) updateReadyMasters(nodes *v1.NodeList, readyMasters *[]strin
 		nodeNameAndCondition[node.Name] = node.Status.Conditions
 		for _, cond := range node.Status.Conditions {
 			if cond.Type == v1.NodeReady && cond.Status == v1.ConditionTrue &&
-				!funk.ContainsString(*readyMasters, node.Status.NodeInfo.SystemUUID) {
+				!funk.ContainsString(*readyMasters, node.Name) {
 				ctx := utils.GenerateRequestContext()
 				log := utils.RequestIDLogger(ctx, i.log)
 				log.Infof("Found a new ready master node %s with id %s", node.Name, node.Status.NodeInfo.SystemUUID)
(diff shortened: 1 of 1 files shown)


src/installer/installer.go
```diff
@@ -581,11 +581,11 @@ func (i *installer) updateReadyMasters(nodes *v1.NodeList, readyMasters *[]strin
 		nodeNameAndCondition[node.Name] = node.Status.Conditions
 		for _, cond := range node.Status.Conditions {
 			if cond.Type == v1.NodeReady && cond.Status == v1.ConditionTrue &&
-				!funk.ContainsString(*readyMasters, node.Status.NodeInfo.SystemUUID) {
+				!funk.ContainsString(*readyMasters, node.Name) {
```
ohadlevy, line 584: are we assuming that name is unique at this point already?
should we warn or do anything with the fact the uuid is not unique?

eranco74: I think this is orthogonal.
We don't know what are the consequences in case the UUID is not unique.
We should look into it and If it's an issue we will add validation in the assisted-service that will block the installation.
So far the installation fails just because the assisted installer assumed the system UUID is unique.
This Pr fix that.

ohadlevy: would a combination of uuid and mac address might be better? or perhaps comparing multiple entries to identify a machine is really unique ? (name, uuid, mac, ip?)

eranco74: >are we assuming that name is unique at this point already?
Yes, we set the hostname prior to the reboot, here we get it from k8s (the node is already a part of the cluster)

tsorya: i think that no matter what we must save as key value that we correlate with assisted.
All the validations we are making in the code are with hostname but key is systemUUID for unknown reason.
I want all our checks to be consistent, if we validate hostname inside controller, we need to use hostname inside installer. If we check what host we are going to update status for with hostname , we need to save hostname as key

PR #198, opened 2021-01-27
NO-ISSUE Use error.wrap instead of creating a new error

eranco74:


src/ignition/ignition.go
```diff
@@ -29,11 +29,11 @@ func NewIgnition() *ignition {
 func (i *ignition) ParseIgnitionFile(path string) (*config31types.Config, error) {
 	configBytes, err := ioutil.ReadFile(path)
 	if err != nil {
-		return nil, errors.Errorf("error reading file %s: %v", path, err)
+		return nil, errors.Wrapf(err, "error reading file %s", err)
```
bogdando, line 32: "error reading file %s", path?

eranco74: #199

PR #200, opened 2021-01-31
MGMT-3752 controller shall print cluster operators status before exit…

tsorya:
…ing and sending last logs

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -598,6 +598,19 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
 	c.log.Infof("Done complete installation step")
 }
 
+// logClusterOperatorsStatus logging cluster operators status
+func (c controller) logClusterOperatorsStatus() {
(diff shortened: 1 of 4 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -598,6 +598,19 @@ func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
 	c.log.Infof("Done complete installation step")
 }
 
+// logClusterOperatorsStatus logging cluster operators status
+func (c controller) logClusterOperatorsStatus() {
+	operators, err := c.kc.ListClusterOperators()
```
romfreiman, line 603: Maybe describe as well?

tsorya: we decided to start with list, looks like we have all relevant info, if not will add more

eranco74: We get the status which is the relevant part of what you get in describe

eranco74 (commented):
/lgtm

PR #202, opened 2021-02-01
Igal/mgmt 3859

tsorya:


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -756,3 +757,30 @@ func (c *controller) UploadLogs(ctx context.Context, cancellog context.CancelFun
...
+			return false
+		}
+
+		_, err = c.kc.ListNodes()
+		if err != nil {
+			c.log.WithError(err).Warningf("Failed to connect to ocp cluster")
+			return false
+		}
+
+		c.log.Infof("I am ready, spawning event")
```
romfreiman, line 776: can it be more descriptive message? Think about customer/support. What does it mean ready?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -756,3 +757,30 @@ func (c *controller) UploadLogs(ctx context.Context, cancellog context.CancelFun
...
+		if _, err := c.kc.CreateEvent(c.Namespace, common.AssistedControllerIsReadyEvent,
+			"Assisted controller is ready", common.AssistedControllerPrefix); err != nil && !apierrors.IsAlreadyExists(err) {
```
romfreiman, line 778: ready - ? connected to the assisted-service?

src/installer/installer.go
```diff
@@ -466,37 +462,38 @@ func (i *installer) waitForController() error {
...
-				if retryCounterUploadControllerLogs > 0 {
-					retryCounterUploadControllerLogs--
-					controllerPod = nil
+	tickerUploadLogs := time.NewTicker(5 * time.Minute)
+	tickerWaitForController := time.NewTicker(waitForControllerPodInterval)
+	// Keep trying until we're time out or get true
+	for {
+		select {
+		case <-tickerWaitForController.C:
+			i.logControllerEvents(kc, events)
```
romfreiman, line 473: why does it called logControllerEvents? Doesnt it listen to events?
Cannot u just register to events rather than polling?

src/assisted_installer_controller/assisted_installer_controller.go
eranco74: It means it can talk to assisted-service and kube-apiserver

src/assisted_installer_controller/assisted_installer_controller.go
eranco74, line 776:
```suggestion
		_, err := c.ic.GetCluster(context.TODO())
		if err != nil {
			c.log.WithError(err).Warningf("Failed to connect to assisted service")
			return false
		}
		c.log.Infof("assisted-service is avilable")

		_, err = c.kc.ListNodes()
		if err != nil {
			c.log.WithError(err).Warningf("Failed to connect to ocp cluster")
			return false
		}

		c.log.Infof("kube-apiserver is avilable")
		c.log.Infof("Sending ready event")
```

src/common/common.go
```diff
@@ -16,7 +16,11 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-const ControllerLogsSecondsAgo = 60 * 60
+const (
+	ControllerLogsSecondsAgo       = 60 * 60
```
eranco74, line 20: I see you didn't add it in this PR, yet what is this const? can it get a better name?

src/installer/installer.go
```diff
@@ -453,7 +449,7 @@ func (i *installer) waitForBootkube(ctx context.Context) {
 }
 
 func (i *installer) waitForController() error {
-	i.log.Infof("Waiting for controller pod to start running")
+	i.log.Infof("Waiting for controller pod to be ready")
```
eranco74, line 452:
```suggestion
	i.log.Infof("Waiting for controller to be ready")
```

src/installer/installer.go
eranco74: Agree. you should use a watcher here.

src/installer/installer.go
```diff
@@ -466,37 +462,38 @@ func (i *installer) waitForController() error {
...
-			if err != nil {
-				if retryCounterUploadControllerLogs > 0 {
-					retryCounterUploadControllerLogs--
-					controllerPod = nil
+	tickerUploadLogs := time.NewTicker(5 * time.Minute)
+	tickerWaitForController := time.NewTicker(waitForControllerPodInterval)
+	// Keep trying until we're time out or get true
+	for {
+		select {
+		case <-tickerWaitForController.C:
```
eranco74, line 472: You should use a watcher/informer here instead of the ticker.
https://pkg.go.dev/k8s.io/client-go/informers#SharedInformerFactory

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: Ready is the same as in any service, it means that i am ok and ready to run.

src/common/common.go
tsorya: Have a better name? will be glad to change. It says bring controller logs since hour ago. Not sure what is the best name for it

tsorya: we can though drop it and just bring all logs

src/installer/installer.go
tsorya: as we talked in this case there is no reason to implement informer logic, this code in 99% will run only once

src/installer/installer.go
tsorya: we are logging all new events that were added since we want to understand what is going on. As i said in message above, this code will 99% run once, i don't think there is any need to create listener on code that will run for maximum 5 minutes and better to have polling. When this code is running mainly all events will be set already.

tsorya: In any case we will need to poll even if we add listener so for such case it is not worth it, it will only complicate it

src/assisted_installer_controller/assisted_installer_controller.go
romfreiman: I know. Will anyone else know what it means? Lets add it to the message?

src/common/common.go
tsorya: next time

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: sure

eranco74 (commented):
/lgtm

PR #206, opened 2021-02-08
MGMT-3964 assisted-controller should wait for node to be ready in k8s…

tsorya:
… before marking it as done

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -135,14 +135,20 @@ func (c *controller) WaitAndUpdateNodesStatus(status *ControllerStatus) {
 			if !ok {
 				continue
 			}
-
-			ctx := utils.GenerateRequestContext()
-			log := utils.RequestIDLogger(ctx, c.log)
-			log.Infof("Found new joined node %s with inventory id %s, kubernetes id %s, updating its status to %s",
-				node.Name, host.Host.ID.String(), node.Status.NodeInfo.SystemUUID, models.HostStageDone)
-			if err := c.ic.UpdateHostInstallProgress(ctx, host.Host.ID.String(), models.HostStageDone, ""); err != nil {
-				log.Errorf("Failed to update node %s installation status, %s", node.Name, err)
-				continue
+			if common.IsK8sNodeIsReady(node) {
+				log.Infof("Found new ready node %s with inventory id %s, kubernetes id %s, updating its status to %s",
+					node.Name, host.Host.ID.String(), node.Status.NodeInfo.SystemUUID, models.HostStageDone)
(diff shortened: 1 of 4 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -135,14 +135,21 @@ func (c *controller) WaitAndUpdateNodesStatus(status *ControllerStatus) {
...
 			log := utils.RequestIDLogger(ctx, c.log)
-			log.Infof("Found new joined node %s with inventory id %s, kubernetes id %s, updating its status to %s",
-				node.Name, host.Host.ID.String(), node.Status.NodeInfo.SystemUUID, models.HostStageDone)
-			if err := c.ic.UpdateHostInstallProgress(ctx, host.Host.ID.String(), models.HostStageDone, ""); err != nil {
-				log.Errorf("Failed to update node %s installation status, %s", node.Name, err)
-				continue
+			if common.IsK8sNodeIsReady(node) {
+				log.Infof("Found new joined node %s with inventory id %s, kubernetes id %s, updating its status to %s",
+					node.Name, host.Host.ID.String(), node.Status.NodeInfo.SystemUUID, models.HostStageDone)
+				if err := c.ic.UpdateHostInstallProgress(ctx, host.Host.ID.String(), models.HostStageDone, ""); err != nil {
```
eranco74, line 143: What if the node is already in `Done`?
No reason to spam the service with more updates
Consider adding `models.HostStageDone` to `ignoreStatuses`

src/installer/installer.go
```diff
@@ -578,22 +578,19 @@ func (i *installer) updateReadyMasters(nodes *v1.NodeList, readyMasters *[]strin
...
-				host, ok := inventoryHostsMap[strings.ToLower(node.Name)]
-				if !ok {
-					log.Warnf("Node %s is not in inventory hosts", node.Name)
-					break
-				}
-				ctx = utils.GenerateRequestContext()
-				if err := i.inventoryClient.UpdateHostInstallProgress(ctx, host.Host.ID.String(), models.HostStageJoined, ""); err != nil {
-					utils.RequestIDLogger(ctx, i.log).Errorf("Failed to update node installation status, %s", err)
-				}
+		if common.IsK8sNodeIsReady(node) && !funk.ContainsString(*readyMasters, node.Name) {
```
eranco74, line 581: :+1:

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: it will not spam. Node in Done status will not be part of assistedInstallerNodesMap. It will be filtered here : https://github.com/openshift/assisted-installer/pull/206/files#diff-1da579fda3bc689940661d48dd744a46f23824270427207cfae4f8c7448eddd2L134

PR #210, opened 2021-02-11
MGMT-3145 Reset CVO timeout in case of a change

YuviGold:


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -556,9 +556,14 @@ func (c controller) validateConsolePod() bool {
...
 
 func (c controller) waitingForClusterVersion() error {
-	isClusterVersionAvailable := func() bool {
+	isClusterVersionAvailable := func(timer *time.Timer) bool {
 		newCvoStatus := c.validateClusterVersion()
 		if c.cvoStatus.isAvailable != newCvoStatus.isAvailable || (c.cvoStatus.message != newCvoStatus.message && newCvoStatus.message != "") {
+			if !timer.Stop() {
+				<-timer.C
+			}
+			timer.Reset(WaitTimeout)
```
romfreiman, line 565: what change are u looking at?

YuviGold: Message/Availability change.
Basically, that timeout indicates static-waiting and not on-going changes.

romfreiman: But message can be:
"Multiple Errors"....

YuviGold: We can start with this check and in the future make a smarter check. Unless you think it's necessary.

nmagnezi: @YuviGold do you have an assisted-service log with this running, for reference?

YuviGold:
Timeout will be handled by the service.

PR #213, opened 2021-02-17
MGMT-3860 Change use of certificates.k8s.io/v1beta1 to

tsorya:
certificates.k8s.io/v1 in assisted-controller, it will remove many
warnings in the logs


eranco74 (commented):
/lgtm

PR #214, opened 2021-02-18
MGMT-3662 Add OLM operators monitoring

machacekondra:
This patch add OLM operators monitoring.


machacekondra:
@pkliczewski @masayag @jakub-dzon @jordigilh @YuviGold @tsorya You can take a look, the patch is not yet ready, as waiting for the backend side, but in general the flow is there.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -538,6 +544,60 @@ func (c controller) addRouterCAToClusterCA() bool {
...
+			ret = append(ret, olms[string(operator.OperatorType)])
+		}
+	}
+	return ret, nil
+}
+
+// TODO: remove
+func updateStatus(olm OLMOperator, phase string) {
+	olm.status = phase
+	olms[olm.name] = olm
```
pkliczewski, line 580: What about keeping track of conditions as well? It could be helpful to show users more information about potential failure.

src/inventory_client/inventory_client.go
```diff
@@ -220,6 +222,18 @@ func (c *inventoryClient) GetCluster(ctx context.Context) (*models.Cluster, erro
...
 
+// TODO: change when appropriate method ready
+func (c *inventoryClient) GetClusterOLM(ctx context.Context) (*models.Operators, error) {
+	cluster, err := c.ai.Installer.GetCluster(ctx, &installer.GetClusterParams{ClusterID: c.clusterId})
+	if err != nil {
+		return nil, aserror.GetAssistedError(err)
+	}
+
+	operators := &models.Operators{}
+	json.Unmarshal([]byte(cluster.Payload.Operators), operators)
```
pkliczewski, line 233: would be good to handle unmarshalling error here.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -538,6 +544,60 @@ func (c controller) addRouterCAToClusterCA() bool {
...
+}
+
+func (c controller) waitForOLMOperators() bool {
+	c.log.Infof("Checking OLM operators")
+	opeartors, _ := c.getOLMOperators()
+	if len(opeartors) == 0 {
```
pkliczewski, line 586: Would be good to add operator status based exit criteria so we won't wait till the timeout occurs.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: In current API design, there are `status` and `status_info` parameters. The `status` would reflect the `phase` and the `status_info` could reflect `message` from CSV's `status.message`. We could also store the conditions. But I think the only reason we store those informations is to decide whether the cluster should be "Succeeded" or "Degraded", not sure we need on the assisted service side to know all the conditions of the CSV. But definitelly it would be nice to log them.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Well, it is. The `getOLMOperators` returns only operattors which are `enabled` and are not in `Succeeded` state. So we monitor only CSVs which are not in `Succeeded` state. But we could actually improve to also check for failures, and don't monitor the CSV which end in failed state.

src/assisted_installer_controller/assisted_installer_controller.go
masayag: The status will be a translation of each operator (olm, cvo, co) status to a common status as the assisted-service will declare.
I think storing the condition will be too much information. The status_info will report the message for failed conditions or progress for cvo.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -538,6 +544,60 @@ func (c controller) addRouterCAToClusterCA() bool {
...
+		return true
+	}
+	for _, olm := range opeartors {
+		csv, err := c.kc.GetCSV(olm.namespace, olm.name)
```
masayag, line 590: I think that in case of a failure to obtain CSV for an operator, it status should be reported.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -538,6 +544,60 @@ func (c controller) addRouterCAToClusterCA() bool {
 
 }
 
+// TODO: remove
+type OLMOperator struct {
+	name      string
+	namespace string
+	enabled   bool
```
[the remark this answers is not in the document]
machacekondra: Well, you can have for example:
`
CNV: false
OCS: true
LSO: true
`

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -538,6 +544,60 @@ func (c controller) addRouterCAToClusterCA() bool {
...
+	status    string
```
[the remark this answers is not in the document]
machacekondra: This is just a temp struct used for local testing before monitoring API is done. It will be removed when operator API is implemented.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Should be reported in the `status_info` attribute?

masayag: yes, the status will indicate that the installation is either in progress or failed - not sure CSV can be created later certain time.
and the status_info will detail that it is due to lack of CSV resource.

src/assisted_installer_controller/assisted_installer_controller.go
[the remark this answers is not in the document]
masayag: this struct should specific the operators that should be installed on the cluster.
If they are specified - they are expected to exist. therefore the 'enabled' can be spared.

src/assisted_installer_controller/assisted_installer_controller.go
jordigilh: s/opeartors/operators

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -538,6 +544,60 @@ func (c controller) addRouterCAToClusterCA() bool {
...
+
+// TODO: remove
+func updateStatus(olm OLMOperator, phase string) {
+	olm.status = phase
+	olms[olm.name] = olm
+}
+
+func (c controller) waitForOLMOperators() bool {
+	c.log.Infof("Checking OLM operators")
+	opeartors, _ := c.getOLMOperators()
```
jordigilh, line 585: same here,
s/opeartors/operators

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -538,6 +544,60 @@ func (c controller) addRouterCAToClusterCA() bool {
...
+	if len(opeartors) == 0 {
+		return true
+	}
+	for _, olm := range opeartors {
+		csv, err := c.kc.GetCSV(olm.namespace, olm.name)
+		if err != nil {
+			c.log.WithError(err).Warnf("Failed to get %s", olm.name)
```
jordigilh, line 592: Shouldn't you `continue` the execution in case of error? I am not sure that the `csv` object could be considered correct, it might be even nil, in which case the next call to get the `csv.Status.Phase` should cause a panic. WDYT?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -538,6 +544,60 @@ func (c controller) addRouterCAToClusterCA() bool {
...
+		}
+		// TODO: c.ic.UpdateClusterOLM(context.TODO(), csv.name)
+		updateStatus(olm, string(csv.Status.Phase)) // TODO: remove
+		c.log.Infof("CSV %s is in status %s.", olm.name, csv.Status.Phase)
+	}
+	return false
```
jordigilh, line 598: I'm confused why are you returning `false` here. I'd expect the return value to depend on the status of the CSV of all operators: if all CSVs have a succeeded state, then the return should be true so that the call in line 293 will continue successfully, else false.
If I'm reading this correctly, this will always return `false` regardless of the status of the CSVs, and only `true` when there are no OLM operators to install.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Good catch, thanks. Maybe we can actually update the status with some error state and not `continue`. Let's see how the end design with status field will look like.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Actually this condition is handled on line 587. `getOLMOperators` should return operators which are still installing, and if there are all in succeded state, we return `true`. If there are some operators not yet in `Succeeded` state we return false and timeout eventually if needed.

jordigilh: Maybe rename `getOLMOperators` to `getPendingOLMOperators` or something more meaningful? If `getOLMOperators` return values depend on the status of the CSV, the current name might lead to confusion to think that the output of the command will always return the list of selected operators (as I wrongly expected), regardless of their deployment status. WDYT?

machacekondra: Makes sense, thanks for review.

src/utils/utils.go
```diff
@@ -247,3 +248,52 @@ func IsVersionLessThan47(openshiftVersion string) (bool, error) {
...
+func CsvStatusToOperatorStatus(csvStatus string) models.OperatorStatus {
+	switch csvStatus {
+	case "Succeeded":
+		return models.OperatorStatusAvailable
+	case "Failed":
+		return models.OperatorStatusFailed
+	default:
+		return models.OperatorStatusProgressing
+	}
+}
```
YuviGold, line 261: Can we use CSVPhase constants? 
https://github.com/operator-framework/api/blob/718d279192d30713e6e07a8833a256d27e0c33d4/pkg/operators/v1alpha1/clusterserviceversion_types.go#L327-L348

src/utils/utils.go
```diff
@@ -247,3 +248,52 @@ func IsVersionLessThan47(openshiftVersion string) (bool, error) {
...
+func UnmarshallOperators(operatorsJson string) ([]*models.MonitoredOperator, error) {
+	operators := make([]*models.MonitoredOperator, 0)
+	if operatorsJson != "" {
+		if err := json.Unmarshal([]byte(operatorsJson), &operators); err != nil {
+			return nil, err
+		}
+	}
+
+	return operators, nil
+}
```
YuviGold, line 299: can be removed

go.mod
```diff
@@ -39,10 +41,13 @@ require (
 
 replace (
 	github.com/metal3-io/baremetal-operator => github.com/openshift/baremetal-operator v0.0.0-20200715132148-0f91f62a41fe // Use OpenShift fork
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20210208192252-670ac3fc997c
+	github.com/openshift/assisted-service => github.com/YuviGold/assisted-service v0.0.0-20210302151747-cb00cbfadaa0
+	github.com/openshift/hive/pkg/apis => github.com/carbonin/hive/pkg/apis v0.0.0-20210209195732-57e8c3ae12d1
```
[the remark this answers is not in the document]
machacekondra: The replacements was already merged here: https://github.com/openshift/assisted-installer/pull/228/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R42

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -537,6 +544,78 @@ func (c controller) addRouterCAToClusterCA() bool {
...
+		monitoredOperator, err := c.ic.GetClusterMonitoredOperator(context.TODO(), c.ClusterID, operator.Name)
+		if err != nil {
+			return ret, err
+		}
+		if monitoredOperator.Status != models.OperatorStatusAvailable && monitoredOperator.Status != models.OperatorStatusFailed {
+			ret = append(ret, operator)
+		}
+	}
+	return ret, nil
+}
```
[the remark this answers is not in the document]
machacekondra: Yes, the logic is:
```
1) Query monitored OLM operators from assisted-service
2 a) For those who are progressing - check the CSV status and update the OLM operator status in assisted-service accordingly.
2 b)For the rest(available, failed) don't do anything.
3) If all operators are either failed or available end the waiting.
     
```

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -537,6 +544,78 @@ func (c controller) addRouterCAToClusterCA() bool {
...
+		if err != nil {
+			c.log.WithError(err).Warnf("Failed to get subscription of operator %s", operatorName)
+			continue
+		}
+
+		csv, err := c.kc.GetCSV(csvName, operatorNamespace)
+		if err != nil {
+			c.log.WithError(err).Warnf("Failed to get %s", operator.Name)
+			continue
+		}
```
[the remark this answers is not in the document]
machacekondra: From `assisted-installer` point of view we know only some abstract kind of name like: `cnv`, `ocs` or `lso`. But we have no idea what's subscription name or the CSV name. So we have a mapping method which maps names like `cnv`, `lso`, to the [subscription name](https://github.com/openshift/assisted-service/blob/master/internal/operators/lso/utils.go#L16). And from the subscription name we can latter know the actual CSV name. I think we should extract the subscription name to some public API, or at least don't store it in internal package in `asssisted-service`, so it can be used in `assisted-installer` code, and we don't have to duplicate it `GetSubscriptionNameAndNamespace` method.

src/utils/utils.go
```diff
@@ -247,3 +248,52 @@ func IsVersionLessThan47(openshiftVersion string) (bool, error) {
...
+func GetOLMNameAndNamespace(name string) (string, string, error) {
+	switch name {
+	case "lso":
+		return "local-storage-operator", "openshift-local-storage", nil
+	case "ocs":
+		return "ocs-operator", "openshift-storage", nil
+	default:
+		return "", "", errors.Errorf("No such operator %s found", name)
+	}
+}
```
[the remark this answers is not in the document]
machacekondra: I think I can't because it's not used really: https://github.com/openshift/assisted-service/blob/master/internal/operators/lso/utils.go#L17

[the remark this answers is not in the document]
machacekondra: That's actually one of the issue I've mentioned above we should fix in some follow-up PR.

src/assisted_installer_controller/assisted_installer_controller.go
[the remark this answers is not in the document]
YuviGold: That's fine. I was thinking this function is running each interval of the loop. So I wasn't expecting any changes from the service side regarding the OLM statuses.

src/utils/utils.go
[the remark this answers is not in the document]
YuviGold: We should fix that on the service side. But this is the API that should be used. We can change "prefix_name" to "subscription name" if you think it's more suitable :)

src/assisted_installer_controller/assisted_installer_controller.go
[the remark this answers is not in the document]
YuviGold: Same API as from the other comment https://github.com/openshift/assisted-installer/pull/214#discussion_r590033931

src/utils/utils.go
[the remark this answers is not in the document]
machacekondra: `prefix_name` is fine, but for me it wasn't clear even from the documentation of that attribute, so documenting it better would be a way to solve it as well :)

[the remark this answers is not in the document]
machacekondra: I will work on the API for that, so will put WIP to this patch, and I will rework based on the API.

machacekondra:
After the installation with OLM enabled, the status is properly updated:
```
{
      "cluster_id": "c14a8792-3906-4f00-8067-0bbe0e5477cc",
      "name": "lso",
      "namespace": "openshift-local-storage",
      "operator_type": "olm",
      "status": "available",
      "status_info": "install strategy completed with no errors",
      "status_updated_at": "0001-01-01T00:00:00.000Z",
      "subscription_name": "local-storage-operator",
      "timeout_seconds": 4200
    }
  "status": "installed",
  "status_info": "installed",
```

YuviGold (approved):
LGTM

YuviGold:
Resolves also:
* Reports Console status https://issues.redhat.com/browse/MGMT-4207 
* Reports CVO status https://issues.redhat.com/browse/MGMT-4206

PR #218, opened 2021-02-18
MGMT-4179 Fix installer handling of ignitions to support multiple ver…

tsorya:
…sions - sno mode

### DIFF
--- a/src/ignition/ignition.go
+++ b/src/ignition/ignition.go
@@ -4,19 +4,23 @@ import (
 	"encoding/json"
 	"io/ioutil"
 
+	ignitionConfigPrevVersion "github.com/coreos/ignition/v2/config/v3_1"
+	ignitionConfig "github.com/coreos/ignition/v2/config/v3_2"
+	"github.com/coreos/ignition/v2/config/v3_2/translate"
+	"github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/coreos/ignition/v2/config/validate"
-
-	config31 "github.com/coreos/ignition/v2/config/v3_1"
-	config31types "github.com/coreos/ignition/v2/config/v3_1/types"
-
 	"github.com/pkg/errors"
 )
 
+// Added to manage ignition versions in one file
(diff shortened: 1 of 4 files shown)


src/ignition/ignition.go
```diff
@@ -26,20 +30,34 @@ func NewIgnition() *ignition {
...
 	if err != nil {
-		return nil, errors.Wrap(err, "error parsing ignition")
+		return nil, err
 	}
-	return &config, nil
+	return config, nil
+}
+
+// ParseIgnitionFile reads an ignition config from a given path on disk
+func (i *ignition) parseToLatest(content []byte) (*types.Config, error) {
```
eranco74, line 46: Why not use the same code as the service (where we force 3.2...)?
I think you can actually use the same code as a library...

src/installer/installer.go
```diff
@@ -153,7 +151,7 @@ func (i *installer) updateSingleNodeIgnition(singleNodeIgnitionPath string) erro
 	}
 	// TODO: update this once we can get the full host specific overrides we have in the ignition
 	// Remove the Config part since we only want the rest of the overrides
-	hostConfig.Ignition.Config = config31types.IgnitionConfig{}
+	hostConfig.Ignition.Config = ignition.EmptyIgnitionConfig
```
eranco74, line 154: Hmmm, seems like this should be the master ignition we get from the service no?

src/ignition/ignition.go
tsorya: it is the same code but it gives us an option just to change import and not go through all variables

tsorya: so it is just naming thing

eranco74: Ok

eranco74 (commented):
/lgtm

src/installer/installer.go
tsorya: why?

tsorya: if i understand right the comment, we just remove all the host specific ignition part

PR #226, opened 2021-02-24
MGMT-4185 Add other minor and pre versions to tests

flaper87:
This extends the unit tests run for the AI installer to verify that both
minor and pre release versions are managed. Just one more unit test


flaper87:
@ronniel1 whenever you have a chance, would love to move this one forward.

PR #229, opened 2021-03-02
MGMT-4402: Set boot from disk using efibootmgr

sacharya:


src/ops/ops.go
```diff
@@ -195,6 +196,21 @@ func (o *ops) Reboot() error {
 	return nil
 }
 
+func (o *ops) SetEfiBootMgr(device string) error {
+	o.log.Info("Setting EFI Boot Manager to boot from disk")
```
eranco74, line 200: I think you should first check if the host is booting from efi.
```suggestion
        _, err := o.ExecPrivilegeCommand(o.logWriter, "test", "-f", "/sys/firmware/efi")
	if err != nil {
		o.log.Info("efi not supported")
		return nil
	}
	o.log.Info("Setting EFI Boot Manager to boot from disk")
```

src/ops/ops.go
```diff
@@ -195,6 +196,27 @@ func (o *ops) Reboot() error {
...
+		return nil
+	}
+
+	o.log.Info("Setting efibootmgr to boot from disk")
+	_, err = o.ExecPrivilegeCommand(o.logWriter, "mount", device+"2", "/mnt")
+	if err != nil {
+		o.log.Errorf("Failed to mount device %s, err: %s", device+"2", err)
+		return err
+	}
+	_, err = o.ExecPrivilegeCommand(o.logWriter, "efibootmgr", "-d", device, "-p", "2", "-c", "-L", "RHCOS", "-l", "\\EFI\\BOOT\\BOOTX64.EFI")
```
celebdor, line 212: Shouldn't it be followed by a efibootmgr --bootorder?

sacharya: Good point. It seems to set the new one created as the first in the order but explicit is always better.

src/installer/installer.go
```diff
@@ -116,6 +116,11 @@ func (i *installer) InstallNode() error {
 	if err = i.writeImageToDisk(ignitionPath); err != nil {
 		return err
 	}
+
+	if err = i.ops.SetBootOrder(i.Device); err != nil {
+		return err
```
ronniel1, line 121: We are always doing this. Not only in the scope of metal3 boot.
In case of boot-it-yourself we should definitely not exit if there was an error in changing boot device.
In some cases the HW does not support it

celebdor: Very good point.

sacharya: There is a check for efi inside the function.
`_, err := o.ExecPrivilegeCommand(o.logWriter, "test", "-d", "/sys/firmware/efi")`

eranco74:
Please add the relevant Jira ticket to the commit message

src/ops/ops.go
sacharya: Looked into it and we don't need an explicit boot order using -n here. -c will add it to the top of the order.
`-c | --create         create new variable bootnum and add to bootorder`

src/installer/installer.go
eranco74: While this should do the trick I think that the SetBootOrder should not fail the installation in case of failure (at least not at first).
I suggest that we just log a warning in case we fail here, so we can get the logic merged and not worry about failing installation due to this change.
After this will run for a couple of weeks and we can check the logs to see how it behaves on customers installations we can add the `return err`

eranco74 (commented):
/lgtm

celebdor:
/retitle MGMT-4402: Set boot from disk using efibootmgr

src/ops/ops.go
```diff
@@ -195,6 +196,27 @@ func (o *ops) Reboot() error {
...
+		o.log.Info("efi not supported")
+		return nil
+	}
+
+	o.log.Info("Setting efibootmgr to boot from disk")
+	_, err = o.ExecPrivilegeCommand(o.logWriter, "mount", device+"2", "/mnt")
+	if err != nil {
+		o.log.Errorf("Failed to mount device %s, err: %s", device+"2", err)
+		return err
+	}
```
YuviGold, line 211: Wouldn't it make the coreos-installer fail since it checks the device for [busy partitions](https://github.com/coreos/coreos-installer/blob/7b689344a1f9f9692345a41d9dc81f810251c9f6/src/install.rs#L123-L142)?

src/ops/ops.go
```diff
@@ -195,6 +196,27 @@ func (o *ops) Reboot() error {
...
+	_, err = o.ExecPrivilegeCommand(o.logWriter, "efibootmgr", "-d", device, "-p", "2", "-c", "-L", "RHCOS", "-l", "\\EFI\\BOOT\\BOOTX64.EFI")
```
YuviGold, line 212: Does coreos-installer always install on the second partition of a device? Please add a comment explaining it.

src/ops/ops.go
```diff
@@ -195,6 +196,27 @@ func (o *ops) Reboot() error {
 	return nil
 }
 
+func (o *ops) SetBootOrder(device string) error {
```
eranco74, line 199: Please don't return an error from this function yet.
I think we should start with "best effort" mode so in case we fail to set the boot order the installation won't fail.
This way it can merge today, and we can see it in action in prod.
Once we gain some certainty about what happens at customers sites, we can add the return err

eranco74 (changes requested):

YuviGold:
We skipped gathering logs on a successful build. Now changed on https://github.com/openshift/release/pull/16792
Let's see if the Packet instances are efi from the logs.

/test ci/prow/e2e-metal-assisted

src/installer/installer.go
sacharya: Done

src/ops/ops.go
sacharya: Done

src/ops/ops.go
sacharya: Doesn't the installer complete before this happens?

src/ops/ops.go
sacharya: Added a comment!

src/ops/ops.go
YuviGold: @eranco74  @tsorya do you know?

YuviGold:
No EFI support for packet servers
```
Mar 24 16:20:07 test-infra-cluster-assisted-installer-master-0 installer[3443]: time="2021-03-24T16:20:07Z" level=info msg="failed executing nsenter [-t 1 -m -i -- test -d /sys/firmware/efi], env vars [PATH=/usr/local/sbin:/usr/local/bin:
Mar 24 16:20:07 test-infra-cluster-assisted-installer-master-0 installer[3443]: time="2021-03-24T16:20:07Z" level=info msg="efi not supported"
```

So no breakages from CI but this code change isn't tested

eranco74 (commented):
/lgtm
/approve

travier:
Could you give us more input on why you needed to add `efibootmgr` calls for RHCOS? We are currently investigating that and planning on doing the reverse: making sure that no EFI boot entry ever get created. See https://github.com/coreos/fedora-coreos-tracker/issues/946

omertuc:
> Could you give us more input on why you needed to add `efibootmgr` calls for RHCOS? We are currently investigating that and planning on doing the reverse: making sure that no EFI boot entry ever get created. See [coreos/fedora-coreos-tracker#946](https://github.com/coreos/fedora-coreos-tracker/issues/946)

The assisted installer is also used to enable zero-touch provisioning of OCP - if we don't create this entry, we see that after installation is complete and we reboot the host, the host doesn't necessarily boot into the newly installed RHCOS. 

I am however not aware of the following details:
- If the host doesn't boot into the RHCOS, what does it boot into? I assume it boots into the installation media again, instead.
- Why do we *only* create an entry but not modify the actual EFI boot order / setting NextBoot. I guess creating an entry is enough?
- I'm not sure if doing this is required in general or required just on some platforms we've tested in our ZTP testing

PR #230, opened 2021-03-04
NO-ISSUE Print a message when logs successfully uploaded

empovit:
It can be helpful to know if uploading log files was completed successfully, 
because there are three possible states:
 1. Failed
 2. Succeeded
 3. Never finished (terminated).

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -739,6 +739,8 @@ func (c *controller) UploadLogs(ctx context.Context, cancellog context.CancelFun
 					err := c.uploadSummaryLogs(podName, c.Namespace, controllerLogsSecondsAgo, status.HasError())
 					if err != nil {
 						c.log.Infof("retry uploading logs in 5 minutes...")
+					} else {
+						c.log.Info("successfully uploaded final controller and cluster logs")
 					}
 					return err == nil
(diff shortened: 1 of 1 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -739,6 +739,8 @@ func (c *controller) UploadLogs(ctx context.Context, cancellog context.CancelFun
 					err := c.uploadSummaryLogs(podName, c.Namespace, controllerLogsSecondsAgo, status.HasError())
 					if err != nil {
 						c.log.Infof("retry uploading logs in 5 minutes...")
+					} else {
+						c.log.Info("successfully uploaded final controller and cluster logs")
```
tsorya, line 743: just move line to 747 before return , no need for else

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -771,6 +773,8 @@ func (c *controller) UploadLogs(ctx context.Context, cancellog context.CancelFun
 			if err != nil {
 				c.log.WithError(err).Warnf("Failed to upload controller logs")
 				continue
+			} else {
```
tsorya, line 776: no need for else

empovit:
Not really helpful because we'll never see this in uploaded logs

PR #233, opened 2021-03-11
MGMT-4479 Specify must-gather image

empovit:
Tell must-gather command which image to use. This should enable
to run must-gather with an available and working image instead of
the default latest in cases when it's not possible to get an image
ID from OCP stream (e.g. the API fails).


asalkeld:
This seems ok to me.

PR #249, opened 2021-04-01
MGMT-4836 Enhance unknown host logs

YuviGold:
The controller doesn't inform in case there's an unknown hostname
in the k8s nodes. It will just get a timeout for not finding the wanted hosts.
    
In addition, the bootstrap does inform about it (warning logs) while it should fail.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1129,6 +1129,7 @@ github.com/operator-framework/api v0.6.1 h1:bTOj1GseGV5oZc5iNQL5ZMqzP430sefNMkPb
 github.com/operator-framework/api v0.6.1/go.mod h1:L7IvLd/ckxJEJg/t4oTTlnHKAJIP/p51AvEslW3wYdY=
 github.com/operator-framework/api v0.7.0 h1:uND/8x4J44eVEsGHeNMIQUYlduggG61CNPXlO8eXf3k=
 github.com/operator-framework/api v0.7.0/go.mod h1:L7IvLd/ckxJEJg/t4oTTlnHKAJIP/p51AvEslW3wYdY=
+github.com/operator-framework/api v0.7.1 h1:L1KKlfvEVp9qP1KnkQApb7xuOnrWG3Ds9XkY6BxDXOc=
 github.com/operator-framework/api v0.7.1/go.mod h1:L7IvLd/ckxJEJg/t4oTTlnHKAJIP/p51AvEslW3wYdY=
 github.com/operator-framework/operator-lifecycle-manager v0.0.0-20200321030439-57b580e57e88 h1:ByKBik0i2aTEr7iKdSCmUGULydHwr6hA0h4INv9LkSA=
 github.com/operator-framework/operator-lifecycle-manager v0.0.0-20200321030439-57b580e57e88/go.mod h1:7Ut8p9jJ8C6RZyyhZfZypmlibCIJwK5Wcc+WZDgLkOA=
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -98,48 +98,61 @@ func (status *ControllerStatus) HasError() bool {
 	return atomic.LoadUint32(&status.errCounter) > 0
 }
 
+func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.HostData) {
+	hostsStatus := make(map[string][]string)
+	for hostname, hostData := range hosts {
+		hostsStatus[hostname] = []string{
+			*hostData.Host.Status,
+			string(hostData.Host.Progress.CurrentStage),
+			hostData.Host.Progress.ProgressInfo}
+	}
+	log.Infof("Hosts status: %v", hostsStatus)
+}
+
 func (c *controller) WaitAndUpdateNodesStatus(status *ControllerStatus) {
 	c.log.Infof("Waiting till all nodes will join and update status to assisted installer")
-	ignoreStatuses := []string{models.HostStatusDisabled, models.HostStatusInstalled}
+	ignoreStatuses := []string{models.HostStatusDisabled}
 	var hostsInError int
 	for {
 		time.Sleep(GeneralWaitInterval)
 		ctx := utils.GenerateRequestContext()
 		log := utils.RequestIDLogger(ctx, c.log)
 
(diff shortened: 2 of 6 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -132,14 +138,16 @@ func (c *controller) WaitAndUpdateNodesStatus(status *ControllerStatus) {
 			break
 		}
 		//otherwise, update the progress status and keep waiting
-		log.Infof("Searching for host to change status, number to find %d", len(assistedInstallerNodesMap))
+		log.Infof("Checking if cluster nodes are ready. %d nodes remained", len(assistedInstallerNodesMap))
```
omertuc, line 141: Remaining?

src/common/common.go
```diff
@@ -47,7 +47,7 @@ func SetConfiguringStatusForHosts(client inventory_client.InventoryClient, inven
 		pat := fmt.Sprintf("(%s).{1,20}(Ignition)", strings.Join(host.IPs, "|"))
 		pattern, err := regexp.Compile(pat)
 		if err != nil {
-			log.WithError(err).Errorf("Failed to compile regex from host %s ips list", host.Host.ID.String())
+			log.WithError(err).Errorf("Failed to compile regex from host %s ips list", key)
```
omertuc, line 50: maybe give the `key` var a better name while you're here?

go.sum
```diff
@@ -1129,6 +1129,7 @@ github.com/operator-framework/api v0.6.1 h1:bTOj1GseGV5oZc5iNQL5ZMqzP430sefNMkPb
 github.com/operator-framework/api v0.6.1/go.mod h1:L7IvLd/ckxJEJg/t4oTTlnHKAJIP/p51AvEslW3wYdY=
 github.com/operator-framework/api v0.7.0 h1:uND/8x4J44eVEsGHeNMIQUYlduggG61CNPXlO8eXf3k=
 github.com/operator-framework/api v0.7.0/go.mod h1:L7IvLd/ckxJEJg/t4oTTlnHKAJIP/p51AvEslW3wYdY=
+github.com/operator-framework/api v0.7.1 h1:L1KKlfvEVp9qP1KnkQApb7xuOnrWG3Ds9XkY6BxDXOc=
```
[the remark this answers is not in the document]
YuviGold: automatically by IDE

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -132,14 +138,16 @@ func (c *controller) WaitAndUpdateNodesStatus(status *ControllerStatus) {
...
+		log.Infof("Checking if cluster nodes are ready. %d nodes remaining", len(assistedInstallerNodesMap))
 		nodes, err := c.kc.ListNodes()
 		if err != nil {
+			log.WithError(err).Error("Failed to get list of nodes from k8s client")
 			continue
 		}
 		for _, node := range nodes.Items {
 			host, ok := assistedInstallerNodesMap[strings.ToLower(node.Name)]
 			if !ok {
+				log.Warnf("Node %s is not in inventory hosts", strings.ToLower(node.Name))
```
tsorya, line 150: it is wrong way to do it. We ignore all the nodes in installed status so they will not be in a list.
Please run this flow and verify logs before pushing

src/common/common.go
```diff
@@ -57,12 +57,12 @@ func SetConfiguringStatusForHosts(client inventory_client.InventoryClient, inven
 			}
 			ctx := utils.GenerateRequestContext()
 			requestLog := utils.RequestIDLogger(ctx, log)
-			requestLog.Infof("Host %s found in mcs logs, moving it to %s state", host.Host.ID.String(), status)
+			requestLog.Infof("Host %s (%s) found in mcs logs, moving it to %s state", hostName, host.Host.ID.String(), status)
```
tsorya, line 60: i think better to change %s to %q and remove ()

src/installer/installer.go
```diff
@@ -567,7 +567,10 @@ func (i *installer) waitForMasterNodes(ctx context.Context, minMasterNodes int,
 			i.log.Warnf("Still waiting for master nodes: %v", err)
 			return false
 		}
-		i.updateReadyMasters(nodes, &readyMasters, inventoryHostsMap)
+		if err = i.updateReadyMasters(nodes, &readyMasters, inventoryHostsMap); err != nil {
+			i.log.Warn(err)
```
tsorya, line 571: log.witherror and add relevant message please

src/installer/installer.go
```diff
@@ -622,16 +625,17 @@ func (i *installer) updateReadyMasters(nodes *v1.NodeList, readyMasters *[]strin
 			*readyMasters = append(*readyMasters, node.Name)
 			host, ok := inventoryHostsMap[strings.ToLower(node.Name)]
 			if !ok {
-				log.Warnf("Node %s is not in inventory hosts", node.Name)
-				break
+				return fmt.Errorf("Node %s is not in inventory hosts", node.Name)
```
tsorya, line 628: as i said previously i am not sure it is right behavior. I think that ready master is part of the inventory hosts and not return error from here. To say the truth if one of the masters is not in the list there is no many sense to continue installation and will be nice to generate event in cluster regarding it.
@eranco74 what do you think?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -98,6 +98,17 @@ func (status *ControllerStatus) HasError() bool {
...
 
+func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.HostData) {
+	hostsStatus := make(map[string][]string)
+	for hostname, hostData := range hosts {
+		hostsStatus[hostname] = []string{
+			*hostData.Host.Status,
+			string(hostData.Host.Progress.CurrentStage),
+			hostData.Host.Progress.ProgressInfo}
+	}
+	log.Infof("Host status: %v", hostsStatus)
```
gamli75, line 109: small remark - please change it to "Hosts status"

src/installer/installer.go
eranco74: I think that in case the `node.Name` isn't in the map we will fail on timeout anyway (prior to this change there we will just log this warning till the timeout due to the break in the next line).
So I think it's better to fail early.
@tsorya do you think there is any scenario that we get this error and later we recover?

PR #250, opened 2021-04-03
MGMT-4865 - Resolve installation disk symlink before running coreos-installer

omertuc:
Overcome https://github.com/coreos/coreos-installer/issues/512 bug.

coreos-installer has a bug where when a disk has busy partitions, it will
print a confusing error message if that disk doesn't have a `/dev/*` style path.
The service may give us paths that don't have the `/dev/*` path format but instead
are symlinks to the actual `/dev/*` path. e.g. `/dev/disk/by-id/wwn-*`.
To fix the bug we simply resolve the symlink and pass the resolved link to coreos-installer.

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -174,9 +174,24 @@ func (o *ops) SystemctlAction(action string, args ...string) error {
 }
 
 func (o *ops) WriteImageToDisk(ignitionPath string, device string, progressReporter inventory_client.InventoryClient, extraArgs []string) error {
+	// Overcome https://github.com/coreos/coreos-installer/issues/512 bug.
+	// coreos-installer has a bug where when a disk has busy partitions, it will
+	// print a confusing error message if that disk doesn't have a `/dev/*` style path.
+	// The service may give us paths that don't have the `/dev/*` path format but instead
+	// are symlinks to the actual `/dev/*` path. e.g. `/dev/disk/by-id/wwn-*`.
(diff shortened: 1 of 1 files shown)


omertuc:
Force pushed to update irrelevant/wrong commit message

src/ops/ops.go
```diff
@@ -174,9 +174,24 @@ func (o *ops) SystemctlAction(action string, args ...string) error {
...
 func (o *ops) WriteImageToDisk(ignitionPath string, device string, progressReporter inventory_client.InventoryClient, extraArgs []string) error {
+	// Overcome https://github.com/coreos/coreos-installer/issues/512 bug.
+	// coreos-installer has a bug where when a disk has busy partitions, it will
+	// print a confusing error message if that disk doesn't have a `/dev/*` style path.
+	// The service may give us paths that don't have the `/dev/*` path format but instead
+	// are symlinks to the actual `/dev/*` path. e.g. `/dev/disk/by-id/wwn-*`.
+	// To fix the bug we simply resolve the symlink and pass the resolved link to coreos-installer.
+	linkTarget, err := filepath.EvalSymlinks(device)
+	if err != nil {
+		o.log.Warnf("Failed to os.ReadLink(%s): %s. Continuing with %s anyway.",
```
michaellevy101, line 185: Please change the message to: Failed to EvalSymlinks...

PR #253, opened 2021-04-06
OCPBUGSM-27183: Update pending OLM operators in case of timeout

machacekondra:
This PR add a functionality to update the pending operators in case of
timeout. Previously we left the OLM operators state as it was. So in
case timeout occur the operators which didn't finished had state
progressing. With this PR we update the state of the operators, which
timed out to Failed state. So the assisted-service can switch the state
to completed.

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller_test.go
+++ b/src/assisted_installer_controller/assisted_installer_controller_test.go
@@ -82,6 +82,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 		"node1": "2834ff2e-8965-48a5-859a-0f1459485a77",
 		"node2": "57df89ee-3546-48a5-859a-0f1459485a66"}
 	l.SetOutput(ioutil.Discard)
+	consoleOperatorName := "console"
 
 	BeforeEach(func() {
 		ctrl = gomock.NewController(GinkgoT())
@@ -399,7 +400,6 @@ var _ = Describe("installer HostRoleMaster role", func() {
 		It("Run PostInstallConfigs - wait for cluster version", func() {
 			cmName := "default-ingress-cert"
 			cmNamespace := "openshift-config-managed"
-			consoleOperatorName := "console"
 			data := make(map[string]string)
 			data["ca-bundle.crt"] = "CA"
 			cm := v1.ConfigMap{Data: data}
@@ -481,6 +481,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 	})
 
 	Context("PostInstallConfigs - not waiting for cluster version ", func() {
+
 		BeforeEach(func() {
 			assistedController.WaitForClusterVersion = false
 			GeneralWaitInterval = 1 * time.Second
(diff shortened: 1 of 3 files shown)


machacekondra:
@YuviGold please review.

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -553,6 +553,36 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
+			).AnyTimes()
+			mockk8sclient.EXPECT().GetCSVFromSubscription("openshift-local-storage", "local-storage-operator").Return("lso-1.1", nil).AnyTimes()
+			mockk8sclient.EXPECT().GetCSV("openshift-local-storage", "lso-1.1").Return(&olmv1alpha1.ClusterServiceVersion{Status: olmv1alpha1.ClusterServiceVersionStatus{Phase: olmv1alpha1.CSVPhaseInstalling}}, nil).AnyTimes()
+			mockbmclient.EXPECT().UpdateClusterOperator(gomock.Any(), gomock.Any(), gomock.Any(), models.OperatorStatusProgressing, gomock.Any()).Return(nil).AnyTimes()
+			mockbmclient.EXPECT().UpdateClusterOperator(gomock.Any(), gomock.Any(), gomock.Any(), models.OperatorStatusAvailable, gomock.Any()).Return(nil).Times(1)
+			mockbmclient.EXPECT().UpdateClusterOperator(gomock.Any(), gomock.Any(), gomock.Any(), models.OperatorStatusFailed, "Waiting for operator timed out").Return(nil).Times(1)
+			mockbmclient.EXPECT().CompleteInstallation(gomock.Any(), "cluster-id", true, "").Return(fmt.Errorf("dummy")).Times(1)
+			mockbmclient.EXPECT().CompleteInstallation(gomock.Any(), "cluster-id", true, "").Return(nil).Times(1)
```
YuviGold, line 584: `CompleteInstallation` can't be called twice

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -553,6 +553,36 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
+			consoleOperatorName := "console"
+			data := make(map[string]string)
+			data["ca-bundle.crt"] = "CA"
+			cm := v1.ConfigMap{Data: data}
+			finalizing := models.ClusterStatusFinalizing
+			cluster := models.Cluster{Status: &finalizing}
+			WaitTimeout = 5 * time.Second
+			mockbmclient.EXPECT().GetCluster(gomock.Any()).Return(&cluster, nil).Times(1)
+			mockk8sclient.EXPECT().GetConfigMap(cmNamespace, cmName).Return(&cm, nil).Times(1)
+			mockbmclient.EXPECT().UploadIngressCa(gomock.Any(), data["ca-bundle.crt"], assistedController.ClusterID).Return(nil).Times(1)
```
YuviGold, line 573: It would be nice if you could move all of this section into a helper function `SetControllerWaitForOLMOperators` or something like that

PR #262, opened 2021-04-14
OCPBUGSM-27526 Sync operator status with the service

YuviGold:
Without the sync, a network issue could occur and an operator update could be
missed. Hence, it is mandatory to verify that the service is aligned
with the operators' statuses before existing.

In addition, a lot of organization in the assisted_installer_controller
UTs.

The OLM implementation already takes care of going over the progressing
monitored operator.

/cc @nmagnezi @tsorya

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -32,14 +32,20 @@ import (
 )
 
 const (
-	generalWaitTimeoutInt    = 30
-	controllerLogsSecondsAgo = 120 * 60
+	generalWaitTimeoutInt     = 30
+	controllerLogsSecondsAgo  = 120 * 60
+	consoleOperatorName       = "console"
+	cvoOperatorName           = "cvo"
+	ingressConfigMapName      = "default-ingress-cert"
+	ingressConfigMapNamespace = "openshift-config-managed"
 )
 
-var GeneralWaitInterval = generalWaitTimeoutInt * time.Second
-var GeneralProgressUpdateInt = 60 * time.Second
-var LogsUploadPeriod = 5 * time.Minute
-var WaitTimeout = 70 * time.Minute
+var (
+	GeneralWaitInterval      = generalWaitTimeoutInt * time.Second
+	GeneralProgressUpdateInt = 60 * time.Second
+	LogsUploadPeriod         = 5 * time.Minute
+	WaitTimeout              = 70 * time.Minute
+)
 
 // assisted installer controller is added to control installation process after  bootstrap pivot
 // assisted installer will deploy it on installation process
@@ -67,16 +73,10 @@ type ControllerStatus struct {
 
 type controller struct {
 	ControllerConfig
-	log       *logrus.Logger
-	ops       ops.Ops
-	ic        inventory_client.InventoryClient
-	kc        k8s_client.K8SClient
-	cvoStatus OperatorStatus
-}
-
-type OperatorStatus struct {
-	isAvailable bool
-	message     string
+	log *logrus.Logger
+	ops ops.Ops
+	ic  inventory_client.InventoryClient
+	kc  k8s_client.K8SClient
 }
 
 func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inventory_client.InventoryClient, kc k8s_client.K8SClient) *controller {
@@ -86,7 +86,6 @@ func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inv
 		ops:              ops,
 		ic:               ic,
 		kc:               kc,
-		cvoStatus:        OperatorStatus{isAvailable: false, message: ""},
 	}
 }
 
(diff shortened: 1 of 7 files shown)


src/utils/utils.go
```diff
@@ -272,5 +272,6 @@ func ClusterOperatorConditionsToMonitoredOperatorStatus(conditions []configv1.Cl
 			return models.OperatorStatusFailed, condition.Message
 		}
 	}
+
```
nmagnezi, line 275: ?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -543,15 +541,13 @@ func (c controller) unpatchEtcd() bool {
 
 // AddRouterCAToClusterCA adds router CA to cluster CA in kubeconfig
 func (c controller) addRouterCAToClusterCA() bool {
-	cmName := "default-ingress-cert"
-	cmNamespace := "openshift-config-managed"
 	ctx := utils.GenerateRequestContext()
 	log := utils.RequestIDLogger(ctx, c.log)
 	log.Infof("Start adding ingress ca to cluster")
-	caConfigMap, err := c.kc.GetConfigMap(cmNamespace, cmName)
+	caConfigMap, err := c.kc.GetConfigMap(ingressConfigMapNamespace, ingressConfigMapName)
```
nmagnezi, line 547: I agree that a `const` is better for this, but I'm not sure why is done in this PR.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -663,29 +659,60 @@ func (c controller) validateConsoleAvailability() bool {
...
+
+	return false
 }
 
 func (c controller) waitingForClusterVersion() error {
 	isClusterVersionAvailable := func() bool {
-		newCvoStatus, conditions := c.validateClusterVersion()
-		if c.cvoStatus.isAvailable != newCvoStatus.isAvailable || (c.cvoStatus.message != newCvoStatus.message && newCvoStatus.message != "") {
-			status := fmt.Sprintf("Cluster version is available: %t , message: %s", newCvoStatus.isAvailable, newCvoStatus.message)
+		c.log.Infof("Checking console operator availability status")
```
nmagnezi, line 679: should probably be `"Checking cluster version operator availability..."`

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -643,17 +639,17 @@ func (c controller) waitForOLMOperators() bool {
 
 // validateConsoleAvailability checks if the console operator is available
 func (c controller) validateConsoleAvailability() bool {
```
nmagnezi, line 641: `validateConsoleAvailability` now has an additional functionality.
It both validates the operator status locally and makes sure, in some cases, that it is up to date on `assisted-service` side.

It looks like it should be done in two different functions, in which the latter gets invoked upon need, for the `assisted-service` side validation.

However, I do see it is all nested under `WaitForPredicate`, and it might make it needlessly complex, so break it into two functions.

So the point that I'm making here is that from the outside, it is not clear `validateConsoleAvailability` makes that server-side validation, so can we at least add a docstring to elaborate that?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -663,29 +659,60 @@ func (c controller) validateConsoleAvailability() bool {
...
+
+	if operatorStatusInService.Status == models.OperatorStatusAvailable {
+		c.log.Infof("Service acknowledged console operator is available for cluster %s", c.ClusterID)
+		return true
+	}
+
+	return false
 }
 
 func (c controller) waitingForClusterVersion() error {
```
nmagnezi, line 677: similar comment to the comment I made about `validateConsoleAvailability`

src/assisted_installer_controller/assisted_installer_controller.go
YuviGold: Why not? :)

nmagnezi: How does it relate to the PR topic? :)

src/assisted_installer_controller/assisted_installer_controller.go
YuviGold: It would require changing the behavior of the CVO reporting. I will do that as part of MGMT-5188 where I will merge the same usage of the CVO reporting with the console.

src/assisted_installer_controller/assisted_installer_controller.go
YuviGold: It doesn't. But do you want to make a separate PR just for that? I wrote in the commit message that I organized the UT :)

nmagnezi: OK ¯\\_(ツ)_/¯

src/assisted_installer_controller/assisted_installer_controller.go
YuviGold: I split the function into two functions: `isOperatorAvailableInCluster` and `isOperatorAvailableInService`

src/assisted_installer_controller/assisted_installer_controller.go
nmagnezi: Since renaming to `waitingForClusterVersionAndAcknowledgeInAssistedService` is not an option, let's just add a short docstring :)

YuviGold: You got it. :) 
```
// waitingForClusterVersion checks the Cluster Version Operator availability in the
// new OCP cluster. A success would be announced only when the service acknowledges
// the CVO availability, in order to avoid unsycned scenarios.
//
// This function would be aligned with the console operator reporting workflow
// as part of the deprecation of the old API in MGMT-5188.
```

nmagnezi (approved):
Thank you for the prompt fixes. Approved.

PR #265, opened 2021-04-19
OCPBUGSM-27762 Automatic boot order setting is done incorrectly when …

eranco74:
…using by-path style device names

Evaluate the installation device symlink to the actual disk path at the start of the installation

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -478,6 +478,7 @@ github.com/go-openapi/runtime v0.19.26 h1:K/6PoVNj5WJXUnMk+VEbELeXjtBkCS1UxTDa04
 github.com/go-openapi/runtime v0.19.26/go.mod h1:BvrQtn6iVb2QmiVXRsFAm6ZCAZBpbVKFfN6QWCp582M=
 github.com/go-openapi/runtime v0.19.27 h1:4zrQCJoP7rqNCUaApDv1MdPkaa5TuPfO05Lq5WLhX9I=
 github.com/go-openapi/runtime v0.19.27/go.mod h1:BvrQtn6iVb2QmiVXRsFAm6ZCAZBpbVKFfN6QWCp582M=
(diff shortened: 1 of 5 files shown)


go.sum
```diff
@@ -478,6 +478,7 @@ github.com/go-openapi/runtime v0.19.26 h1:K/6PoVNj5WJXUnMk+VEbELeXjtBkCS1UxTDa04
 github.com/go-openapi/runtime v0.19.26/go.mod h1:BvrQtn6iVb2QmiVXRsFAm6ZCAZBpbVKFfN6QWCp582M=
 github.com/go-openapi/runtime v0.19.27 h1:4zrQCJoP7rqNCUaApDv1MdPkaa5TuPfO05Lq5WLhX9I=
 github.com/go-openapi/runtime v0.19.27/go.mod h1:BvrQtn6iVb2QmiVXRsFAm6ZCAZBpbVKFfN6QWCp582M=
+github.com/go-openapi/runtime v0.19.28 h1:9lYu6axek8LJrVkMVViVirRcpoaCxXX7+sSvmizGVnA=
```
omertuc, line 481: Why?

src/installer/installer_test.go
```diff
@@ -400,6 +405,21 @@ var _ = Describe("installer HostRoleMaster role", func() {
 			ret := installerObj.InstallNode()
 			Expect(ret).Should(Equal(err))
 		})
+	})
```
omertuc, line 408: Did you intend to commit this?

eranco74: Yes, in the next line I created a new context `Context("Bootstrap role waiting for controller"...` that doesn't require the new `evaluateDiskSymlinkSuccess` mock
This way I could add the mock in all other relevant contexts instead of every test case.

go.sum
eranco74: God (Go mod) works in mysterious ways.

jakub-dzon (approved):

flaper87 (approved):

PR #266, opened 2021-04-20
Reduced UTs execution time by lowering the wait interval during the tests  

eranco74:


omertuc:
Do those ms sleeps make a difference? any reason to still keep them around? 

If they are required, I'm afraid that at this sleep resolution it might flake depending on machine performance

YuviGold:
Please elaborate on the change you made that causes a "fix"

omertuc:
> Please elaborate on the change you made that causes a "fix"

It's organized into 3 commits with details commit messages, good enough for me. Too bad GitHub hides it in the depths of the "Commits" tab

eranco74:
> Do those ms sleeps make a difference? any reason to still keep them around?
> 
> If they are required, I'm afraid that at this sleep resolution it might flake depending on machine performance.
> 
> Might be wrong though, because I don't really understand why they're there in the first place

I agree, in the installer there is no issue.
In the controller logs (where you see all the milliseconds) I can't say.
The controller is running multiple go routines so I preserve the ratio in the test between the interval and timeout just replaced X seconds with X * 10 milliseconds.
Seems to work just the same and save time.

PR #267, opened 2021-04-22
MGMT-4458 Remove old CompleteInstallation API

tsorya:
This api is doing nothing in assisted-service already
Removing it will allow us to spawn less error logs on kube-api
installation flow

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -258,14 +258,11 @@ func (c controller) PostInstallConfigs(wg *sync.WaitGroup, status *ControllerSta
 		break
 	}
 
-	errMessage := ""
 	err := c.postInstallConfigs()
 	if err != nil {
-		errMessage = err.Error()
 		status.Error()
+		c.log.WithError(err).Error("Failed to finish post install configurations")
 	}
-	success := err == nil
-	c.sendCompleteInstallation(success, errMessage)
 }
 
 func (c controller) postInstallConfigs() error {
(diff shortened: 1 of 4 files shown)


YuviGold:
/hold
This can't be done.
The service still depends on the controller to tell him of errors, even timeouts. Otherwise, the service will get stuck in an installation.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -731,19 +728,6 @@ func (c controller) waitingForClusterVersion() error {
 	return nil
 }
 
-func (c controller) sendCompleteInstallation(isSuccess bool, errorInfo string) {
-	c.log.Infof("Start complete installation step, with params success:%t, error info %s", isSuccess, errorInfo)
-	for {
-		ctx := utils.GenerateRequestContext()
-		if err := c.ic.CompleteInstallation(ctx, c.ClusterID, isSuccess, errorInfo); err != nil {
```
eranco74, line 738: What do you mean it does nothing?
The assisted-service CompleteInstallation API is no op?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -258,14 +258,11 @@ func (c controller) PostInstallConfigs(wg *sync.WaitGroup, status *ControllerSta
...
 
-	errMessage := ""
 	err := c.postInstallConfigs()
 	if err != nil {
-		errMessage = err.Error()
 		status.Error()
+		c.log.WithError(err).Error("Failed to finish post install configurations")
 	}
-	success := err == nil
-	c.sendCompleteInstallation(success, errMessage)
```
eranco74, line 268: If it's no longer required just replace it with a log:
c.log.Infof("Installation completed")

PR #270, opened 2021-04-27
OCPBUGSM-27367 Cancel controller when installation is cancelled

slaviered:
- Indication of log collection complete is sent on this case too
  - Save resources and irrelevant setup work on abandoned clusters

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build
 reports
 **/RCS/**
+.vscode
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -27,12 +27,12 @@ import (
 	"github.com/openshift/assisted-installer/src/k8s_client"
 	"github.com/openshift/assisted-installer/src/ops"
 	"github.com/openshift/assisted-installer/src/utils"
+	. "github.com/openshift/assisted-installer/src/utils/waiting"
 	"github.com/openshift/assisted-service/models"
 	mapiv1beta1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 )
 
 const (
-	generalWaitTimeoutInt     = 30
 	controllerLogsSecondsAgo  = 120 * 60
 	consoleOperatorName       = "console"
 	cvoOperatorName           = "cvo"
@@ -41,10 +41,8 @@ const (
 )
 
 var (
-	GeneralWaitInterval      = generalWaitTimeoutInt * time.Second
 	GeneralProgressUpdateInt = 60 * time.Second
 	LogsUploadPeriod         = 5 * time.Minute
-	WaitTimeout              = 70 * time.Minute
 )
 
 // assisted installer controller is added to control installation process after  bootstrap pivot
@@ -69,6 +67,8 @@ type Controller interface {
 
 type ControllerStatus struct {
 	errCounter uint32
+	mutex      *sync.Mutex
+	cancels    []context.CancelFunc
 }
 
 type controller struct {
@@ -89,12 +89,32 @@ func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inv
 	}
 }
 
+func NewControllerStatus() *ControllerStatus {
+	return &ControllerStatus{
+		mutex:   &sync.Mutex{},
+		cancels: make([]context.CancelFunc, 0),
+	}
+}
+
 func (status *ControllerStatus) Error() {
 	atomic.AddUint32(&status.errCounter, 1)
+	status.Abort()
 }
 
 func (status *ControllerStatus) HasError() bool {
 	return atomic.LoadUint32(&status.errCounter) > 0
+}
+
+func (status *ControllerStatus) Abort() {
+	for _, cancelFunc := range status.cancels {
+		cancelFunc()
+	}
+}
+
+func (status *ControllerStatus) AddWatch(cancelFunc context.CancelFunc) {
+	status.mutex.Lock()
+	status.cancels = append(status.cancels, cancelFunc)
+	status.mutex.Unlock()
 }
 
 func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.HostData) {
@@ -184,7 +204,7 @@ func (c *controller) getMCSLogs() (string, error) {
 		return "", nil
 	}
 	for _, pod := range pods {
-		podLogs, err := c.kc.GetPodLogs(namespace, pod.Name, generalWaitTimeoutInt*10)
+		podLogs, err := c.kc.GetPodLogs(namespace, pod.Name, int64(GeneralWaitTimeoutInt)*10)
 		if err != nil {
 			c.log.WithError(err).Warnf("Failed to get logs of pod %s", pod.Name)
 			return "", nil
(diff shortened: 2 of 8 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -127,7 +137,17 @@ func (c *controller) WaitAndUpdateNodesStatus(status *ControllerStatus) {
 
 		hostsInProgressMap := common.GetHostsInStatus(assistedNodesMap, []string{models.ClusterStatusInstalled}, false)
 		errNodesMap := common.GetHostsInStatus(hostsInProgressMap, []string{models.HostStatusError}, true)
+		cancelNodesMap := common.GetHostsInStatus(hostsInProgressMap, []string{models.HostStatusCancelled}, true)
```
tsorya, line 140: can't we use the same map as errors? why not to add cancelled status to https://github.com/openshift/assisted-installer/pull/270/files#diff-1da579fda3bc689940661d48dd744a46f23824270427207cfae4f8c7448eddd2R139
?

slaviered: It's possible under the assumption that cancel and error behaves the same (all hosts are moved to that status once it detected in the service)

tsorya: Can we validate it? I think it is same behavior

slaviered: Yes, I validated it and all host transit to either cancel or error. However, in the controller there is no unified behavior for cancel vs error: on error it tried to collect must-gather and on cancel it does not. This is why we need to have separate indication for cancel vs error

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -889,8 +909,8 @@ func (c *controller) UploadLogs(ctx context.Context, cancellog context.CancelFun
 				podName = pods[0].Name
 			}
 
-			if status.HasError() {
-				c.log.Infof("Error detected. Closing logs and aborting...")
+			if status.HasError() || status.Cancelled() {
```
tsorya, line 912: Maybe i forgot something but i don't remember why we have logic here? 
In any case i don't like the idea of adding new status. If this is needed somehow (i truly believe we can cancel logs from outside) i prefer to make status struct to handle cancel and error case together.

tsorya: why we can't just disable this statement 
https://github.com/openshift/assisted-installer/blob/master/src/main/assisted-installer-controller/assisted_installer_main.go#L84 ?

slaviered: After discussion I am implementing a mechanism to cancel all the go routines on cancel and error

src/installer/installer.go
```diff
@@ -346,13 +346,14 @@ func (i *installer) downloadHostIgnition() (string, error) {
 }
 
 func (i *installer) waitForNetworkType(kc k8s_client.K8SClient) error {
-	return utils.WaitForPredicate(waitForeverTimeout, 5*time.Second, func() bool {
+	err, _, _ := Wait().WithNoTimeout().WithInterval(5 * time.Second).WithPredicate(func() bool {
```
tsorya, line 349: why without timeout?

src/utils/waiting/waiting.go
```diff
@@ -0,0 +1,86 @@
...
+		interval: GeneralWaitInterval,
+	}
+}
+
+func (p *wait) WithPredicate(predicate func() bool) *wait {
+	p.predicate = predicate
+	return p
+}
+
+func (p *wait) WithTimeout(timeout time.Duration) *wait {
```
tsorya, line 36: where it is used?

src/utils/waiting/waiting.go
```diff
@@ -0,0 +1,86 @@
...
+	p.cancelctx = cancelctx
+	return p
+}
+
+func (p *wait) WithMessage(message string) *wait {
+	p.message = message
+	return p
+}
+
+func (p *wait) Start() (error, error, bool) {
```
tsorya, line 61: why 2 errors are needed? why does it matter if it was canceled or not? 
Cancel is coming from outside so what's point of verifying error if i just cancelled it by myself?

src/utils/waiting/waiting.go
```diff
@@ -0,0 +1,86 @@
+package waiting
```
tsorya, line 1: I must say that i don't really understand what this package gives us. It sounds like all we need one function and transformed it to object that still only one function from it is used? maybe i miss the point?

src/installer/installer.go
tsorya: i am not sure i really like the usage i must say. In any case i think it should start with WithPredicate and then adding with interval and etc. 
maybe we can just move to https://github.com/avast/retry-go ?

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,36 +55,48 @@ func main() {
...
 	var wgLogs sync.WaitGroup
-	var status assistedinstallercontroller.ControllerStatus
 
-	ctxApprove, cancelApprove := context.WithCancel(context.Background())
+	//every context should be added to the watch list before calling
+	//the routines because they may be canceled within those routines
+	//espceially, they should be called before WaitAndUpdateNodesStatus()
+	//that listen to the user's abort command
+
+	ctxApprove, cancelApprove := context.WithTimeout(context.Background(), 10*time.Minute)
```
tsorya, line 70: why not to have same context for all of them? except logs?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -89,14 +89,34 @@ func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inv
...
 	return atomic.LoadUint32(&status.errCounter) > 0
 }
 
+func (status *ControllerStatus) Abort() {
+	for _, cancelFunc := range status.cancels {
+		cancelFunc()
+	}
+}
+
+func (status *ControllerStatus) AddWatch(cancelFunc context.CancelFunc) {
```
tsorya, line 114: i don't understand this :( why?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -241,108 +261,130 @@ func isCsrApproved(csr *certificatesv1.CertificateSigningRequest) bool {
...
+	//if the condition for post install steps is not met within the timeout
+	//or the operation is canceled, return without configuring
+	if !ok {
+		return
 	}
 
-	errMessage := ""
-	err := c.postInstallConfigs()
+	var errMessage string
+	err, canceled := c.postInstallConfigs(ctx)
```
tsorya, line 284: you can just verify context was cancelled or not

.gitignore
```diff
@@ -1,3 +1,4 @@
 build
 reports
 **/RCS/**
+.vscode
```
eranco74, line 4: Please remove.
You can add it to `~/.gitignore_global`

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -776,14 +778,14 @@ var _ = Describe("installer HostRoleMaster role", func() {
 
 	})
 
-	Context("Upload logs with oc must-gather", func() {
+	FContext("Upload logs with oc must-gather", func() {
```
eranco74, line 781: What's Fcontext?

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -807,7 +810,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				}).AnyTimes()
 			reportLogProgressSuccess()
 		})
-		It("Validate upload logs (with must-gather logs)", func() {
+		FIt("Validate upload logs (with must-gather logs)", func() {
```
eranco74, line 813: Is this a thing?

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,36 +55,48 @@ func main() {
...
-	wg.Add(1)
-	go assistedController.UpdateBMHs(&wg)
+
+	go assistedController.UpdateBMHs(bmhconfig, &wg)
 	wg.Add(1)
 
-	ctxLogs, cancelLogs := context.WithCancel(context.Background())
-	go assistedController.UploadLogs(ctxLogs, cancelLogs, &wgLogs, &status)
+	go assistedController.UploadLogs(ctxLogs, &wgLogs, status)
 	wgLogs.Add(1)
```
eranco74, line 86: I think that now you don't need a special WaitGroup for the logs goroutine.

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,36 +55,48 @@ func main() {
...
+	wg.Add(1)
 
 	logger.Infof("Waiting for all go routines to finish")
 	wg.Wait()
-	if !status.HasError() {
-		//with error the logs are canceled within UploadLogs
-		logger.Infof("closing logs...")
-		cancelLogs()
-	}
+	// if !status.HasError() {
```
eranco74, line 96: Why comment out and not delete?

eranco74 (commented):
This seems like an overkill if all we want is to stop the controller in case of cancellation. 
Consider doing the refactoring in a separate PR.

slaviered:
> This seems like an overkill if all we want is to stop the controller in case of cancellation.
> Consider doing the refactoring in a separate PR.

How would you stop the multiple wait loops in a simpler way (i.e. without passing the context to each of the goroutine and check in every iteration of the wait loop that the controller is not aborted?)

PR #271, opened 2021-04-28
MGMT-4668: Add support to create manifest in assisted-installer

machacekondra:
This PR add a possibility to create manifests from assisted-installer.
In some cases it may be useful to create manifest from
assisted-isntaller instead of openshift-installer. For example when
installing CNV/LSO/OCS we need to have predefined CRDs for the
operators, because at the time of the operator is being created the CRDs
are not yet created by the OLM.

This PR add support to read the files from `custom_manifests.yaml` which is retrieved
from the assisted-service. The manifest are created via `oc create -f` command,
using the `kubeconfig-noingress` file.

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"math"
 	"net"
+	"os"
 	"path"
 	"strings"
 	"sync"
@@ -46,9 +47,11 @@ const (
 	maxDNSServiceIPAttempts   = 45
 	KeepWaiting               = false
 	ExitWaiting               = true
+	customManifestsFile       = "custom_manifests.yaml"
 )
 
 var (
+	retryPostManifestTimeout = 10 * time.Minute
 	GeneralWaitInterval      = generalWaitTimeoutInt * time.Second
 	GeneralProgressUpdateInt = 60 * time.Second
 	LogsUploadPeriod         = 5 * time.Minute
@@ -406,6 +409,13 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 		return errors.Errorf("Timeout while waiting for console to become available")
 	}
 
+	// Apply post install manifests
+	err = utils.WaitForPredicateWithContext(ctx, retryPostManifestTimeout, GeneralWaitInterval, c.applyPostInstallManifests)
+	if err != nil {
+		c.log.WithError(err).Warnf("Failed to apply post manifests.")
+		return err
+	}
+
 	waitTimeout := c.getMaximumOLMTimeout()
 	err = utils.WaitForPredicateWithContext(ctx, waitTimeout, GeneralWaitInterval, c.waitForOLMOperators)
 	if err != nil {
@@ -414,10 +424,38 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 		if err = c.updatePendingOLMOperators(); err != nil {
 			return errors.Errorf("Timeout while waiting for some of the operators and not able to update its state")
 		}
-		c.log.Warnf("Timeout while waiting for OLM operators be installed")
+		c.log.WithError(err).Warnf("Timeout while waiting for OLM operators be installed")
+		return err
 	}
 
 	return nil
+}
+
+func (c controller) applyPostInstallManifests() bool {
+	ctx := utils.GenerateRequestContext()
+	tempDir, err := ioutil.TempDir("", "controller-custom-manifests-")
+	if err != nil {
+		return false
+	}
+	defer os.RemoveAll(tempDir)
+
+	customManifestPath := path.Join(tempDir, customManifestsFile)
+	if err := c.ic.DownloadFile(ctx, customManifestsFile, customManifestPath); err != nil {
+		return false
+	}
+
+	kubeconfigName, err := c.downloadKubeconfigNoingress(ctx, tempDir)
+	if err != nil {
+		return false
+	}
+
+	err = c.ops.CreateManifests(kubeconfigName, customManifestPath)
+	if err != nil {
+		c.log.WithError(err).Error("Failed to apply manifest file.")
+		return false
+	}
+
+	return true
 }
 
 func (c controller) UpdateBMHs(ctx context.Context, wg *sync.WaitGroup) {
@@ -935,6 +973,19 @@ func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSec
 	return nil
 }
 
+func (c controller) downloadKubeconfigNoingress(ctx context.Context, dir string) (string, error) {
+	// Download kubeconfig file
+	kubeconfigFileName := "kubeconfig-noingress"
+	kubeconfigPath := path.Join(dir, kubeconfigFileName)
+	err := c.ic.DownloadFile(ctx, kubeconfigFileName, kubeconfigPath)
+	if err != nil {
+		c.log.Errorf("Failed to download noingress kubeconfig %v\n", err)
+		return "", err
+	}
+
+	return kubeconfigPath, nil
+}
+
 func (c controller) collectMustGatherLogs(ctx context.Context, mustGatherImg string) (string, error) {
 	tempDir, ferr := ioutil.TempDir("", "controller-must-gather-logs-")
 	if ferr != nil {
@@ -942,12 +993,8 @@ func (c controller) collectMustGatherLogs(ctx context.Context, mustGatherImg str
 		return "", ferr
 	}
 
-	//download kubeconfig file
-	kubeconfigFileName := "kubeconfig-noingress"
-	kubeconfigPath := path.Join(tempDir, kubeconfigFileName)
-	err := c.ic.DownloadFile(ctx, kubeconfigFileName, kubeconfigPath)
+	kubeconfigPath, err := c.downloadKubeconfigNoingress(ctx, tempDir)
 	if err != nil {
-		c.log.Errorf("Failed to download noingress kubeconfig %v\n", err)
 		return "", err
 	}
 
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -47,6 +47,7 @@ type Ops interface {
 	CreateRandomHostname(hostname string) error
 	GetHostname() (string, error)
 	EvaluateDiskSymlink(string) string
+	CreateManifests(string, string) error
 }
 
 const (
@@ -565,4 +566,15 @@ func (o *ops) CreateRandomHostname(hostname string) error {
 
 func (o *ops) GetHostname() (string, error) {
 	return os.Hostname()
+}
+
+func (o *ops) CreateManifests(kubeconfig string, manifestFilePath string) error {
+	command := fmt.Sprintf("oc --kubeconfig=%s apply -f %s", kubeconfig, manifestFilePath)
+	output, err := o.ExecCommand(o.logWriter, "bash", "-c", command)
+	if err != nil {
+		return err
+	}
+	o.log.Info(output)
+
+	return nil
 }
--- a/src/assisted_installer_controller/assisted_installer_controller_test.go
+++ b/src/assisted_installer_controller/assisted_installer_controller_test.go
@@ -545,6 +545,9 @@ var _ = Describe("installer HostRoleMaster role", func() {
 					Return(&models.MonitoredOperator{Status: models.OperatorStatusAvailable, StatusInfo: availableClusterVersionCondition.Status.Conditions[0].Message}, nil).Times(1)
 
 				// Completion
+				mockbmclient.EXPECT().DownloadFile(gomock.Any(), "custom_manifests.yaml", gomock.Any()).Return(nil).Times(1)
+				mockbmclient.EXPECT().DownloadFile(gomock.Any(), "kubeconfig-noingress", gomock.Any()).Return(nil).Times(1)
+				mockops.EXPECT().CreateManifests(gomock.Any(), gomock.Any()).Return(nil).Times(1)
 				mockbmclient.EXPECT().GetClusterMonitoredOLMOperators(gomock.Any(), gomock.Any()).Return([]models.MonitoredOperator{}, nil).Times(2)
 				mockbmclient.EXPECT().CompleteInstallation(gomock.Any(), "cluster-id", true, "").Return(fmt.Errorf("dummy")).Times(1)
 				mockbmclient.EXPECT().CompleteInstallation(gomock.Any(), "cluster-id", true, "").Return(nil).Times(1)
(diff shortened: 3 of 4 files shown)


pkliczewski:
@eranco74 @tsorya please review

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -380,6 +380,12 @@ func (c controller) postInstallConfigs() error {
 		return errors.Errorf("Timeout while waiting for console to become available")
 	}
 
+	// Aplly post install manifests:
+	err = c.applyPostInstallManifests()
+	if err != nil {
+		c.log.WithError(err).Warnf("Failed to apply post manifests.")
```
ori-amizur, line 386: Shouldn't we return error ?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -394,6 +400,32 @@ func (c controller) postInstallConfigs() error {
 	return nil
 }
 
+func (c controller) applyPostInstallManifests() error {
+	ctx := utils.GenerateRequestContext()
+	dest, err := ioutil.TempFile("", "custom_manifests")
+	if err != nil {
+		return err
+	}
+	defer dest.Close()
```
ori-amizur, line 409: I think the dest file should be closed immediately and the file should be deleted when exiting the function  (by defer) since it is not used any more.

pkliczewski:
@machacekondra please rebase

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Right, will do, thanks.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Nope, we should try to create as much manifests as possible and ignore those which failed, but log the error, so user is aware.

ronniel1: The user in this case is the user of the Assisted-Installer.
Logging the error will not show this to the user

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -418,7 +419,42 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 		if err = c.updatePendingOLMOperators(); err != nil {
 			return errors.Errorf("Timeout while waiting for some of the operators and not able to update its state")
 		}
-		c.log.Warnf("Timeout while waiting for OLM operators be installed")
+		c.log.WithError(err).Warnf("Timeout while waiting for OLM operators be installed")
+	}
+
+	// Aplly post install manifests:
+	err = c.applyPostInstallManifests()
```
ronniel1, line 426: Users do not see logs .
If we fail applying manifests for an operator, we should probably mark that operator as failed, not the entire cluster installation

machacekondra: Sure, we do that, that's why we don't return error at line 428. We update the status of the operator [here](https://github.com/openshift/assisted-installer/blob/4f5ad6a52b0fa1bcacb19be70aaeedb09870dda0/src/assisted_installer_controller/assisted_installer_controller.go#L739)

machacekondra: The reason we don't return error here and also on line 422 is because the [PostInstallConfig](https://github.com/openshift/assisted-installer/blob/bc513cf9e7188afac3fbbc4d55e8228542347ee4/src/assisted_installer_controller/assisted_installer_controller.go#L851) method decide whether to set the installation as successful based on the error returned from this method. So we do not return the error on purpose. If we would return error here the installation would be marked as failed. If we just log the error here and we update the status of the operators in assisted-service API with related error, we can mark the installation as successful, and report what's wrong with monitored operators.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -418,7 +419,44 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	if err := c.ic.DownloadFile(ctx, "custom_manifests.yaml", dest.Name()); err != nil {
+		return err
+	}
+
+	// download kubeconfig file
+	kubeconfigFileName := "kubeconfig-noingress"
```
tsorya, line 450: don't we have such code already? if we do lets have it in one place. I any case lets move it to function

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: @machacekondra right now controller is not deciding if installation is successful, it only decides if it is not.
So if you will return error here it will not really matter. In any case we have ticket to remove send complete api from here.
@YuviGold can you please assist here?

tsorya: But main problem that this is not the right place to add this code. It must be before 
https://github.com/openshift/assisted-installer/pull/271/files#diff-1da579fda3bc689940661d48dd744a46f23824270427207cfae4f8c7448eddd2L414
or even part of waitForOLMOperators (run only once on success though)

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -418,7 +419,44 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+	// Aplly post install manifests:
+	err = c.applyPostInstallManifests()
+	if err != nil {
+		c.log.WithError(err).Warnf("Failed to apply post manifests.")
+	}
+
+	return nil
+}
+
+func (c controller) applyPostInstallManifests() error {
```
tsorya, line 436: all this logic should be retriable

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Yeah, there is must-gather, I will make it reusable.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Done

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -46,9 +46,11 @@ const (
 	maxDNSServiceIPAttempts   = 45
 	KeepWaiting               = false
 	ExitWaiting               = true
+	customManifestsFile       = "custom_manifests.yaml"
 )
 
 var (
+	retryPostManifestTimeout = 5 * time.Minute
```
tsorya, line 53: enough? not sure how much time takes to olm though

machacekondra: I've increased it, just to be sure.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Based on the discussion I understand, this is really not important. So should I change or keep? @tsorya

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -418,9 +426,39 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	customManifestPath := path.Join(tempDir, customManifestsFile)
+	if err := c.ic.DownloadFile(ctx, customManifestsFile, customManifestPath); err != nil {
+		return err
+	}
+
+	kubeconfigName, err := c.downloadKubeconfigNoingress(ctx, tempDir)
```
tsorya, line 449: will we retry in case of failure?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -418,9 +426,39 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+		// Don't return the error here so the installation is marked as sucessful.
+		// We only log the error, and update the state of the monitored operators in assisted-service API.
+	}
+
+	return nil
+}
+
+func (c controller) applyPostInstallManifests() error {
+	ctx := utils.GenerateRequestContext()
+	tempDir, err := ioutil.TempDir("", "controller-custom-manifests-")
```
tsorya, line 439: not sure what can happened but will we retry on error?

src/ops/ops.go
```diff
@@ -566,3 +567,14 @@ func (o *ops) CreateRandomHostname(hostname string) error {
 func (o *ops) GetHostname() (string, error) {
 	return os.Hostname()
 }
+
+func (o *ops) CreateManifests(kubeconfig string, manifestFilePath string) error {
+	command := fmt.Sprintf("oc --kubeconfig=%s apply -f %s", kubeconfig, manifestFilePath)
+	output, err := o.ExecCommand(o.logWriter, "bash", "-c", command)
+	if err != nil {
+		return err
```
tsorya, line 575: will we have log in case of failure? i mean if ExecCommand is logging failure?

machacekondra: No. Thanks, will add.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: We retry only on `oc apply` failures, do you think we should retry in all the errors?

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: We retry only on `oc apply` failures, do you think we should retry in all the errors?

tsorya: i think we should cause on any failure we will fail installation.

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: yep

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: for example service can be not reachable for a minute and we will fail the installation cause of it?

machacekondra: Right, will re-work.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -410,6 +413,13 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 		return errors.Errorf("Timeout while waiting for console to become available")
 	}
 
+	// Aplly post install manifests:
+	err = utils.WaitForPredicateWithContext(ctx, retryPostManifestTimeout, GeneralWaitInterval, c.applyPostInstallManifests)
+	if err != nil {
+		c.log.WithError(err).Warnf("Failed to apply post manifests.")
+		return err
```
machacekondra, line 420: @tsorya I think it's enough to return the error here, because we [fail](https://github.com/openshift/assisted-service/blob/master/internal/bminventory/inventory.go#L3976) installation in case assisted-installer  `postInstallConfigs` method return error. Why do you think we should move it before console?

tsorya: Assisted-service will set installed from the point controller will send that console operator is fine. 
Controller doesn't set installed by itself. (and error logic should be moved to service too, we have ticket for it somewhere).
If you will return here error we will move cluster to degraded i think but maybe i miss something
@YuviGold you know this logic the best, can you assist please

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -410,6 +413,13 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 		return errors.Errorf("Timeout while waiting for console to become available")
 	}
 
+	// Aplly post install manifests:
```
YuviGold, line 416:
```suggestion
	// Apply post install manifests
```

src/assisted_installer_controller/assisted_installer_controller.go
YuviGold: > Nope, we should try to create as much manifests as possible and ignore those which failed, but log the error, so user is aware.

Wouldn't it affect the inspecting of the OLM operators? The service expects the OLM statuses to be either "available" or "failure". In case there will be no status sent - it will wait until timeout, and the user will waste a lot of time

src/assisted_installer_controller/assisted_installer_controller.go
YuviGold: In case you return an error - the controller will return an error and abort the installation (=cluster will move to "error" state).
In case it's a specific operator manifest that failed to be applied - its status should be updated with "failure".
Otherwise, if it's a general malfunction - I think it means that something isn't right with the cluster - and we can abort the installation. @tsorya WDYT?

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: That's correct, but in case we return an error here, the assisted-service will report failed installation, and the state of the OLM is not important, if I am not mistaken.

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: @machacekondra sorry i forgot this logic. we will not set installed after console was available. Totally forgot, so yes your logic is totally right

machacekondra: @tsorya @YuviGold so do we agree that current code is fine. Or should we update the monitor operator states even if custom manifests fails?

tsorya: we agree it is fine for this version. Let rework all retry stuff and push it

src/assisted_installer_controller/assisted_installer_controller.go
eranco74: We should retry in all errors but we should avoid repeating successful steps, meaning if you already downloaded the files there is no need to download them again.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: @tsorya  The retry code is there on line 413.

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: event sound nice but i think we already agreed to fail installation with relevant reason

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -410,6 +413,13 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+	err = utils.WaitForPredicateWithContext(ctx, retryPostManifestTimeout, GeneralWaitInterval, c.applyPostInstallManifests)
```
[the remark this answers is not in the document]
tsorya: we need some timeout cause in this status service is not managing timeout for us

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: @eranco74 it was agreed to fail the installation for now

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: sounds good idea to delete file only on success and to check if it exists before download.
Though can be optimization in future PR

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: @machacekondra i meant retry on all the errors and not apply one

machacekondra: It is. It is retry on all the errors that can ocure in applyManifest method, including downlaoding kubeapi or creating tempdir.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: I just wonder how often this could happen, so it should be optimalized?

PR #279, opened 2021-05-09
MGMT-4992 - stop controller after 10 errors of 401/404.

tsorya:
MGMT-[OCPBUGSM-27367](https://issues.redhat.com/browse/OCPBUGSM-27367) Reset cluster mentioned 5/6 installation logs were collected  but when downloading there are 6

Controller will have monitoring loop that will monitor connection to
service. If cluster will be cancelled or installer monitor will not set
error and exit. On cluster error monitor will set error that will say to
upload logs to upload must gather.
If 10 401/404 errors will occured in a row, it will exit 0

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/coreos/ignition/v2 v2.10.1
 	github.com/go-openapi/runtime v0.19.28
 	github.com/go-openapi/strfmt v0.20.1
+	github.com/go-openapi/swag v0.19.9
 	github.com/golang/mock v1.5.0
 	github.com/google/uuid v1.2.0
 	github.com/hashicorp/go-version v1.3.0
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -51,8 +51,10 @@ var (
 	GeneralProgressUpdateInt = 60 * time.Second
 	LogsUploadPeriod         = 5 * time.Minute
 	WaitTimeout              = 70 * time.Minute
+	CompleteTimeout          = 30 * time.Minute
 	DNSAddressRetryInterval  = 20 * time.Second
 	DeletionRetryInterval    = 10 * time.Second
+	LongWaitTimeout          = 10 * time.Hour
 )
 
 // assisted installer controller is added to control installation process after  bootstrap pivot
@@ -116,79 +118,96 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
 	log.Infof("Hosts status: %v", hostsStatus)
 }
 
-func (c *controller) WaitAndUpdateNodesStatus(status *ControllerStatus) {
+// WaitAndUpdateNodesStatus waits till all nodes joins the cluster and become ready
+// it will update joined/done status
+// approve csr will run as routine and cancelled whenever all nodes are ready and joined
+// this will allow to run it and end only when it is needed
+func (c *controller) WaitAndUpdateNodesStatus(ctx context.Context, wg *sync.WaitGroup) {
+	approveCtx, approveCancel := context.WithCancel(ctx)
+	defer func() {
+		approveCancel()
+		c.log.Infof("WaitAndUpdateNodesStatus finished")
+		wg.Done()
+	}()
+	// starting approve csrs
+	go c.ApproveCsrs(approveCtx)
 
 	c.log.Infof("Waiting till all nodes will join and update status to assisted installer")
+	_ = utils.WaitForPredicateWithContext(ctx, LongWaitTimeout, GeneralWaitInterval, c.waitAndUpdateNodesStatus)
+}
+
+func (c *controller) waitAndUpdateNodesStatus() bool {
 	ignoreStatuses := []string{models.HostStatusDisabled}
 	var hostsInError int
-	for {
-		time.Sleep(GeneralWaitInterval)
-		ctx := utils.GenerateRequestContext()
-		log := utils.RequestIDLogger(ctx, c.log)
+	ctxReq := utils.GenerateRequestContext()
+	log := utils.RequestIDLogger(ctxReq, c.log)
 
-		assistedNodesMap, err := c.ic.GetHosts(ctx, log, ignoreStatuses)
-		if err != nil {
-			log.WithError(err).Error("Failed to get node map from the assisted service")
+	assistedNodesMap, err := c.ic.GetHosts(ctxReq, log, ignoreStatuses)
+	if err != nil {
+		log.WithError(err).Error("Failed to get node map from the assisted service")
+		return false
+	}
+
+	logHostsStatus(log, assistedNodesMap)
+
+	hostsInProgressMap := common.GetHostsInStatus(assistedNodesMap, []string{models.HostStatusInstalled}, false)
+	errNodesMap := common.GetHostsInStatus(hostsInProgressMap, []string{models.HostStatusError}, true)
+	hostsInError = len(errNodesMap)
+
+	//if all hosts are in error, mark the failure and finish
+	if hostsInError > 0 && hostsInError == len(hostsInProgressMap) {
+		c.log.Infof("Done waiting for all the nodes. Nodes in error status: %d\n", hostsInError)
+		return true
+	}
+	//if all hosts are successfully installed, finish
+	if len(hostsInProgressMap) == 0 {
+		c.log.Infof("All nodes were successfully installed")
+		return true
+	}
+	//otherwise, update the progress status and keep waiting
+	log.Infof("Checking if cluster nodes are ready. %d nodes remaining", len(hostsInProgressMap))
+	nodes, err := c.kc.ListNodes()
+	if err != nil {
+		log.WithError(err).Error("Failed to get list of nodes from k8s client")
+		return false
+	}
+	for _, node := range nodes.Items {
+		host, ok := hostsInProgressMap[strings.ToLower(node.Name)]
+		if !ok {
+			if _, ok := assistedNodesMap[strings.ToLower(node.Name)]; !ok {
+				log.Warnf("Node %s is not in inventory hosts", strings.ToLower(node.Name))
+			}
+
 			continue
 		}
-
-		logHostsStatus(log, assistedNodesMap)
-
-		hostsInProgressMap := common.GetHostsInStatus(assistedNodesMap, []string{models.ClusterStatusInstalled}, false)
-		errNodesMap := common.GetHostsInStatus(hostsInProgressMap, []string{models.HostStatusError}, true)
-		hostsInError = len(errNodesMap)
-
-		//if all hosts are in error, mark the failure and finish
-		if hostsInError > 0 && hostsInError == len(hostsInProgressMap) {
-			status.Error()
-			break
-		}
-		//if all hosts are successfully installed, finish
-		if len(hostsInProgressMap) == 0 {
-			break
-		}
-		//otherwise, update the progress status and keep waiting
-		log.Infof("Checking if cluster nodes are ready. %d nodes remaining", len(hostsInProgressMap))
-		nodes, err := c.kc.ListNodes()
-		if err != nil {
-			log.WithError(err).Error("Failed to get list of nodes from k8s client")
-			continue
-		}
-		for _, node := range nodes.Items {
-			host, ok := hostsInProgressMap[strings.ToLower(node.Name)]
-			if !ok {
-				if _, ok := assistedNodesMap[strings.ToLower(node.Name)]; !ok {
-					log.Warnf("Node %s is not in inventory hosts", strings.ToLower(node.Name))
-				}
-
+		if common.IsK8sNodeIsReady(node) {
+			log.Infof("Found new ready node %s with inventory id %s, kubernetes id %s, updating its status to %s",
+				node.Name, host.Host.ID.String(), node.Status.NodeInfo.SystemUUID, models.HostStageDone)
+			if err := c.ic.UpdateHostInstallProgress(ctxReq, host.Host.ID.String(), models.HostStageDone, ""); err != nil {
+				log.WithError(err).Errorf("Failed to update node %s installation status", node.Name)
 				continue
 			}
-			if common.IsK8sNodeIsReady(node) {
-				log.Infof("Found new ready node %s with inventory id %s, kubernetes id %s, updating its status to %s",
-					node.Name, host.Host.ID.String(), node.Status.NodeInfo.SystemUUID, models.HostStageDone)
-				if err := c.ic.UpdateHostInstallProgress(ctx, host.Host.ID.String(), models.HostStageDone, ""); err != nil {
-					log.WithError(err).Errorf("Failed to update node %s installation status", node.Name)
-					continue
-				}
-			} else if host.Host.Progress.CurrentStage == models.HostStageConfiguring {
-				log.Infof("Found new joined node %s with inventory id %s, kubernetes id %s, updating its status to %s",
-					node.Name, host.Host.ID.String(), node.Status.NodeInfo.SystemUUID, models.HostStageJoined)
-				if err := c.ic.UpdateHostInstallProgress(ctx, host.Host.ID.String(), models.HostStageJoined, ""); err != nil {
-					log.WithError(err).Errorf("Failed to update node %s installation status", node.Name)
-					continue
-				}
+		} else if host.Host.Progress.CurrentStage == models.HostStageConfiguring {
+			log.Infof("Found new joined node %s with inventory id %s, kubernetes id %s, updating its status to %s",
+				node.Name, host.Host.ID.String(), node.Status.NodeInfo.SystemUUID, models.HostStageJoined)
+			if err := c.ic.UpdateHostInstallProgress(ctxReq, host.Host.ID.String(), models.HostStageJoined, ""); err != nil {
+				log.WithError(err).Errorf("Failed to update node %s installation status", node.Name)
+				continue
 			}
 		}
-		c.updateConfiguringStatusIfNeeded(assistedNodesMap)
 	}
-	c.log.Infof("Done waiting for all the nodes. Nodes in error status: %d\n", hostsInError)
+	c.updateConfiguringStatusIfNeeded(assistedNodesMap)
+	return false
 }
 
 func (c *controller) HackDNSAddressConflict(wg *sync.WaitGroup) {
 
 	c.log.Infof("Making sure service %s can reserve the .10 address", dnsServiceName)
 
-	defer wg.Done()
+	defer func() {
+		c.log.Infof("HackDNSAddressConflict finished")
+		wg.Done()
+	}()
 	networks, err := c.kc.GetServiceNetworks()
 	if err != nil || len(networks) == 0 {
 		c.log.Errorf("Failed to get service networks: %s", err)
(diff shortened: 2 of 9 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -116,79 +118,92 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
 	log.Infof("Hosts status: %v", hostsStatus)
 }
 
-func (c *controller) WaitAndUpdateNodesStatus(status *ControllerStatus) {
+func (c *controller) WaitAndUpdateNodesStatus(ctx context.Context, wg *sync.WaitGroup) {
+	approveCtx, approveCancel := context.WithCancel(ctx)
```
slaviered, line 122: handling CSR within another go routine is very confusing and hard to maintain. This should be a goroutine fired from main() and the way to handle multiple context cancel is with the cancel listener

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -116,79 +118,92 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
...
+	defer func() {
+		approveCancel()
+		c.log.Infof("WaitAndUpdateNodesStatus finished")
+		wg.Done()
+	}()
+	// starting approve csrs
+	go c.ApproveCsrs(approveCtx)
 
 	c.log.Infof("Waiting till all nodes will join and update status to assisted installer")
+	_ = utils.WaitForPredicateWithContext(ctx, LongWaitTimeout, GeneralWaitInterval, c.waitAndUpdateNodesStatus)
```
slaviered, line 132: Why did you increase the wait for 3 hours? This logic runs prior to operators updates and the condition should be met in reasonable time. You will get a wait-for-control-plane timeout much sooner from the service which will result in terminating the thread all together

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -116,79 +118,92 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
 	log.Infof("Hosts status: %v", hostsStatus)
 }
 
-func (c *controller) WaitAndUpdateNodesStatus(status *ControllerStatus) {
+func (c *controller) WaitAndUpdateNodesStatus(ctx context.Context, wg *sync.WaitGroup) {
+	approveCtx, approveCancel := context.WithCancel(ctx)
+	defer func() {
+		approveCancel()
```
slaviered, line 124: This is a change from the previous logic where the CSR still lived 10 minutes after the node scan finished. And in OCP there is no timeout at all. Now it will be terminated exactly when the wait for node finishes. Is this done on purpose?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -116,79 +118,92 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
...
+	return false
 }
 
 func (c *controller) HackDNSAddressConflict(wg *sync.WaitGroup) {
 
 	c.log.Infof("Making sure service %s can reserve the .10 address", dnsServiceName)
 
-	defer wg.Done()
+	defer func() {
+		c.log.Infof("HackDNSAddressConflict finished")
```
slaviered, line 204: Most chances is that these shutdown logs will not show in the log file because the log routine is terminated as well

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -279,8 +294,7 @@ func (c *controller) updateConfiguringStatusIfNeeded(hosts map[string]inventory_
 	common.SetConfiguringStatusForHosts(c.ic, hosts, logs, false, c.log)
 }
 
-func (c *controller) ApproveCsrs(ctx context.Context, wg *sync.WaitGroup) {
-	defer wg.Done()
+func (c *controller) ApproveCsrs(ctx context.Context) {
```
slaviered, line 297: This is exactly the logic of WaitForPredicateWithCancel and it is better to use one function

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -318,46 +332,56 @@ func isCsrApproved(csr *certificatesv1.CertificateSigningRequest) bool {
...
+	if err != nil {
+		return
 	}
 
 	errMessage := ""
-	err := c.postInstallConfigs()
+	err = c.postInstallConfigs(ctx)
+	// context was cancelled
+	// no reason to set error
+	if ctx.Err() != nil {
```
slaviered, line 361: This distinction works only if we do not use context with Timeout but the existing predicate utility function. So it's need to be documented

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: it will make much more complex to maintain it from main. This routine is relevant to WaitAndUpdateNodesStatus and depends on it. When WaitAndUpdateNodesStatus exits, i want to exit this one. I think it is better to have it here. Not sure why it is bad to call go routine from function?

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: This is what was expected. After all nodes are added there is no point to have approve csr up.

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,38 +62,77 @@ func main() {
...
-	assistedController.WaitAndUpdateNodesStatus(&status)
-	logger.Infof("Sleeping for 10 minutes to give a chance to approve all csrs")
-	time.Sleep(10 * time.Minute)
-	cancelApprove()
+	waitForInstallation(client, logger, &status)
+	// stop all go routines
+	cancelRoutines()
 
 	logger.Infof("Waiting for all go routines to finish")
 	wg.Wait()
```
slaviered, line 92: you do need the wait group anymore because waitAndUpdateNodesStatus() serves as the blocking wait routine and cancel routine is what release the wait

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: till now there was no timeout for this one. 
I decreased timeout from run_for_ever to 10 hours

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: it is made for checking logs inside pod. It is helpful for local debug or when you have connection to cluster

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,38 +62,77 @@ func main() {
...
-	wgLogs.Add(1)
+	go assistedController.UploadLogs(ctxRoutines, &wg, &status)
+	wg.Add(1)
 
-	assistedController.SetReadyState()
-	assistedController.WaitAndUpdateNodesStatus(&status)
-	logger.Infof("Sleeping for 10 minutes to give a chance to approve all csrs")
-	time.Sleep(10 * time.Minute)
-	cancelApprove()
+	waitForInstallation(client, logger, &status)
```
slaviered, line 87: I strongly prefer the listener option, but if you implement this this way at least document that returning from this function means that we either completed or aborted and there is nothing that should come between this routine an cancel context

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: just don't want to change places when it was needed. We can do it in another PR

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,38 +62,77 @@ func main() {
...
 	logger.Infof("Waiting for all go routines to finish")
 	wg.Wait()
-	if !status.HasError() {
-		//with error the logs are canceled within UploadLogs
-		logger.Infof("closing logs...")
-		cancelLogs()
+	logger.Infof("Finished all")
+}
+
+func waitForInstallation(client inventory_client.InventoryClient, log logrus.FieldLogger, status *assistedinstallercontroller.ControllerStatus) {
```
slaviered, line 96: This function should be in the controller file for better maintainability

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,38 +62,77 @@ func main() {
...
+	log.Infof("monitor cluster installation status")
+	reqCtx := utils.GenerateRequestContext()
+	errCounter := 0
+	for {
```
slaviered, line 100: predicate util is better

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: most of them will be part of the logs. I think only upload logs will be missed

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,38 +62,77 @@ func main() {
...
+			// exit all
+			if errCounter >= maximumErrorsBeforeExit {
+				log.Infof("Got more than %d errors from assisted service in a row, exiting", maximumErrorsBeforeExit)
+				exit(0)
+			}
+			continue
+		}
+		errCounter = 0
+		switch *cluster.Status {
+		case models.ClusterStatusError:
```
slaviered, line 126: Please add documentation

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,38 +62,77 @@ func main() {
...
+	reqCtx := utils.GenerateRequestContext()
+	errCounter := 0
+	for {
+		time.Sleep(assistedinstallercontroller.GeneralWaitInterval)
+		cluster, err := client.GetCluster(reqCtx)
+		if err != nil {
+			// In case cluster was deleted or controller is not authorised
+			// we should exit controller
+			switch err.(type) {
+			case *installer.GetClusterNotFound:
```
slaviered, line 107: I would rather not Exit here but gracefully terminate the context.However, if you remain with Exit at least put here a warning not to put here other error codes. This could result in abnormal termination of the controller in spotty network conditions with out completing the logs

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,38 +62,77 @@ func main() {
...
+				errCounter++
+				log.WithError(err).Errorf("Cluster was not found in inventory or user is not authorized")
+			case *installer.GetClusterUnauthorized:
```
slaviered, line 110: I recommend moving this to a retry utility or use a 3-rd party library

src/utils/utils.go
```diff
@@ -187,11 +187,17 @@ func GetHostIpsFromInventory(inventory *models.Inventory) ([]string, error) {
...
+	return WaitForPredicateWithContext(context.TODO(), timeout, interval, predicate)
+}
+
+func WaitForPredicateWithContext(ctx context.Context, timeout time.Duration, interval time.Duration, predicate func() bool) error {
 	timeoutAfter := time.After(timeout)
 	ticker := time.NewTicker(interval)
 	// Keep trying until we're time out or get true
 	for {
 		select {
+		case <-ctx.Done():
```
slaviered, line 199: return here context.Cancel error. It would be more standard and flexible. Future code can query the error

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/coreos/ignition/v2 v2.10.1
 	github.com/go-openapi/runtime v0.19.28
 	github.com/go-openapi/strfmt v0.20.1
+	github.com/go-openapi/swag v0.19.9
```
slaviered, line 13: Is adding swag necessary? or just simple de-referencing  can do

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -116,79 +118,92 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
...
+		return false
+	}
 
-		logHostsStatus(log, assistedNodesMap)
+	logHostsStatus(log, assistedNodesMap)
 
-		hostsInProgressMap := common.GetHostsInStatus(assistedNodesMap, []string{models.ClusterStatusInstalled}, false)
-		errNodesMap := common.GetHostsInStatus(hostsInProgressMap, []string{models.HostStatusError}, true)
-		hostsInError = len(errNodesMap)
+	hostsInProgressMap := common.GetHostsInStatus(assistedNodesMap, []string{models.ClusterStatusInstalled}, false)
```
slaviered, line 149: Shouldn't that be HostStatusInstalled?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -318,46 +332,56 @@ func isCsrApproved(csr *certificatesv1.CertificateSigningRequest) bool {
...
-		time.Sleep(GeneralWaitInterval)
-		ctx := utils.GenerateRequestContext()
+func (c controller) PostInstallConfigs(ctx context.Context, wg *sync.WaitGroup, status *ControllerStatus) {
+	defer func() {
+		c.log.Infof("Finished PostInstallConfigs")
+		wg.Done()
+	}()
+	err := utils.WaitForPredicateWithContext(ctx, LongWaitTimeout, GeneralWaitInterval, func() bool {
+		ctxReq := utils.GenerateRequestContext()
 		cluster, err := c.ic.GetCluster(ctx)
```
slaviered, line 342: There are 3 different goroutine here that query the service independently. I would have this query in a single goroutine and communicate the result to other threads via a channel

src/main/assisted-installer-controller/assisted_installer_main.go
tsorya: Not sure you understood the issue. There is no cluster in the system, there is nothing to update. No api calls will work. The idea is to just exit.

src/main/assisted-installer-controller/assisted_installer_main.go
tsorya: Not sure it is needed in this case. It is more visible what is going on. Another way i will need to pack in function and use it with retry package. I didn't like it, so i used this approach

go.mod
tsorya: i think we will need it in any case for unitests, not sure why not to use it and make test code more complex

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: optimazation. can be done another day

src/main/assisted-installer-controller/assisted_installer_main.go
tsorya: sure i need. If i will not wait for routines to finish they will not finish and controller will exit. It will cause upload logs not to send summary for example

src/main/assisted-installer-controller/assisted_installer_main.go
tsorya: i think it is very clear from the code

src/main/assisted-installer-controller/assisted_installer_main.go
tsorya: don't fully agree. I use exit(0) and i think better to have it in main

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: It's much more than mere optimization. It's quite a load. we will profile it and see the impact

src/main/assisted-installer-controller/assisted_installer_main.go
slaviered: As the reader of the code it was actually less visible. At least add detailed comment about the need and the reason for trying 10 times

src/main/assisted-installer-controller/assisted_installer_main.go
slaviered: This is the current issue. My concern is that in 3 months time someone will come and add conditions here that are not kubeapi related.

src/main/assisted-installer-controller/assisted_installer_main.go
tsorya: don't see big difference in this particular case. It waits forever so not sure what it will add.  I think this way it easier to test it

src/main/assisted-installer-controller/assisted_installer_main.go
slaviered: I'm afraid it is really not (and I actually knew the code before reviewing it)

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: It will depend on the random order of shutdown

src/main/assisted-installer-controller/assisted_installer_main.go
tsorya: Not sure i understand the complexity of it but i will document

src/main/assisted-installer-controller/assisted_installer_main.go
tsorya: I think that if you can't run apis , there is no need to try. What you are asking will require adding logic to routines to handle this specific case and that what i want to avoid

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: in any case we will wait for upload logs that will takes some time to finish, so we will have this log line in 90% of the cases. Other 10% will give better viability when we have connection to cluster

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: one api in a minute it is load? if yes we need to optimize service and not controller

src/main/assisted-installer-controller/assisted_installer_main.go
tsorya: I think "for" already implement retry logic, not sure that it is better to have retry inside retry .

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: Agree we can add more optimizations but i don't think it is scope of this PR

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -55,38 +65,85 @@ func main() {
...
+				errCounter++
+				log.WithError(err).Errorf("User is not authenticated to perform the operation")
+			}
+
+			// if we get maximumErrorsBeforeExit errors in a row
+			// there is no point to try to reach assisted service
+			// exit all
+			if errCounter >= maximumErrorsBeforeExit {
+				log.Infof("Got more than %d errors from assisted service in a row, exiting", maximumErrorsBeforeExit)
+				exit(0)
```
YuviGold, line 130: Document why we return zero value and not an invalid return value.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -116,79 +118,96 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
...
-		errNodesMap := common.GetHostsInStatus(hostsInProgressMap, []string{models.HostStatusError}, true)
-		hostsInError = len(errNodesMap)
+	hostsInProgressMap := common.GetHostsInStatus(assistedNodesMap, []string{models.HostStatusInstalled}, false)
+	errNodesMap := common.GetHostsInStatus(hostsInProgressMap, []string{models.HostStatusError}, true)
+	hostsInError = len(errNodesMap)
+
+	//if all hosts are in error, mark the failure and finish
+	if hostsInError > 0 && hostsInError == len(hostsInProgressMap) {
+		c.log.Infof("Done waiting for all the nodes. Nodes in error status: %d\n", hostsInError)
+		return true
```
YuviGold, line 160: Let's make typedefs for waitForPredicate
```
const (
     KeepWaiting = false
     ExitWaitingRoutine = true
)
```

PR #281, opened 2021-05-10
MGMT-3145 Reset CVO timeout in case of a change

YuviGold:
Reopen https://github.com/openshift/assisted-installer/pull/210

/cc @eranco74 @tsorya

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -800,7 +800,7 @@ func (c controller) validateConsoleAvailability() bool {
 // This function would be aligned with the console operator reporting workflow
 // as part of the deprecation of the old API in MGMT-5188.
 func (c controller) waitingForClusterVersion(ctx context.Context) error {
-	isClusterVersionAvailable := func() bool {
+	isClusterVersionAvailable := func(timer *time.Timer) bool {
 		c.log.Infof("Checking cluster version operator availability status")
 		co, err := c.kc.GetClusterVersion("version")
 		if err != nil {
@@ -822,6 +822,13 @@ func (c controller) waitingForClusterVersion(ctx context.Context) error {
 		operatorStatus, operatorMessage := utils.ClusterOperatorConditionsToMonitoredOperatorStatus(co.Status.Conditions)
 
 		if cvoStatusInService.Status != operatorStatus || (cvoStatusInService.StatusInfo != operatorMessage && operatorMessage != "") {
+			// This is a common pattern to ensure the channel is empty after a stop has been called
+			// More info on time/sleep.go documentation
+			if !timer.Stop() {
+				<-timer.C
+			}
+			timer.Reset(WaitTimeout)
+
 			status := fmt.Sprintf("Cluster version status: %s message: %s", operatorStatus, operatorMessage)
 			c.log.Infof(status)
 
@@ -834,9 +841,9 @@ func (c controller) waitingForClusterVersion(ctx context.Context) error {
 		return false
 	}
 
-	err := utils.WaitForPredicateWithContext(ctx, WaitTimeout, GeneralProgressUpdateInt, isClusterVersionAvailable)
+	err := utils.WaitForPredicateWithTimer(ctx, WaitTimeout, GeneralProgressUpdateInt, isClusterVersionAvailable)
 	if err != nil {
-		return errors.Errorf("Timeout while waiting for cluster version to be available")
+		return errors.Wrapf(err, "Timeout while waiting for cluster version to be available")
 	}
 	return nil
 }
--- a/src/utils/utils.go
+++ b/src/utils/utils.go
@@ -186,28 +186,40 @@ func GetHostIpsFromInventory(inventory *models.Inventory) ([]string, error) {
 	return ips, nil
 }
 
-func WaitForPredicate(timeout time.Duration, interval time.Duration, predicate func() bool) error {
-	return WaitForPredicateWithContext(context.TODO(), timeout, interval, predicate)
-}
-
-func WaitForPredicateWithContext(ctx context.Context, timeout time.Duration, interval time.Duration, predicate func() bool) error {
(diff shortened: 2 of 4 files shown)


eranco74 (commented):
/lgtm

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -351,7 +351,7 @@ func (c controller) postInstallConfigs() error {
 	c.log.Infof("Waiting for cluster version operator: %t", c.WaitForClusterVersion)
 
 	if c.WaitForClusterVersion {
-		err = c.waitingForClusterVersion()
+		err = c.waitingForClusterVersion(context.TODO())
```
slaviered, line 354: This must be the context of postInstallConfigs so it can be cancelled. I suggest you wait for @tsorya fix to be merged first because interfaces has changed here

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -789,6 +789,11 @@ func (c controller) waitingForClusterVersion() error {
 		operatorStatus, operatorMessage := utils.ClusterOperatorConditionsToMonitoredOperatorStatus(co.Status.Conditions)
 
 		if cvoStatusInService.Status != operatorStatus || (cvoStatusInService.StatusInfo != operatorMessage && operatorMessage != "") {
+			if !timer.Stop() {
```
slaviered, line 792: here also you should check for cancellation

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -801,9 +806,9 @@ func (c controller) waitingForClusterVersion() error {
 		return false
 	}
 
-	err := utils.WaitForPredicate(WaitTimeout, GeneralProgressUpdateInt, isClusterVersionAvailable)
+	err := utils.WaitForPredicateWithTimer(ctx, WaitTimeout, GeneralProgressUpdateInt, isClusterVersionAvailable)
```
slaviered, line 809: @tsorya is changing this very utility. You have a clash here. @tsorya this is the benefit of the DSL pattern that it does not care about adding dependencies

src/utils/utils.go
```diff
@@ -186,6 +186,26 @@ func GetHostIpsFromInventory(inventory *models.Inventory) ([]string, error) {
 	return ips, nil
 }
 
+func WaitForPredicateWithTimer(ctx context.Context, timeout time.Duration, interval time.Duration, predicate func(timer *time.Timer) bool) error {
```
slaviered, line 189: see my comment above about the cancel PR

src/assisted_installer_controller/assisted_installer_controller.go
YuviGold: Why?

src/utils/utils.go
```diff
@@ -186,6 +186,26 @@ func GetHostIpsFromInventory(inventory *models.Inventory) ([]string, error) {
...
+	timeoutTimer := time.NewTimer(timeout)
```
tsorya, line 190: defer timeoutTimer.Stop()

tsorya: maybe we should defer ticket.stop too? i think i missed it too

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -822,6 +822,11 @@ func (c controller) waitingForClusterVersion(ctx context.Context) error {
 		operatorStatus, operatorMessage := utils.ClusterOperatorConditionsToMonitoredOperatorStatus(co.Status.Conditions)
 
 		if cvoStatusInService.Status != operatorStatus || (cvoStatusInService.StatusInfo != operatorMessage && operatorMessage != "") {
+			if !timer.Stop() {
+				<-timer.C
```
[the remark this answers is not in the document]
tsorya: looks like it is common pattern.

src/utils/utils.go
tsorya: can we change usage of wait with context :
`func WaitForPredicateWithContext(ctx context.Context, timeout time.Duration, interval time.Duration, predicate func() bool) error {
	return WaitForPredicateWithTimer(ctx, timeout, interval, func(timer *time.Timer) bool {
		return predicate()
	})
}

func WaitForPredicateWithTimer(ctx context.Context, timeout time.Duration, interval time.Duration, predicate func(timer *time.Timer) bool) error {
	timeoutTimer := time.NewTimer(timeout)
	ticker := time.NewTicker(interval)
	// Keep trying until we're time out or get true
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		// Got a timeout! fail with a timeout error
		case <-timeoutTimer.C:
			return errors.New("timed out")
		// Got a tick, we should check on checkSomething()
		case <-ticker.C:
			if predicate(timeoutTimer) {
				return nil
			}
		}
	}
}`

src/utils/utils.go
```diff
@@ -186,6 +186,26 @@ func GetHostIpsFromInventory(inventory *models.Inventory) ([]string, error) {
...
+			return errors.New("timed out")
+		// Got a tick, we should check on checkSomething()
+		case <-ticker.C:
+			if predicate(timeoutTimer) {
+				return nil
+			}
+		}
+	}
+}
+
```
tsorya, line 208: can we change WaitForPredicateWithContext:
```suggestion
func WaitForPredicateWithContext(ctx context.Context, timeout time.Duration, interval time.Duration, predicate func() bool) error {
	return WaitForPredicateWithTimer(ctx, timeout, interval, func(timer *time.Timer) bool {
		return predicate()
	})
}
```

slaviered: +1 - now all loops has to have cancel option

src/assisted_installer_controller/assisted_installer_controller.go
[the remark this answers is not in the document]
YuviGold: I'll add a comment

PR #283, opened 2021-05-11
NO-ISSUE - removing assisted-installer-controller-ocp as it is not used

tsorya:


YuviGold:
Depends on https://github.com/openshift/release/pull/18478

PR #287, opened 2021-05-26
OCPBUGSM-29397 decrease waiting time in case of cluster was not found or not authorized

tsorya:
In case of 404 controller will exit immediately(up to one minute) in case of 401 controller will exit after 3 minutes


flaper87 (changes requested):
Change looks good.

Could you please fix the title and the description in *both* the commit and the PR? 

Here's a link to k8s guidelines, which I have been following for assisted-service: 

https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines

eranco74 (commented):
/lgtm

PR #288, opened 2021-05-26
MGMT-6584 Add host progress stages while waiting for control plane

eranco74:
When the bootstrap node host status is "waiting for control plane" it's actually waiting for multiple different things:

1. Wait for a minimal number of ready master nodes
2. Wait for bootkube to finish
3. Wait for the controller to be ready

After this change the assisted-service will reflect the above stages in the host progress


ronniel1:
@eranco74 Please provide a summary and description

PR #290, opened 2021-05-31
Refactor finalizing condition to be clearer

YuviGold:


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -350,11 +350,7 @@ func (c controller) PostInstallConfigs(ctx context.Context, wg *sync.WaitGroup,
 			utils.RequestIDLogger(ctxReq, c.log).WithError(err).Errorf("Failed to get cluster %s from assisted-service", c.ClusterID)
 			return false
 		}
-		// waiting till cluster will be installed(3 masters must be installed)
```
tsorya, line 353: fix comment but not delete it

PR #291, opened 2021-05-31
Bug 1966621: Do not use run-level label for assisted-installer namespace

mkowalski:
Namespaces using run-level label are considered to be highly privileged
and Security Context Constrains are not applied to the workload running
in them.

One of the deployment models for assisted-installer uses cluster
deployed by the AI service to deploy next clusters. In this scenario, if
the same `assisted-installer` namespace is used for deploying Assisted
Service Operator, the pods do not get any securityContext properties
applied. This, apart from the potential security violations, causes
functional errors as e.g. Postgres container is running using wrong UID.

This PR changes configuration of the `assisted-installer` namespace so
that it does not have run-level label applied and is treated like any
other customer namespace.

Closes: [OCPBUGSM-29833](https://issues.redhat.com/browse/OCPBUGSM-29833)

### DIFF
--- a/deploy/assisted-installer-controller/assisted-installer-controller-nm.yaml
+++ b/deploy/assisted-installer-controller/assisted-installer-controller-nm.yaml
@@ -2,5 +2,3 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: assisted-installer
-  labels:
-    openshift.io/run-level: "0"
--- a/deploy/assisted-installer-controller/assisted-installer-controller-role.yaml
+++ b/deploy/assisted-installer-controller/assisted-installer-controller-role.yaml
@@ -154,3 +154,17 @@ rules:
       - pods
     verbs:
       - deletecollection
+  - apiGroups:
+      - "security.openshift.io"
+    resourceNames:
+      - "anyuid"
+      - "nonroot"
+      - "hostmount-anyuid"
+      - "machine-api-termination-handler"
+      - "hostnetwork"
+      - "hostaccess"
+      - "node-exporter"
+    resources:
+      - securitycontextconstraints
+    verbs:
+      - use
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -434,19 +434,6 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 		}
 	}
 
-	// Unlabel run-level from assisted-installer namespace after the installation.
-	// Keeping the `run-level` label represents a security risk as it overwrites the SecurityContext configurations
-	// used for applications deployed in this namespace.
-	data := []byte(`{"metadata":{"labels":{"$patch": "delete", "openshift.io/run-level":"0"}}}`)
-	c.log.Infof("Removing run-level label from %s namespace", c.ControllerConfig.Namespace)
-	err = c.kc.PatchNamespace(c.ControllerConfig.Namespace, data)
-	if err != nil {
-		// It is a conscious decision not to fail an installation if for any reason patching the namespace
-		// in order to remove the `run-level` label has failed. This will be redesigned in the next release
-		// so that the `run-level` label is not created in the first place.
-		c.log.Warn("Failed to unlabel AI namespace after the installation.")
-	}
-
 	err = utils.WaitForPredicateWithContext(ctx, WaitTimeout, GeneralWaitInterval, c.addRouterCAToClusterCA)
 	if err != nil {
 		return errors.Wrapf(err, "Timeout while waiting router ca data")
--- a/src/assisted_installer_controller/assisted_installer_controller_test.go
+++ b/src/assisted_installer_controller/assisted_installer_controller_test.go
@@ -49,8 +49,6 @@ var (
 		MustGatherImage:       "quay.io/test-must-gather:latest",
 	}
 
-	aiNamespaceRunlevelPatch = []byte(`{"metadata":{"labels":{"$patch": "delete", "openshift.io/run-level":"0"}}}`)
-
 	progressClusterVersionCondition = &configv1.ClusterVersion{
 		Status: configv1.ClusterVersionStatus{
 			Conditions: []configv1.ClusterOperatorStatusCondition{{Type: configv1.OperatorProgressing,
@@ -644,9 +642,6 @@ var _ = Describe("installer HostRoleMaster role", func() {
 
 				setConsoleAsAvailable("cluster-id")
 
-				// Patching NS
-				mockk8sclient.EXPECT().PatchNamespace(defaultTestControllerConf.Namespace, aiNamespaceRunlevelPatch).Return(nil)
-
 				// CVO
 				mockbmclient.EXPECT().GetClusterMonitoredOperator(gomock.Any(), gomock.Any(), cvoOperatorName).
 					Return(&models.MonitoredOperator{Status: "", StatusInfo: ""}, nil).Times(1)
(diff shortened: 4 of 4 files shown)


mkowalski:
Ref.: https://access.redhat.com/solutions/4902471

flaper87:
/assign @tsorya 

Igal, since you created this file, was there a reason for `run-level` to be set? It's not needed in the hub cluster so I wonder why it's being set in the spoke cluster.

/approve

flaper87:
Commit that added this change: https://github.com/openshift/assisted-installer/commit/c9c293ed3b262e0c6c849f7f29ce4542dd53ff4a

tsorya:
@flaper87 i added it cause i wanted controller to start as quickly as possible and update host statuses. This flag allowed to do it. 
Already thought about removing it and it will work. We just need to understand that it will cause delay in updating statuses and will cause debugging to be more complicated due to it. It will increase installation time too.

flaper87:
> @flaper87 i added it cause i wanted controller to start as quickly as possible and update host statuses. This flag allowed to do it.
> Already thought about removing it and it will work. We just need to understand that it will cause delay in updating statuses and will cause debugging to be more complicated due to it. It will increase installation time too.

Gotcha! Thanks a lot for the context. 

Three options come to mind:

1. Don't re-use the same namespace for both, the controller and the assisted-service
2. Have the installer remove `run-level` once the installation is complete (or before running the AI oeprator/service)
3. Have the operator remove `run-level` before running AI.

Of all of the above, I'm leaning more on N. 2 as we would keep the current (desired) behavior and we will also clean up after ourselves allowing other parts of the flow to work as expected.

flaper87:
I would also like to clarify that I think using `run-level` for the purpose that it's being used right now is not ideal. I would prefer us to find a better way to achieve the same without using `run-level`.  

The options I proposed are to have a temporary fix if there's no easy/better way to get rid of `run-level` at this point

tsorya:
> I would also like to clarify that I think using `run-level` for the purpose that it's being used right now is not ideal. I would prefer us to find a better way to achieve the same without using `run-level`.
> 
> The options I proposed are to have a temporary fix if there's no easy/better way to get rid of `run-level` at this point

agree it is not idea but i don't think there is any other option
In anycase i think we can remove this flag and it should be ok.

mkowalski:
How about pod priorities? Wouldn't they put our workflow higher in the scheduler so that we could achieve a similar result?

flaper87:
> In anycase i think we can remove this flag and it should be ok.

If we can remove it and let things be, then let's go with it :)

tsorya:
@mkowalski hmm sounds like a nice idea but we will still need to wait till SCC will be pushed and that flag allowed to start much before it

tsorya:
if we want to apply this fix (though it can influence our success rate in cloud envs), we should set larger number 
here https://github.com/openshift/assisted-installer/blob/master/src/assisted_installer_controller/assisted_installer_controller.go#L285

mkowalski:
> we should set larger number
> here https://github.com/openshift/assisted-installer/blob/master/src/assisted_installer_controller/assisted_installer_controller.go#L285

Do we have any numbers or benchmarks on this? I have just run the installation with this patch and it succeeded without any timeouts on the way

mkowalski:
/hold

After discussing it's not the best idea to remove the flag at this moment now.

deads2k:
> @flaper87 i added it cause i wanted controller to start as quickly as possible and update host statuses. This flag allowed to do it.
> Already thought about removing it and it will work. We just need to understand that it will cause delay in updating statuses and will cause debugging to be more complicated due to it. It will increase installation time too.

SCCs are rendered during installation (applied during bootkube) and the namespace UID range annotations are added by the cluster-policy-controller in the bootstrap kube-controller-manager pod.  

Runlevels 0 should only be applied to namespaces that have to run pods before the kube-apiserver can start (basically nothing should require this) and runlevel 1 should only be applied to namespaces that to run pods before the openshift-apiserver or oauth-apiserver can run (this is stuff like the SDN).

Around 4.5/4.6 we decided that modest installation speed improvements are not enough of a reason for long-term, unnecessary security exposure.  This PR is in keeping with the posture of the platform and should be merged in the first possible release.

mkowalski:
@flaper87 let's get this one into the testing pipeline; I'll create a build containing this change and post it here

mkowalski:
`quay.io/mkowalski/assisted-installer-controller:bz1951812-pr291-23jun` is a controller build containing this change

flaper87:
/lgtm

@mkowalski you have my blessing to `lgtm` this PR if it loses the tag again :rofl: 

Let's get this in

mkowalski:
Images containing the current version of the code

```
quay.io/mkowalski/assisted-installer:bz1951812-pr291-08jul
quay.io/mkowalski/assisted-installer-controller:bz1951812-pr291-08jul
```

flaper87:
@mkowalski would it be possible to get a new build for this PR? thanks

mkowalski:
New images

```
quay.io/mkowalski/assisted-installer:bz1951812-pr291-02aug
quay.io/mkowalski/assisted-installer-controller:bz1951812-pr291-02aug
```

flaper87:
/lgtm

Feel free to unhold when you are ready to merge this PR!

flaper87:
/hold

too noisy, should check what's going on with the CI and unhold again

eranco74:
Seems that the controller isn't starting:
```
INFO     root:utils.py:192 Asked hosts to be in one of the statuses from ['installed'] and currently hosts statuses are [(1, 'a385da5f-7c6d-4f89-8cc5-e046e0722490', 'test-infra-cluster-a9b50993-worker-0', 'worker', 'installing-in-progress', 'Waiting for control plane'), (2, 'a7691745-751d-44fb-859a-4f16e31908d9', 'test-infra-cluster-a9b50993-worker-1', 'worker', 'installing-in-progress', 'Waiting for control plane'), (3, 'dec0c926-4f8c-4a47-b8c1-7d1f48dfe146', 'test-infra-cluster-a9b50993-master-1', 'master', 'installing-in-progress', 'Waiting for controller'), (4, '4d922a2c-9fa4-4ab3-9142-af2f52f29089', 'test-infra-cluster-a9b50993-master-0', 'master', 'installing-in-progress', 'Joined'), (5, 'e828ed86-810d-45f0-acb6-5fb80d8e9c13', 'test-infra-cluster-a9b50993-master-2', 'master', 'installing-in-progress', 'Joined')]
---------------------------- Captured log teardown -----------------------------
```

eranco74:
From the must-gather logs it seems that the assisted-installer-controller was never created
(no assisted-installer namespace)

eranco74:
From the installer log in the bootstrap node:
```
Sep 07 13:40:26 test-infra-cluster-a9b50993-master-1 installer[5633]: time="2021-09-07T13:40:26Z" level=info msg="Assisted controller new event: Error creating: pods \"assisted-installer-controller-\" is forbidden: unable to validate against any security context constraint: [provider \"anyuid\": Forbidden: not usable by user or serviceaccount, provider restricted: .spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used, spec.containers[0].securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used, provider \"nonroot\": Forbidden: not usable by user or serviceaccount, provider \"hostmount-anyuid\": Forbidden: not usable by user or serviceaccount, provider \"machine-api-termination-handler\": Forbidden: not usable by user or serviceaccount, provider \"hostnetwork\": Forbidden: not usable by user or serviceaccount, provider \"hostaccess\": Forbidden: not usable by user or serviceaccount, provider \"node-exporter\": Forbidden: not usable by user or serviceaccount, provider \"privileged\": Forbidden: not usable by user or serviceaccount]"

```
/hold
@mkowalski

mkowalski:
Ok, that's changed since the last iteration, I'll check what are the permissions that changed

flaper87:
/hold cancel

(I think @mkowalski forgot to  cancel it)

PR #294, opened 2021-06-02
Add check commit script

YuviGold:


hack/check-commit-message.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -o nounset
+
+commit_file=${1}
+commit_message="$(cat ${commit_file})"
```
celebdor, line 6:
```suggestion
commit_message="$(cat "${commit_file}")"
```

hack/check-commit-message.sh
```diff
@@ -0,0 +1,24 @@
...
+valid_commit_regex='^(NO-ISSUE|Merge|Revert|((Bug |[A-Z]+-|#)[0-9]+:))'
+
+error_msg="""Aborting commit: ${commit_message}
```
celebdor, line 9: I don't think """ has any meaning in bash. IIRC, this just means the following in python:

    error_msg = "" + """"Aborting commit: ${commit_message}

A single quote would be enough for multiline:
```suggestion
error_msg="Aborting commit: ${commit_message}
```

hack/check-commit-message.sh
```diff
@@ -0,0 +1,24 @@
...
+---
+Your commit message should start with a JIRA issue ('JIRA-1111'), a GitHub issue ('#39'), or a BugZilla issue ('Bug 123')
+with a following colon(:).
+i.e. 'MGMT-42: Summary of the commit message'
+You can also ignore the ticket checking with 'NO-ISSUE' for master only.
+
+Your message is preserved at '${commit_file}'
+"""
```
celebdor, line 17: Same as above, a single double quote would do:
```suggestion
"
```

hack/check-commit-message.sh
```diff
@@ -0,0 +1,24 @@
...
+
+status=$(echo "${commit_message}" | grep -qE "${valid_commit_regex}")
```
celebdor, line 19: I don't see status being used at all. What is it for?

hack/check-commit-message.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [ $? -gt 0 ]; then
```
celebdor, line 21: If the status was only to run the grep for this if, you can simply do:
```suggestion
if grep -qE "${valid_commit_regex}" <<< "$commit_message"; then
```
Instead of lines 19 and 21

hack/check-commits.sh
```diff
@@ -0,0 +1,26 @@
...
+
+__dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+parent_branch="master"
+git remote add openshift https://github.com/openshift/assisted-installer.git &> /dev/null || true
+git fetch openshift ${parent_branch}
+
+current_branch="$(git rev-parse --abbrev-ref HEAD)"
+
+revs=$(git rev-list "FETCH_HEAD".."${current_branch}")
```
celebdor, line 15: Why close the quotations after FETCH_HEAD? all the argument to `rev-list` should be a single string.
```suggestion
revs=$(git rev-list "FETCH_HEAD..${current_branch}")
```

hack/check-commits.sh
```diff
@@ -0,0 +1,26 @@
...
+
+for commit in ${revs};
+do
+    commit_message=$(git cat-file commit ${commit} | sed '1,/^$/d')
+    tmp_commit_file="$(mktemp)"
+    echo "${commit_message}" > ${tmp_commit_file}
```
celebdor, line 21:
```suggestion
    echo "${commit_message}" > "$tmp_commit_file"
```
In case the path to the tmp file has spaces

hack/check-commits.sh
```diff
@@ -0,0 +1,26 @@
...
+    ${__dir}/check-commit-message.sh "${tmp_commit_file}"
```
celebdor, line 22:
```suggestion
    "${__dir}/check-commit-message.sh" "${tmp_commit_file}"
```

In case __dir has spaces

celebdor (changes requested):

hack/check-commits.sh
```diff
@@ -0,0 +1,26 @@
...
+current_branch="$(git rev-parse --abbrev-ref HEAD)"
+
+revs=$(git rev-list "FETCH_HEAD..${current_branch}")
+
+for commit in ${revs};
+do
+    commit_message=$(git cat-file commit "${commit}" | sed '1,/^$/d')
+    tmp_commit_file="$(mktemp)"
+    echo "${commit_message}" > "${tmp_commit_file}"
+    "${__dir}/check-commit-message.sh" "${tmp_commit_file}"
```
celebdor, line 22: This works. You can also consider that if you don't want to deal with temporary files, you can always use variable redirection and process substitution
```suggestion
    "${__dir}/check-commit-message.sh" <(echo "$commit_message")
```

It can also be written as:
```
    "${__dir}/check-commit-message.sh" /dev/fd/64 64<<<"$commit_message"
```

nmagnezi (approved):
/lgtm

PR #301, opened 2021-06-04
Bug 1951812: Remove label from AI namespace after cluster installation

mkowalski:
# Description

Currently the `assisted-installer` namespace on the cluster installed
using Assisted Installer is labeled with run-level what causes any
workload running there to bypass security constrains. In a scenario when
an operator tries to reuse this namespace to deploy Assisted Service on
such a cluster, the installation will fail because services expect to be
run using the safe UID/GID ranges.

In order to overcome the issue, after the installation of the
cluster we will remove the label from the `assisted-installer` namespace
so that if an operator wants to deploy Assisted Service, it will be
necessary to recreate the namespace (this time without setting a run-level).

Please note that if for any reason removing the label fails, we will not
fail the installation of the cluster.

Long-term solution is to remove run-level label from the initially
created namespace and use another mechanism to ensure that the AI
controller starts as early as possible. Those efforts are tracked in bug
1966621.

Closes: [OCPBUGSM-28041](https://issues.redhat.com/browse/OCPBUGSM-28041)

# What environments does this code impact?

- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

# How was this code tested?

Please, select one or more if needed:

- [ ] assisted-test-infra environment
- [x] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Manual test

I have deployed a SNO using code from this PR. After the installation I have the following state of the `assisted-installer` namespace

```
[root@edge-14 assisted-installer-manifests-bis]# oc1 get all
NAME                                      READY   STATUS      RESTARTS   AGE
pod/assisted-installer-controller-jjszz   0/1     Completed   2          127m

NAME                                      COMPLETIONS   DURATION   AGE
job.batch/assisted-installer-controller   1/1           126m       127m
[root@edge-14 assisted-installer-manifests-bis]# oc1 describe ns assisted-installer
Name:         assisted-installer
Labels:       kubernetes.io/metadata.name=assisted-installer
Annotations:  openshift.io/sa.scc.mcs: s0:c8,c2
              openshift.io/sa.scc.supplemental-groups: 1000060000/10000
              openshift.io/sa.scc.uid-range: 1000060000/10000
Status:       Active

No resource quota.

No LimitRange resource.
```

In the controller logs I can see that the code unlabeling the namespace has been executed

```
[root@edge-14 assisted-installer-manifests-bis]# oc1 logs pod/assisted-installer-controller-jjszz                                                                                                                                    
[...]
time="2021-06-04T08:52:59Z" level=info msg="Checking cluster version operator availability status"
time="2021-06-04T08:52:59Z" level=info msg="Service acknowledged CVO is available for cluster c18b34c0-11a2-4b5d-b54f-aada21ce403d"                                                                                                  
time="2021-06-04T08:52:59Z" level=info msg="Removing run-level label from assisted-installer namespace"
time="2021-06-04T08:53:29Z" level=info msg="Start adding ingress ca to cluster" request_id=1c9fb496-be5b-4bff-9e64-3ab06d131142                                                                                                      
[...]
```

Hub cluster reports installation as succeeded

```
[root@edge-14 dev-scripts]# oc describe -n assisted-installer agentclusterinstall bmac-test
[...]
    Last Probe Time:             2021-06-04T08:54:08Z
    Last Transition Time:        2021-06-04T08:54:08Z
    Message:                     The installation has completed: Cluster is installed
    Reason:                      InstallationCompleted
    Status:                      True
    Type:                        Completed
[...]
```

# Assignees

/assign 

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] Reviewers have been listed
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change includes unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- [ ] Are the title and description (in both PR and commit) meaningful and clear?
- [ ] Is there a bug required (and linked) for this change?
- [ ] Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -386,6 +386,19 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 		}
 	}
 
+	// Unlabel run-level from assisted-installer namespace after the installation.
+	// Keeping the `run-level` label represents a security risk as it overwrites the SecurityContext configurations
+	// used for applications deployed in this namespace.
+	data := []byte(`{"metadata":{"labels":{"$patch": "delete", "openshift.io/run-level":"0"}}}`)
+	c.log.Infof("Removing run-level label from %s namespace", c.ControllerConfig.Namespace)
+	err = c.kc.PatchNamespace(c.ControllerConfig.Namespace, data)
+	if err != nil {
+		// It is a conscious decision not to fail an installation if for any reason patching the namespace
+		// in order to remove the `run-level` label has failed. This will be redesigned in the next release
+		// so that the `run-level` label is not created in the first place.
+		c.log.Warn("Failed to unlabel AI namespace after the installation.")
+	}
+
 	err = utils.WaitForPredicateWithContext(ctx, WaitTimeout, GeneralWaitInterval, c.addRouterCAToClusterCA)
 	if err != nil {
 		return errors.Errorf("Timeout while waiting router ca data")
--- a/src/assisted_installer_controller/assisted_installer_controller_test.go
+++ b/src/assisted_installer_controller/assisted_installer_controller_test.go
@@ -48,6 +48,8 @@ var (
 		MustGatherImage:       "quay.io/test-must-gather:latest",
 	}
 
+	aiNamespaceRunlevelPatch = []byte(`{"metadata":{"labels":{"$patch": "delete", "openshift.io/run-level":"0"}}}`)
+
 	progressClusterVersionCondition = &configv1.ClusterVersion{
 		Status: configv1.ClusterVersionStatus{
 			Conditions: []configv1.ClusterOperatorStatusCondition{{Type: configv1.OperatorProgressing,
@@ -529,6 +531,9 @@ var _ = Describe("installer HostRoleMaster role", func() {
 					fmt.Errorf("bad-conditions-status")).Times(1)
 				setConsoleAsAvailable("cluster-id")
 
+				// Patching NS
+				mockk8sclient.EXPECT().PatchNamespace(defaultTestControllerConf.Namespace, aiNamespaceRunlevelPatch).Return(nil)
+
 				// CVO
 				mockk8sclient.EXPECT().GetClusterVersion("version").Return(nil, fmt.Errorf("dummy")).Times(1)
 
@@ -591,6 +596,9 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				mockbmclient.EXPECT().CompleteInstallation(gomock.Any(), "cluster-id", true, "").Return(fmt.Errorf("dummy")).Times(1)
 				mockbmclient.EXPECT().CompleteInstallation(gomock.Any(), "cluster-id", true, "").Return(nil).Times(1)
 
+				// Patching NS
+				mockk8sclient.EXPECT().PatchNamespace(defaultTestControllerConf.Namespace, aiNamespaceRunlevelPatch).Return(nil)
+
 				wg.Add(1)
 				assistedController.PostInstallConfigs(context.TODO(), &wg, status)
 				wg.Wait()
@@ -602,6 +610,9 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				mockk8sclient.EXPECT().GetConfigMap(gomock.Any(), gomock.Any()).Return(nil, fmt.Errorf("aaa")).MinTimes(1)
 				mockbmclient.EXPECT().CompleteInstallation(gomock.Any(), "cluster-id", false,
 					"Timeout while waiting router ca data").Return(nil).Times(1)
+
+				// Patching NS
+				mockk8sclient.EXPECT().PatchNamespace(defaultTestControllerConf.Namespace, aiNamespaceRunlevelPatch).Return(nil)
 
 				wg.Add(1)
 				go assistedController.PostInstallConfigs(context.TODO(), &wg, status)
@@ -637,6 +648,9 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				mockbmclient.EXPECT().CompleteInstallation(gomock.Any(), "cluster-id", true, "").Return(fmt.Errorf("dummy")).Times(1)
 				mockbmclient.EXPECT().CompleteInstallation(gomock.Any(), "cluster-id", true, "").Return(nil).Times(1)
 
+				// Patching NS
+				mockk8sclient.EXPECT().PatchNamespace(defaultTestControllerConf.Namespace, aiNamespaceRunlevelPatch).Return(nil)
+
 				wg.Add(1)
 				assistedController.PostInstallConfigs(context.TODO(), &wg, status)
 				wg.Wait()
@@ -656,6 +670,9 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				mockbmclient.EXPECT().UpdateClusterOperator(gomock.Any(), "cluster-id", "lso", models.OperatorStatusProgressing, gomock.Any()).Return(nil).AnyTimes()
 				mockbmclient.EXPECT().UpdateClusterOperator(gomock.Any(), "cluster-id", "lso", models.OperatorStatusFailed, "Waiting for operator timed out").Return(nil).Times(1)
 				mockbmclient.EXPECT().CompleteInstallation(gomock.Any(), "cluster-id", true, "").Return(nil).Times(1)
+
+				// Patching NS
+				mockk8sclient.EXPECT().PatchNamespace(defaultTestControllerConf.Namespace, aiNamespaceRunlevelPatch).Return(nil)
 
 				wg.Add(1)
 				assistedController.PostInstallConfigs(context.TODO(), &wg, status)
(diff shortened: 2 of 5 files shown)


mkowalski:
/rename Bug 1951812: Remove label from AI namespace after cluster installation

mkowalski:
/retitle Bug 1951812: Remove label from AI namespace after cluster installation

mkowalski:
Thank you, dear bot. Please remind me, once you open a cherry-pick PR, that I need to rename it to point to https://bugzilla.redhat.com/show_bug.cgi?id=1967945

romfreiman:
/hold
I think we agreed that we not gonna touch it

mkowalski:
> I think we agreed that we not gonna touch it

My understanding was that we agreed not to remove the label *completely*, but that we will remove the label *after* the controller has started. What do I miss?

romfreiman:
why? how does it help?

mkowalski:
What the bugzilla reported was that **after** cluster is installed, they try to deploy assisted service in the namespace that's precreated. Given that the namespace is labeled with `run-level`, it makes the service to fail.

What we do in this PR is to remove the `run-level` during the cluster installation (after the controller has started but before the cluster is marked as installed) so that if someone tries to use this precreated namespace afterwards, they are not bypassing any security constrains.

This model allows us to keep the stuff in the namespace (so it's possible to debug it or whatever) and at the same time it's possible to deploy workflow in this namespace.

romfreiman:
but there is no such flow for production users. for now at least. they
cannot install hub cluster with assisted since it's disconnected env.
basically, this flow affects only the internal users, and hence I would
like to avoid more rist at this point.

mkowalski:
I see... In such a case we should not introduce any cheap mitigations whatsoever and only document that the namespace must not be reused unless the label is manually removed.

For the next release we should really work towards not using this label in the first place (there's already BZ and PR for that)

mkowalski:
/hold

Ok, so after the discussion on Slack - SaaS follows master, therefore we should focus on fixing it via https://bugzilla.redhat.com/show_bug.cgi?id=1966621. For anyone hitting the issue here (i.e. because they reuse the namespace) we should craft a piece of documentation/recommendation to either manually remove this label if desired or not to use the same namespace at all.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -386,6 +386,15 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 		}
 	}
 
+	// Unlabel run-level from assisted-installer namespace after the installation
+	data := []byte(`{"metadata":{"labels":{"$patch": "delete", "openshift.io/run-level":"0"}}}`)
+	c.log.Infof("Removing run-level label from %s namespace", c.ControllerConfig.Namespace)
+	err = c.kc.PatchNamespace(c.ControllerConfig.Namespace, data)
+	if err != nil {
+		c.log.WithError(err).Warn("Failed to unlabel AI namespace after the installation.")
+		return err
```
flaper87, line 395: What do you think about adding a docstring with something like:

```
Keeping the `run-level` label represents a security risk, and it overwrites the SecurityContext configurations used for applications deployed in this namespace. Because of this, failing to remove the `run-level` label is considered a deployment failure, hence the `return err`
```

Feel free to rephrase it :)

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -529,6 +531,9 @@ var _ = Describe("installer HostRoleMaster role", func() {
 					fmt.Errorf("bad-conditions-status")).Times(1)
 				setConsoleAsAvailable("cluster-id")
 
+				// Patching NS
+				mockk8sclient.EXPECT().PatchNamespace(defaultTestControllerConf.Namespace, aiNamespaceRunlevelPatch).Return(nil)
```
flaper87, line 535: Counter proposal: What do you think about explicitly creating the namespace in the e`BeforeEach` phase? This will (should?) guarantee that *every* test in the installer uses the namespace with that label set. Removal will only happen at the end of the flow, so I think that should be the exception in the test suite.


All that said, these are just unit-tests so I don't think setting the label will have any actual functional effect but it does get us closer to reality, at least in form.

src/k8s_client/k8s_client.go
```diff
@@ -182,6 +183,11 @@ func (c *k8sClient) DeletePods(namespace string) error {
 	return c.client.CoreV1().Pods(namespace).DeleteCollection(context.TODO(), metav1.DeleteOptions{}, metav1.ListOptions{})
 }
 
+func (c *k8sClient) PatchNamespace(namespace string, data []byte) error {
+	_, err := c.client.CoreV1().Namespaces().Patch(context.TODO(), namespace, types.StrategicMergePatchType, data, metav1.PatchOptions{})
```
flaper87, line 187: TODO for our future selves: We should be passing the context to these functions.

flaper87 (changes requested):
Couple of comments

flaper87:
After further investigation and discussion, it's clear that the use and, perhaps more importantly, keeping this label is a security issue for the overall cluster. It will allow for POD's running under this namespace to skip any SCC that has been configured. This https://github.com/openshift/assisted-installer/pull/291#issuecomment-855891951 goes into a bit more details (more details have been provided in other platforms) on why we should try to stop using this label.

Changing the workflow at this point (as proposed in #291) is a bit of a risk, considering the release is about to be cut. Instead, I think we could go ahead with what's been proposed here so that the label is removed *only* after the deployment has been completed and succeeded. Furthermore, I would say that failures to removing the label should be ignored (for now) to make sure this won't impact the current flow.

We will work on a better solution for future versions, until then, this should improve the situation and prevent us from shipping something that could be exploited later on.

I know @romfreiman has a different opinion here, tho. Happy to discuss it further. I'd like to make the user experience/security better without compromising the stability of the installation flow.

flaper87:
/retitle OCPBUGSM-30037: Remove label from AI namespace after cluster installation

flaper87:
/retitle Bug 1951812: Remove label from AI namespace after cluster installation

(sorry, was testing something :smiley:  )

src/assisted_installer_controller/assisted_installer_controller.go
flaper87: Let's maybe not handle the error here. If the patch command for whatever reason fail, we should not stop the workflow on this specific case... at least until we have a proper solution.

flaper87 (changes requested):

src/assisted_installer_controller/assisted_installer_controller.go
mkowalski: Changed, the error is not handled so that the logic we are adding has no chance of failing the cluster deployment

src/assisted_installer_controller/assisted_installer_controller_test.go
mkowalski: Given that I have already a PR for the next release that removes the label completely, I'm not sure if it's worth redesigning the tests in here

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -386,6 +386,19 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
 	}
 
+	// Unlabel run-level from assisted-installer namespace after the installation.
+	// Keeping the `run-level` label represents a security risk as it overwrites the SecurityContext configurations
+	// used for applications deployed in this namespace.
+	data := []byte(`{"metadata":{"labels":{"$patch": "delete", "openshift.io/run-level":"0"}}}`)
+	c.log.Infof("Removing run-level label from %s namespace", c.ControllerConfig.Namespace)
+	err = c.kc.PatchNamespace(c.ControllerConfig.Namespace, data)
+	if err != nil {
+		// It is a conscious decision no to fail an installation if for any reason patching the namespace
```
flaper87, line 396:
```suggestion
		// It is a conscious decision not to fail an installation if for any reason patching the namespace
```

flaper87 (approved):
@romfreiman 

In a conversation with folks from ACM, the idea of using an ACM policy seems like it's going to be a bit more complex than anticipated. There are 2 main things:

1. To have the node managed by ACM, it has to be imported into ACM
2. Using ACM makes more sense when the lifecycle of the nodes needs to be managed by ACM itself. ACM's strength is in scaling fleets of nodes.
3. We only need to run this once. Using an ACM policy means that ACM will guarantee that the state of the cluster is what is in the policies.

All the above makes the ACM Policy idea less appealing. I'm going to go ahead and move this forward. I'm confident the way it's been crafted is safe enough for the deployment and won't impact it negatively. The patch command is best-effort, meaning errors will be ignored and instead a warning will be logged.

src/assisted_installer_controller/assisted_installer_controller.go
mkowalski: Fixed

PR #306, opened 2021-06-15
OCPBUGSM-30942: ACM/ZTP with Wan emulation, SNO cluster installs do not show as installed although they are

tsorya:
In case it takes long time fopr dns operator to start we can get timeout
in assisted-service cause assisted-controller can't reach the service.
Another issue is that we are workarounding dns pod issue in
assisted-controller and that's why this part should run before
controller tries to reach the service.
From now controller will use host network and will not be dependent on
dns pod.


eranco74 (commented):
/lgtm

PR #313, opened 2021-06-17
Bug 1973314: Update the uefi boot entry use shimx64

derekhiggins:
Refering to shim as BOOTX64.EFI triggers a bug
we'd like to avoid.

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -229,7 +229,7 @@ func (o *ops) SetBootOrder(device string) error {
 	o.log.Info("Setting efibootmgr to boot from disk")
 
 	// efi-system is installed onto partition 2
-	_, err = o.ExecPrivilegeCommand(o.logWriter, "efibootmgr", "-d", device, "-p", "2", "-c", "-L", "RHCOS", "-l", "\\EFI\\BOOT\\BOOTX64.EFI")
+	_, err = o.ExecPrivilegeCommand(o.logWriter, "efibootmgr", "-d", device, "-p", "2", "-c", "-L", "Red Hat Enterprise Linux", "-l", "\\EFI\\redhat\\shimx64.efi")
 	if err != nil {
 		o.log.Errorf("Failed to set efibootmgr to boot from disk %s, err: %s", device, err)
 		return err


derekhiggins:
/hold
need to verify

src/ops/ops.go
```diff
@@ -229,7 +229,7 @@ func (o *ops) SetBootOrder(device string) error {
 	o.log.Info("Setting efibootmgr to boot from disk")
 
 	// efi-system is installed onto partition 2
-	_, err = o.ExecPrivilegeCommand(o.logWriter, "efibootmgr", "-d", device, "-p", "2", "-c", "-L", "RHCOS", "-l", "\\EFI\\BOOT\\BOOTX64.EFI")
+	_, err = o.ExecPrivilegeCommand(o.logWriter, "efibootmgr", "-d", device, "-p", "2", "-c", "-L", "Red Hat Enterprise Linux", "-l", "\\EFI\\redhat\\shimx64.efi")
```
omertuc, line 232: Does the label change make any difference? It's technically RHCOS and not RHEL so I'm not sure why it was added to the PR

derekhiggins: Yes it does, shim looks at the file /boot/EFI/redhat/BOOTX64.CSV , to find the label it expects and the path it expects

dtantsur: Reference: https://github.com/rhboot/shim/blob/a1170bb00a116783cc6623b403e785d86b2f97d7/README.fallback#L33-L46 (I cannot understand from it if the label is required tbh).

derekhiggins:
> Other than the label change that I believe is unnecessary (and wrong since it's RHCOS and not RHEL) this looks okay to me.
If the label doesn't match whats in the file BOOTX64.csv, then shim will create a new entry, 

> 
> However - this needs to be tested - we have to make sure ZTP still works with this change.
> 
> Also, if possible, please use a more verbose/correct commit message:

Will do, I placed a /hold on this bug because I'm not familiar with assisted installer, I was just pushing up what I thought 
would fix the problem so that somebody could test it, I then intended to fix up the patch.

> 
> > Refering to shim as BOOTX64.EFI triggers a bug
> > we'd like to avoid.
> 
> What kind of bug? I presume there were secure boot issues that this solves?

The bug is in the PR title ( Bug 1973314)

> 
> > Referring to shim as BOOTX64.EFI
> 
> `\\EFI\\redhat\\shimx64.efi` is the shim, the PR changes the new boot entry to point at the shim rather than at `BOOT64X.EFI`, the current wording is a bit confusing

shimx64.efi and BOOT64X.EFI are copies of the same file, but they follow a different code path based on the filename

omertuc:
> The bug is in the PR title ( Bug 1973314)

Yeah I saw that, but it's just a dump of a very long conversation - I would expect some short summary / overview / conclusions in the commit message that relate more specifically to this commit and what it fixes exactly

omertuc:
In any case, thank you for the PR. We'll have someone test this to make sure it doesn't break ZTP. 

For now I'll hold
/hold

src/ops/ops.go
omertuc: Thank you for the link.

> if it doesn't see the entry it expects it creates a new one

I went over the bugzilla / shim code / README.fallback and I'm not sure that's what's happening. The way I understood it, if the shim (which is present as both `BOOT/BOOTX64.efi` and `redhat/shimx64.efi`) gets booted as `BOOT/BOOTx64.efi` (rather than the correct `redhat/shimx64.efi`, and it sees that the fallback `BOOT/fbx64.efi` bootloader exists - it assumes that it happened because the uefi nvram boot entry got corrupt for some reason and as a result the firmware is trying to treat the disk in the default manner of booting `BOOT/BOOTx64.efi`.

In response to that, the shim runs the fallback bootloader `BOOT/fbx64.efi` whose purpose is to re-create the boot entry using the information present in `redhat/BOOTX64.CSV`.

Doesn't seem like it compares the label at any point to trigger this behavior, so it leads me to believe that the choice of label created by AI is of no consequence.

However, I guess if it falls-back to `Red Hat Enterprise Linux` on corruption, we might as well use the same name from the get go, even though it's a bit misleading since it's RHCOS and not RHEL

ronniel1: @flaper87 , @celebdor please take a look

flaper87:
> In any case, thank you for the PR. We'll have someone test this to make sure it doesn't break ZTP.

Generating a build for this that can tested

celebdor:
This PR should refer to https://bugzilla.redhat.com/show_bug.cgi?id=1966129 which is the 4.9.0. The cherry-pick should be the one referring to https://bugzilla.redhat.com/show_bug.cgi?id=1973314

flaper87:
@mcornea helped testing this PR. Here's a comment from him:

> @flaper87 I think we're good to go. I tried your images on both the BM and on a VMs environment to make sure it didn't regress

/retest

flaper87:
/retitle Bug 1966129: Update the uefi boot entry use shimx64

flaper87:
@celebdor we will need a clone of  bz#1966129 for 4.9.0 and on the `assisted-installer` component. bz#1966129 is already VERIFIED and it was created for Ironic.

flaper87:
/retitle Bug 1973314: Update the uefi boot entry use shimx64

PR #317, opened 2021-06-28
Delete namespace

tsorya:


eranco74:
@tsorya please update or close

PR #319, opened 2021-06-30
OCPBUGSM-30066: Retry wait for operator in case it failed

machacekondra:
This commit add a retry mechanism while waiting for the operator to be
ready. If we apply the CR of the operator it may happen (bug 1968606),
that the OLM will report the Failed state, even that it's actually
progressing. So we decided to ignore the failed state for few times.

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -34,6 +34,9 @@ import (
 )
 
 const (
+	// We retry 10 times in 30sec interval meaning that we tolerate the operator to be in failed
+	// state for 5minutes.
+	failedOperatorRetry       = 10
 	generalWaitTimeoutInt     = 30
 	controllerLogsSecondsAgo  = 120 * 60
 	consoleOperatorName       = "console"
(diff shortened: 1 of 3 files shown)


pkliczewski:
/assign @eranco74 
please review

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -34,6 +34,7 @@ import (
 )
 
 const (
+	failedOperatorRetry       = 10
```
eranco74, line 37: Consider adding a comment that 10 retries provide 5 minutes buffer for the operator to be in Failed status
The waitForOLMOperators is called with 30 seconds interval:
https://sourcegraph.com/github.com/openshift/assisted-installer@64bb7de193acba8112c0d632befb500863580538/-/blob/src/assisted_installer_controller/assisted_installer_controller.go?L436

eranco74 (commented):
/lgtm

Please add a UT

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -785,6 +790,14 @@ func (c controller) waitForOLMOperators() bool {
 		}
 
 		operatorStatus := utils.CsvStatusToOperatorStatus(string(csv.Status.Phase))
+
+		// FIXME: We retry the check of the opreator status in case it's in failed state to WA bug 1968606
```
masayag, line 794: s/opreator/operator

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -785,6 +790,14 @@ func (c controller) waitForOLMOperators() bool {
...
+		if utils.IsStatusFailed(operatorStatus) && c.retryMap[operator.Name] < failedOperatorRetry {
+			c.retryMap[operator.Name]++
+			c.log.Warnf("Operator %s has failed state retry(%v) the check.", operator.Name, c.retryMap[operator.Name])
```
masayag, line 797: perhaps `...retry(%d/%d)` which will print `...retry(5/10)` ?

pkliczewski:
@eranco74 please review again

eranco74 (commented):
/lgtm

PR #325, opened 2021-07-12
MGMT-7210 Upgrade Go version to 1.16

omertuc:
older versions of go are out of support, so for security compliance, we were trying to get all components on the latest version. 1.14 is already out of support, i.e. https://endoflife.date/go


YuviGold:
/hold
Till OpenShift CI would have a PR pending with matched changes.

omertuc:
https://github.com/openshift/release/pull/20096

PR #327, opened 2021-07-13
Bug 1981465: Assisted installer wait for ready master nodes on bootst…

eranco74:
…rap kube-apiserver though the kube-apiserver moved to one of the masters

On the bootstrap node the assisted-installer use the loopback-kubeconfig
to query the kube-apiserver for the number of ready master nodes.

Usually both master nodes join the cluster and become ready before
bootkube takes down the bootstrap control plane so the loopback kubeconfig works.
But in case clusteer bootstrap finish before the 2 master nodes
are ready the assisted-installer will wait forever since the it's
using the loopback-kubeconfig and the bootstrap control plane is down
resulting in "connection refused"

The assisted-installer should query the kube-apiserver running on
one of the master nodes, for that to work it should use the real
kubeconfig instead of the loopback kubeconfig.

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -26,7 +26,6 @@ import (
 
 const (
 	InstallDir                   = "/opt/install-dir"
-	KubeconfigPathLoopBack       = "/opt/openshift/auth/kubeconfig-loopback"
 	KubeconfigPath               = "/opt/openshift/auth/kubeconfig"
 	minMasterNodes               = 2
 	dockerConfigFile             = "/root/.docker/config.json"
@@ -337,12 +336,18 @@ func (i *installer) waitForNetworkType(kc k8s_client.K8SClient) error {
 }
 
 func (i *installer) waitForControlPlane(ctx context.Context) error {
-	kc, err := i.kcBuilder(KubeconfigPathLoopBack, i.log)
+	err := i.ops.ReloadHostFile("/etc/resolv.conf")
+	if err != nil {
+		i.log.WithError(err).Error("Failed to reload resolv.conf")
+		return err
+	}
+	kc, err := i.kcBuilder(KubeconfigPath, i.log)
 	if err != nil {
 		i.log.Error(err)
 		return err
 	}
 	i.UpdateHostInstallProgress(models.HostStageWaitingForControlPlane, "")
+
 	if err = i.waitForMinMasterNodes(ctx, kc); err != nil {
 		return err
 	}
@@ -364,7 +369,7 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 	i.waitForBootkube(ctx)
 
 	// waiting for controller pod to be running
-	if err := i.waitForController(); err != nil {
+	if err := i.waitForController(kc); err != nil {
 		i.log.Error(err)
 		return err
 	}
@@ -469,20 +474,9 @@ func (i *installer) waitForBootkube(ctx context.Context) {
 	}
 }
 
-func (i *installer) waitForController() error {
+func (i *installer) waitForController(kc k8s_client.K8SClient) error {
 	i.log.Infof("Waiting for controller to be ready")
 	i.UpdateHostInstallProgress(models.HostStageWaitingForController, "waiting for controller pod ready event")
-	err := i.ops.ReloadHostFile("/etc/resolv.conf")
-	if err != nil {
-		i.log.WithError(err).Error("Failed to reload resolv.conf")
-		return err
-	}
-
-	kc, err := i.kcBuilder(KubeconfigPath, i.log)
-	if err != nil {
-		i.log.WithError(err).Errorf("Failed to create kc client from %s", KubeconfigPath)
-		return err
-	}
 
 	events := map[string]string{}
 	tickerUploadLogs := time.NewTicker(5 * time.Minute)
--- a/src/installer/installer_test.go
+++ b/src/installer/installer_test.go
@@ -404,7 +404,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 			Expect(ret).Should(Equal(err))
 		})
 	})
-	Context("Bootstrap role waiting for controller", func() {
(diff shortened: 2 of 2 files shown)


eranco74:
This doesn't seem to work, the bootstrap node can't seem to resolve the cluster DNS:

```
time="2021-07-13T12:24:22Z" level=warning msg="Still waiting for master nodes: Get \"https://api-int.test-infra-cluster-assisted-installer.redhat.com:6443/api/v1/nodes?labelSelector=node-role.kubernetes.io%2Fmaster\": dial tcp: lookup api-int.test-infra-cluster-assisted-installer.redhat.com on 192.168.126.1:53: no such host"
time="2021-07-13T12:24:27Z" level=warning msg="Still waiting for master nodes: Get \"https://api-int.test-infra-cluster-assisted-installer.redhat.com:6443/api/v1/nodes?labelSelector=node-role.kubernetes.io%2Fmaster\": dial tcp: lookup api-int.test-infra-cluster-assisted-installer.redhat.com on 192.168.126.1:53: no such host"
time="2021-07-13T12:24:32Z" level=warning msg="Still waiting for master nodes: Get \"https://api-int.test-infra-cluster-assisted-installer.redhat.com:6443/api/v1/nodes?labelSelector=node-role.kubernetes.io%2Fmaster\": dial tcp: lookup api-int.test-infra-cluster-assisted-installer.redhat.com on 192.168.126.1:53: no such host"
time="2021-07-13T12:24:37Z" level=warning msg="Still waiting for master nodes: Get \"https://api-int.test-infra-cluster-assisted-installer.redhat.com:6443/api/v1/nodes?labelSelector=node-role.kubernetes.io%2Fmaster\": dial tcp: lookup api-int.test-infra-cluster-assisted-installer.redhat.com on 192.168.126.1:53: no such host"
time="2021-07-13T12:24:42Z" level=warning msg="Still waiting for master nodes: Get \"https://api-int.test-infra-cluster-assisted-installer.redhat.com:6443/api/v1/nodes?labelSelector=node-role.kubernetes.io%2Fmaster\": dial tcp: lookup api-int.test-infra-cluster-assisted-installer.redhat.com on 192.168.126.1:53: no such host"
```

eranco74:
There's coredns running on the node that should allow the assisted-installer (running with net=host) to resolve the api-int DNS.
Moved the reload of /etc/resolv.conf prior of the kubeconfig usage.

eranco74:
`Packet setup failed.`
/retest

src/installer/installer.go
```diff
@@ -337,12 +337,18 @@ func (i *installer) waitForNetworkType(kc k8s_client.K8SClient) error {
 }
 
 func (i *installer) waitForControlPlane(ctx context.Context) error {
-	kc, err := i.kcBuilder(KubeconfigPathLoopBack, i.log)
```
YuviGold, line 340: Delete the unused `KubeconfigPathLoopBack` constant

src/installer/installer.go
```diff
@@ -472,11 +478,6 @@ func (i *installer) waitForBootkube(ctx context.Context) {
 func (i *installer) waitForController() error {
 	i.log.Infof("Waiting for controller to be ready")
 	i.UpdateHostInstallProgress(models.HostStageWaitingForController, "waiting for controller pod ready event")
-	err := i.ops.ReloadHostFile("/etc/resolv.conf")
-	if err != nil {
-		i.log.WithError(err).Error("Failed to reload resolv.conf")
-		return err
-	}
 
 	kc, err := i.kcBuilder(KubeconfigPath, i.log)
```
YuviGold, line 482: No reason to create k8s_client also in here. You can pass the k8s client from the calling `waitForControlPlane` function.

src/installer/installer.go
```diff
@@ -337,12 +336,18 @@ func (i *installer) waitForNetworkType(kc k8s_client.K8SClient) error {
 }
 
 func (i *installer) waitForControlPlane(ctx context.Context) error {
-	kc, err := i.kcBuilder(KubeconfigPathLoopBack, i.log)
+	err := i.ops.ReloadHostFile("/etc/resolv.conf")
+	if err != nil {
+		i.log.WithError(err).Error("Failed to reload resolv.conf")
+		return err
+	}
+	kc, err := i.kcBuilder(KubeconfigPath, i.log)
```
romfreiman, line 344: so u basically pass this stage when there will be at least resolvable 1 api server ? Meaning at least 1 master.
But can it happen that it will move to another node later on, for exmaple in waitForMinMasterNodes - do u deal with the disruption in the kcBuilder? Or the assumption is the api is stable

src/installer/installer.go
```diff
@@ -364,7 +369,7 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 	i.waitForBootkube(ctx)
```
romfreiman, line 369: this will always return immidetely, no?

src/installer/installer.go
eranco74: Not exactly, this means we don't care if the kube-apiserver is running on the bootstrap or one of the master nodes.
The disadvantage (compared to loopback kubeconfig) is that now we depend on coredns and keepalived in order to resolve the cluster DNS to the node running the kube-apiserver

src/installer/installer.go
eranco74: This will return if bootkube service compelted.
In 4.6 it took some time (due to the etcd patch) but since 4.7 I think it returns quickly since bootkube finish before the 2 master nodes are ready.

src/installer/installer.go
eranco74: I guess this is the tradeoff.
What is more prone to failure:
1. Current implementation that depends on master nodes being ready before kube-apiserver move from the bootstrap.
2. This PR implementation that relay on keepalived and coredns for accessing the kube-apiserver.

YuviGold:
/lgtm

Waiting for @tsorya about it
/hold

eranco74:
logs look good.
/unhold

PR #328, opened 2021-07-14
Bug 1968164,1968181: Modify the controller to check and wait for Stor…

rewantsoni:
…age Cluster of OCS operator to be ready.

Revert the changes made to implement retry mechanism.

CSV status of OCS is updated as:
1. It succeeds when OCS is deployed
2. It fails when the StorageCluster is being applied
3. It succeeds when the StorageCluster is applied

So we check the status of the StorageCluster CR, till it's fixed in OCS.

Signed-off-By: Rewant Soni <resoni@redhat.com>


go.mod
```diff
@@ -45,9 +45,29 @@ replace (
...
+	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.21.0
+	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.21.0
+	k8s.io/kube-proxy => k8s.io/kube-proxy v0.21.0
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.21.0
+	k8s.io/kubectl => k8s.io/kubectl v0.21.0
+	k8s.io/kubelet => k8s.io/kubelet v0.21.0
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.21.0
+	k8s.io/metrics => k8s.io/metrics v0.21.0
+	k8s.io/mount-utils => k8s.io/mount-utils v0.21.0
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.21.0
```
YuviGold, line 70: That's a lot of replaces... are those mandatory?
cc @eranco74

src/utils/utils.go
```diff
@@ -41,6 +41,18 @@ const (
...
+// https://github.com/openshift/ocs-operator/blob/master/controllers/util/status.go#L14-L32
+// These constants represent the Phase as used by SC.Status.Phase for OCS
+var (
+	// SCPhaseError is used when SetErrorCondition is called
+	SCPhaseError = "Error"
+	// SCPhaseReady is used when SetCompleteCondition is called
+	SCPhaseReady = "Ready"
+	// SCPhaseProgressing is used when SetProgressingCondition is called
+	SCPhaseProgressing = "Progressing"
+)
```
machacekondra, line 54: Can't we reuse it from the `ocs-operator` code?

src/k8s_client/k8s_client.go
```diff
@@ -423,6 +425,20 @@ func (c *k8sClient) SetProxyEnvVars() error {
...
+func (c *k8sClient) GetSC() (*ocsv1.StorageCluster, error) {
+	//hard coding this as it's been hardcoded in assisted-service
+	sc := &ocsv1.StorageCluster{}
+	err := c.olmClient.RESTClient().Get().
+		Resource("storageclusters").
+		Namespace("openshift-storage").
+		Name("ocs-storagecluster").
+		Do(context.TODO()).
+		Into(sc)
+	if err!=nil{
```
machacekondra, line 437: Please format.

src/k8s_client/k8s_client.go
```diff
@@ -423,6 +425,20 @@ func (c *k8sClient) SetProxyEnvVars() error {
 	return nil
 }
 
+func (c *k8sClient) GetSC() (*ocsv1.StorageCluster, error) {
```
machacekondra, line 428: I would name it `GetStorageCluster`

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -773,31 +768,45 @@ func (c controller) waitForOLMOperators() bool {
...
-		csv, err := c.kc.GetCSV(operator.Namespace, csvName)
-		if err != nil {
-			c.log.WithError(err).Warnf("Failed to get %s", operator.Name)
-			continue
-		}
+		//FIXME: We check the SC of the OCS operator WA bug 1968606
+		//Remove this code when bug 1968606 is fixed
+		if operator.Name == "ocs" {
+			//get SC
+			sc, err := c.kc.GetSC()
```
machacekondra, line 776: I just wonder whether maybe we don't want to check both SC and CSV for OCS? Or SC is enough?

go.mod
rewantsoni: Yes, they are required by OCS

src/utils/utils.go
rewantsoni: Yes, done

src/k8s_client/k8s_client.go
rewantsoni: Done

src/assisted_installer_controller/assisted_installer_controller.go
rewantsoni: SC should be enough

machacekondra: Well SC could be also `StorageClass` for example, which may be confusing.

go.mod
YuviGold: We would like to avoid that if possible...
The last time we did it for hive was troublesome
cc @filanov

ronniel1:
@pkliczewski Please take a look at this.
We put in a lot of effort to make sure that our code is generic and put all the operator-specific logic into the "plugins"

eranco74:
@machacekondra @rewantsoni wasn't this handled by https://github.com/openshift/assisted-installer/pull/319?
OCPBUGSM-30066 points to https://bugzilla.redhat.com/show_bug.cgi?id=1968164
The code already waits for the operator to recover. 
I don't think we want to keep adding patches around a bug in the operator.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -773,31 +768,45 @@ func (c controller) waitForOLMOperators() bool {
...
-		}
 
-		csv, err := c.kc.GetCSV(operator.Namespace, csvName)
-		if err != nil {
-			c.log.WithError(err).Warnf("Failed to get %s", operator.Name)
-			continue
-		}
+		//FIXME: We check the SC of the OCS operator WA bug 1968606
+		//Remove this code when bug 1968606 is fixed
+		if operator.Name == "ocs" {
```
pkliczewski, line 774: We should not introduce operator specific logic in the installer even though it is temporal fix pending fix in OCS. I think we need to make the decision on the service side and have generic logic here. @rewantsoni please fix it.

pkliczewski:
> I don't think we want to keep adding patches around a bug in the operator.

We need to workaround OCS issue because we are not reporting correct installation status. Once it is fixed in OCS (probably in next release) we can remove it. I agree this is not the place to solve it.

go.mod
filanov: You think that we can suggest OCS to work with sub moduls?

src/assisted_installer_controller/assisted_installer_controller.go
gobindadas: Correct we don't need any operator specific code. Could be 
sc, err := c.kc.GetStorageCluster()  is enough
and check if sc is not null then proceed, fo other operator sc will be always null,

rewantsoni: @pkliczewski @gobindadas I have updated it can you review it again?

go.mod
rewantsoni: Because of https://github.com/kubernetes/kubernetes/issues/79384, we have to update the version

src/assisted_installer_controller/assisted_installer_controller.go
rewantsoni: @ronniel1 Could you also review it?

go.mod
filanov: Because you are only using utils and api there i don't think that you need all the replaces.
We had similar issue with Hive when we had to replace everything that they replaced and then every component that import from assisted-service did the same, it was very hard to manage. 
So what we did was to add sub modules in Hive, we used only the API so why do we need to import anything else? 
Maybe we need to suggest the same to ocs? 
Here is an example https://github.com/openshift/hive/tree/master/apis

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -773,31 +768,39 @@ func (c controller) waitForOLMOperators() bool {
...
-		}
 
-		csv, err := c.kc.GetCSV(operator.Namespace, csvName)
-		if err != nil {
-			c.log.WithError(err).Warnf("Failed to get %s", operator.Name)
-			continue
-		}
+		//FIXME: We check the SC of the OCS operator WA bug 1968606
+		//Remove this code when bug 1968606 is fixed
+		sc, err := c.kc.GetStorageCluster()
```
machacekondra, line 774: What if current `operator` is CNV, you chech StorageCLuster which is available, and update the CNV operator by the StorageCluster status. I think the previous option was correct.

rewantsoni: I have added the operator.namespace, so if its a cnv or any other operator, StorageCluster will return nil, It's a hack. The other option can be to wait after applying the post manifest install yaml, but we don't know how long we have to wait as it may depend on the cluster.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -773,31 +768,39 @@ func (c controller) waitForOLMOperators() bool {
...
 
-		csv, err := c.kc.GetCSV(operator.Namespace, csvName)
-		if err != nil {
-			c.log.WithError(err).Warnf("Failed to get %s", operator.Name)
-			continue
-		}
+		//FIXME: We check the StorageCluster of the OCS operator WA bug 1968606
+		//Remove this code when bug 1968606 is fixed
+		sc, err := c.kc.GetStorageCluster(operator.Namespace)
+		if sc != nil {
```
pkliczewski, line 775: It is still very specific for ocs operator. What do you think if we could define what should we watch to get the status? In this case we could make this decision in operator specific code in the service and here we could use it to track CSV or any other specific CR.

rewantsoni: Yes, I think that could work

PR #329, opened 2021-07-14
MGMT-4893: Add Must-Gather reports when olm controllers fail

slaviered:
- Add support for JSON-formatted MUST_GATHER_IMAGE variable
   - Backward compatability with other formats of MUST_GATHER_IMAGE
   - When one of the Olm operator fails or timesout it is marked on the controller status
   - At the end of the installation process (either normal or aborted) we check if must-gather
     report should be collected and with what scope

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -939,7 +939,7 @@ func (c controller) parseMustGatherImages() []string {
 		return images
 	}
 
-	c.log.Infof("collecting must-gather logs using a this image configuration %s", c.MustGatherImage)
+	c.log.Infof("collecting must-gather logs using this image configuration %s", c.MustGatherImage)
 	var imageMap map[string]string
 	err := json.Unmarshal([]byte(c.MustGatherImage), &imageMap)
 	if err != nil {
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -86,14 +86,17 @@ type Controller interface {
 
 type ControllerStatus struct {
 	errCounter uint32
+	components map[string]bool
+	lock       sync.Mutex
 }
 
 type controller struct {
 	ControllerConfig
-	log *logrus.Logger
-	ops ops.Ops
-	ic  inventory_client.InventoryClient
-	kc  k8s_client.K8SClient
+	Status *ControllerStatus
+	log    *logrus.Logger
+	ops    ops.Ops
+	ic     inventory_client.InventoryClient
+	kc     k8s_client.K8SClient
 }
 
 func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inventory_client.InventoryClient, kc k8s_client.K8SClient) *controller {
@@ -103,6 +106,13 @@ func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inv
 		ops:              ops,
 		ic:               ic,
 		kc:               kc,
+		Status:           NewControllerStatus(),
+	}
+}
+
+func NewControllerStatus() *ControllerStatus {
+	return &ControllerStatus{
+		components: make(map[string]bool),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


rohantmp:
Hi @slaviered , perhaps you wanted to assign another Rohan? I haven't really worked on this component before, just explained local-storage-operator to people working on integrating assisted-installer with local-storage-operator.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -894,6 +929,34 @@ func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSec
 	return nil
 }
 
+func (c controller) parseMustGatherImages() []string {
+	images := make([]string, 0)
+	if c.MustGatherImage == "" {
+		c.log.Infof("collecting must-gather logs into using image from release\n")
```
YuviGold, line 935: You can remove the `\n` from your logs.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -894,6 +929,34 @@ func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSec
...
+		return images
+	}
+
+	var imageMap map[string]string
+	err := json.Unmarshal([]byte(c.MustGatherImage), &imageMap)
+	c.log.Infof("collecting must-gather logs into using a this image setup %s\n", c.MustGatherImage)
```
YuviGold, line 941: Something with this log syntax isn't right.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -894,6 +929,34 @@ func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSec
...
+	if err != nil {
+		images = append(images, c.MustGatherImage)
```
YuviGold, line 943: Why would it be appended if there was an error?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -103,6 +106,13 @@ func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inv
 		ops:              ops,
 		ic:               ic,
 		kc:               kc,
+		Status:           NewControllerStatus(),
+	}
+}
+
+func NewControllerStatus() *ControllerStatus {
+	return &ControllerStatus{
+		components: make(map[string]bool, 4),
```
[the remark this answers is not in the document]
slaviered: 3 olm operators + 1 ocp image

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: In this case the c.MustGatherImage is just a string

slaviered: Actually it is always a string

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: It means that the structure of the variable is the old one (non-JSON) and therefor can be taken as is

src/assisted_installer_controller/assisted_installer_controller.go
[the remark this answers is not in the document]
slaviered: fixed

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: moved up and re-worded

src/assisted_installer_controller/assisted_installer_controller.go
YuviGold: In what scenario would it not be a JSON? is that backward-compatible? If so - please add a comment.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -898,6 +932,40 @@ func (c controller) uploadSummaryLogs(podName string, namespace string, sinceSec
...
 }
 
+func (c controller) parseMustGatherImages() []string {
+	images := make([]string, 0)
+	if c.MustGatherImage == "" {
+		c.log.Infof("collecting must-gather logs into using image from release")
+		return images
+	}
+
+	c.log.Infof("collecting must-gather logs using a this image configuration %s", c.MustGatherImage)
```
YuviGold, line 942:
```suggestion
	c.log.Infof("collecting must-gather logs using this image configuration %s", c.MustGatherImage)
```

src/assisted_installer_controller/assisted_installer_controller.go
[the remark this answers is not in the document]
YuviGold: It shouldn't be hardcoded

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: Yes it is backward compatible - This is the part that takes card of the backward comparability. comment was added

YuviGold (approved):

PR #331, opened 2021-07-18
MGMT-7149: Merge wait for operators

YuviGold:
Switch to a a generic `isOperatorAvailable` function that would be used
for getting any class that inherits `OperatorHandler`.

This function gets the operator status from the service, if the status
is available - stop running. Otherwise, get the operator status
locally and check if it is different than the status at the service, if
so - send an update to the service.

There are 3 implementations for the OperatorHandler` interface:

- ClusterOperatorHandler (such as console)
- ClusterVersionHandler (only CVO)
- ClusterServiceVersionHandler (such as OCS, LSO)

/cc @machacekondra  @eranco74 @tsorya

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -23,7 +23,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/assisted-installer/src/common"
 	"github.com/openshift/assisted-installer/src/inventory_client"
 	"github.com/openshift/assisted-installer/src/k8s_client"
@@ -40,7 +39,6 @@ const (
 	generalWaitTimeoutInt     = 30
 	controllerLogsSecondsAgo  = 120 * 60
 	consoleOperatorName       = "console"
-	cvoOperatorName           = "cvo"
 	ingressConfigMapName      = "default-ingress-cert"
 	ingressConfigMapNamespace = "openshift-config-managed"
 	dnsServiceName            = "dns-default"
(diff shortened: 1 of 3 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -766,151 +754,52 @@ func (c controller) updatePendingOLMOperators() error {
...
-			c.log.Infof("CSV %s updated, status: %s -> %s, message: %s -> %s.", operator.Name, operator.Status, operatorStatus, operator.StatusInfo, csv.Status.Message)
+		currentHandlers := handlers[:]
 
-			err = c.ic.UpdateClusterOperator(context.TODO(), c.ClusterID, operator.Name, operatorStatus, csv.Status.Message)
-			if err != nil {
-				c.log.WithError(err).Warnf("Failed to update olm %s status", operator.Name)
-				continue
+		for index := range currentHandlers {
+			if c.isOperatorAvailable(currentHandlers[index]) {
+				handlers = append(handlers[:index], handlers[index+1:]...)
```
eranco74, line 781: This seems buggy.
Looks like you remove a handler from `handlers` but iterate over `currentHandlers`.
After c.isOperatorAvailable(currentHandlers[index]) is true once the rest of the loop is out of sync.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -721,14 +716,7 @@ func (c controller) addRouterCAToClusterCA() bool {
...
 
-func (c controller) getMaximumOLMTimeout() time.Duration {
-
-	operators, err := c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID)
-	if err != nil {
-		c.log.WithError(err).Warningf("Failed to connect to assisted service")
-		return WaitTimeout
-	}
-
+func (c controller) getMaximumOLMTimeout(operators []models.MonitoredOperator) time.Duration {
```
eranco74, line 719: Consider
```suggestion
func (c controller) getMaximumOLMTimeout(operators []*models.MonitoredOperator) time.Duration {
```

src/assisted_installer_controller/assisted_installer_controller.go
YuviGold: You're right. fixed.

eranco74 (commented):
/lgtm

PR #332, opened 2021-07-20
OCPBUGSM-31942: Return error on failed status operator update

machacekondra:
The method call getProgressingOLMOperators could return error in case
the assisted service API is unavailable. In that case we wouldn't update
the status of the operators which are pending and it may happend that
the cluster will hang in finalizing state.


YuviGold (approved):
LGTM

PR #333, opened 2021-07-23
MGMT-7292: Wait for the OLM to be initilized before CR apply

machacekondra:
Depends on: https://github.com/openshift/assisted-service/pull/2266

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -2,6 +2,7 @@ package assisted_installer_controller
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -44,11 +45,12 @@ const (
 	dnsServiceName            = "dns-default"
 	dnsServiceNamespace       = "openshift-dns"
 	dnsOperatorNamespace      = "openshift-dns-operator"
+	maxFetchAttempts          = 5
 	maxDeletionAttempts       = 5
 	maxDNSServiceIPAttempts   = 45
 	KeepWaiting               = false
 	ExitWaiting               = true
-	customManifestsFile       = "custom_manifests.yaml"
+	customManifestsFile       = "custom_manifests.json"
 	kubeconfigFileName        = "kubeconfig-noingress"
 )
 
@@ -61,6 +63,7 @@ var (
 	CompleteTimeout          = 30 * time.Minute
 	DNSAddressRetryInterval  = 20 * time.Second
 	DeletionRetryInterval    = 10 * time.Second
+	FetchRetryInterval       = 10 * time.Second
 	LongWaitTimeout          = 10 * time.Hour
 	CVOMaxTimeout            = 3 * time.Hour
 )
@@ -98,6 +101,14 @@ type controller struct {
 	ops    ops.Ops
 	ic     inventory_client.InventoryClient
 	kc     k8s_client.K8SClient
+}
+
+// manifest store the operator manifest used by assisted-installer to create CRs of the OLM:
+type manifest struct {
+	// name of the operator the CR manifest we want create
+	name string
+	// content of the manifest of the opreator
+	content string
 }
 
 func NewController(log *logrus.Logger, cfg ControllerConfig, ops ops.Ops, ic inventory_client.InventoryClient, kc k8s_client.K8SClient) *controller {
@@ -459,16 +470,55 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 		return errors.Wrapf(err, "Timeout while waiting for console to become available")
 	}
 
+	// Wait for OLM operators
+	if err = c.waitForOLMOperators(ctx); err != nil {
+		return errors.Wrapf(err, "Error while initializing OLM operators")
+	}
+
+	return nil
+}
+
+func (c controller) waitForOLMOperators(ctx context.Context) error {
+	var operators []models.MonitoredOperator
+	var err error
+
+	// Get the monitored operators:
+	err = utils.Retry(maxFetchAttempts, FetchRetryInterval, c.log, func() error {
+		operators, err = c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID)
+		if err != nil {
+			return errors.Wrapf(err, "Error while fetch the monitored operators from assisted-service.")
+		}
+		return nil
+	})
+	if err != nil {
+		return errors.Wrapf(err, "Failed to fetch monitored operators")
+	}
+	if len(operators) == 0 {
+		c.log.Info("No OLM operators found.")
+		return nil
+	}
+
+	// Get maximum wait timeout for OLM operators:
+	waitTimeout := c.getMaximumOLMTimeout(operators)
+	c.log.Infof("OLM operators %v wait timeout %v", waitTimeout, operators)
+
+	// Wait for the CSV state of the OLM operators, before applying OLM CRs
+	err = utils.WaitForPredicateParamsWithContext(ctx, waitTimeout, GeneralWaitInterval, c.waitForCSVBeCreated, operators)
+	if err != nil {
+		// We continue in case of failure, because we want to try to apply manifest at least for operators which are ready.
+		c.log.WithError(err).Warnf("Failed to wait for some of the OLM operators to be initilized")
+	}
+
 	// Apply post install manifests
-	err = utils.WaitForPredicateWithContext(ctx, retryPostManifestTimeout, GeneralWaitInterval, c.applyPostInstallManifests)
+	err = utils.WaitForPredicateParamsWithContext(ctx, retryPostManifestTimeout, GeneralWaitInterval, c.applyPostInstallManifests, operators)
 	if err != nil {
 		return errors.Wrapf(err, "Failed to apply post manifests")
 	}
 
-	if err != c.waitForOLMOperators(ctx) {
+	if err != c.waitForCSV(ctx, waitTimeout) {
 		// In case the timeout occur, we have to update the pending OLM operators to failed state,
 		// so the assisted-service can update the cluster state to completed.
-		if err = c.updatePendingOLMOperators(); err != nil {
+		if err = c.updatePendingOLMOperators(ctx); err != nil {
 			return errors.Errorf("Timeout while waiting for some of the operators and not able to update its state")
 		}
 		return errors.Wrapf(err, "Timeout while waiting for OLM operators be installed")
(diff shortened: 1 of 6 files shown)


tsorya:
@machacekondra can you please explain why this is needed?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -493,11 +549,48 @@ func (c controller) applyPostInstallManifests() bool {
...
 	}
+	if err = json.Unmarshal(data, &manifests); err != nil {
+		c.log.WithError(err).Errorf("Failed to unmarshall custom manifest file content %s.", data)
+		return false
+	}
+
+	// Create the manifests of the opreators, which are properly initialized:
+	readyOperators, _, err := c.getReadyOperators()
+	if err != nil {
+		return false
```
tsorya, line 566: log please

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -473,6 +496,39 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 	return nil
 }
 
+func (c controller) getReadyOperators() ([]string, []models.MonitoredOperator, error) {
+	var readyOperators []string
+	operators, err := c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID)
```
tsorya, line 501: list of operators can't be changed, why not to provide it once?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -473,6 +496,39 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+	if err != nil {
+		return readyOperators, operators, err
+	}
+
+	for index := range operators {
+		handler := NewClusterServiceVersionHandler(c.kc, &operators[index], c.Status)
```
tsorya, line 507: why not to create handlers once?and provide them to function?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -473,6 +496,39 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+		operatorStatus, _, err := handler.GetStatus()
+		if err != nil {
+			return readyOperators, operators, err
```
tsorya, line 510: i think logging can help

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -455,13 +464,27 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 		return errors.Wrapf(err, "Timeout while waiting for console to become available")
 	}
 
+	// Get maximum wait timeout for OLM operators:
```
tsorya, line 467: can we move all this logic to function? I mean that will be nice to have here only 
waitForOlm... 
and all the flow should be there

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -473,6 +496,39 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+			return readyOperators, operators, err
+		}
+		if utils.IsStatusSucceeded(operatorStatus) {
+			readyOperators = append(readyOperators, handler.GetName())
+		}
+	}
+	return readyOperators, operators, nil
+}
+
+func (c controller) waitForCSVsBeReady() bool {
```
tsorya, line 519: what means waitForCSVsBeReady ? does it mean that we wait till OLM operator is available already? 
I think i just missed something cause i don't know enough regarding the flow.
But doesn't it means that we wait for OLM operator to be ready twice ? or maybe it is another status that we check here? Just trying to understand

src/utils/utils.go
```diff
@@ -287,6 +287,10 @@ func IsStatusFailed(operatorStatus models.OperatorStatus) bool {
 	return operatorStatus == models.OperatorStatusFailed
 }
 
+func IsStatusSucceeded(operatorStatus models.OperatorStatus) bool {
```
tsorya, line 290: i think it should be pard of operators handler file

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Right, we can save some assisted-service API calls.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Done.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Done.

src/utils/utils.go
machacekondra: Done.

src/ops/ops.go
```diff
@@ -565,13 +565,21 @@ func (o *ops) GetHostname() (string, error) {
 	return os.Hostname()
 }
 
-func (o *ops) CreateManifests(kubeconfig string, manifestFilePath string) error {
-	command := fmt.Sprintf("oc --kubeconfig=%s apply -f %s", kubeconfig, manifestFilePath)
+func (o *ops) CreateManifests(kubeconfig string, content []byte) error {
+	// Create temp file, where we store the content to be create by oc command:
```
YuviGold, line 569: I think you forgot to use the `content`

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Done.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Done.

machacekondra:
> @machacekondra can you please explain why this is needed?

Currently, in case an OLM subscription has failed, then we would get stuck on applying the customized manifests since the CRDs aren't available. So with approach introduced in this PR, we wait only for OLM, which are properly subscribed.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: You are right, we wait twice. First we wait to be sure the subscription is properly applied, and so it makes sense to apply also the CR of the OLM operator. Second time we wait that the CR is properly applied and CSV report proper state.

YuviGold: I think it should be `waitForCSVBeCreated`

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -473,6 +496,39 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
 	return nil
 }
 
+func (c controller) getReadyOperators() ([]string, []models.MonitoredOperator, error) {
```
YuviGold, line 499: Maybe I'm wrong, but it seems like we are waiting twice for CSVs to be available.
It should wait for them to be created, and only after everything else finished, we need to look for them to be available (as it is implement ATM)

YuviGold (commented):
Should add new tests

src/ops/ops.go
machacekondra: Right, thanks.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: The problem is that, the CSV report the state the same way `status=succeed` in case of the subscription and CRs, so that's why we wait for same condition. But it's different.

YuviGold:
/approve
@tsorya what do you think?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -455,16 +464,39 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+	if err = c.waitForOLMOperators(ctx); err != nil {
+		return errors.Wrapf(err, "Error while initializing OLM operators")
+	}
+
+	return nil
+}
+
+func (c controller) waitForOLMOperators(ctx context.Context) error {
+	// Get maximum wait timeout for OLM operators:
+	operators, err := c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID)
```
tsorya, line 477: I think this part must be retry-able no?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -455,16 +464,39 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "Error while fetch the monitored operators from assisted-service.")
+	}
+	waitTimeout := c.getMaximumOLMTimeout(operators)
+	c.log.Infof("OLM operators wait timeout %v", waitTimeout)
+
+	// Wait for the CSV state of the OLM operators, before applying OLM CRs
+	err = utils.WaitForPredicateParamsWithContext(ctx, waitTimeout, GeneralWaitInterval, c.waitForCSVBeCreated, operators)
+	if err != nil {
+		c.log.WithError(err).Warnf("Failed to wait for some of the OLM operators to be initilized")
```
tsorya, line 487: can you please comment why we continue in case of failure?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -455,16 +464,39 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+	// Wait for OLM operators
+	if err = c.waitForOLMOperators(ctx); err != nil {
+		return errors.Wrapf(err, "Error while initializing OLM operators")
+	}
+
+	return nil
+}
+
+func (c controller) waitForOLMOperators(ctx context.Context) error {
+	// Get maximum wait timeout for OLM operators:
```
tsorya, line 476: We should have a check that will say of we have olm operators or not.
If we don't have olm operators we should exit immediately as this creates confusing behavior in assisted-service cause we are trying to do stuff that should not work in advance.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Done.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Done.

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: Done.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -455,16 +466,55 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+
+func (c controller) waitForOLMOperators(ctx context.Context) error {
+	var operators []models.MonitoredOperator
+	var err error
+
+	// Get the monitored operators:
+	err = utils.Retry(maxFetchAttempts, FetchRetryInterval, c.log, func() error {
+		operators, err = c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID)
+		if err != nil {
+			return errors.Wrapf(err, "Error while fetch the monitored operators from assisted-service.")
```
tsorya, line 485: log.WithError

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -455,16 +466,55 @@ func (c controller) postInstallConfigs(ctx context.Context) error {
...
+		return errors.Wrapf(err, "Failed to fetch monitored operators")
+	}
+	if len(operators) == 0 {
+		c.log.Info("No OLM operators found.")
+		return nil
+	}
+
+	// Get maximum wait timeout for OLM operators:
+	waitTimeout := c.getMaximumOLMTimeout(operators)
+	c.log.Infof("OLM operators wait timeout %v", waitTimeout)
```
tsorya, line 499: maybe log list of operators too ?

src/assisted_installer_controller/assisted_installer_controller.go
machacekondra: It's logged in `utils.Retry` method.

PR #334, opened 2021-07-27
OCPBUGSM-27526: Cluster deployment freeze on Finalizing stage forever on Cluster Version Operator

tsorya:
Fix for very specific case when cvo has new messages all the time but in
reality it is stuck.
Adding CVOMaxTimeout with 3 hours that will cause waitForPredicate to exit even though timer is updated due to new cvo message

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller_test.go
+++ b/src/assisted_installer_controller/assisted_installer_controller_test.go
@@ -1283,6 +1283,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 		BeforeEach(func() {
 			GeneralProgressUpdateInt = 100 * time.Millisecond
 			WaitTimeout = 150 * time.Millisecond
+			CVOMaxTimeout = 1 * time.Second
 		})
 
 		for i := range tests {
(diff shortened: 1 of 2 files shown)


src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -1283,6 +1283,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 		BeforeEach(func() {
 			GeneralProgressUpdateInt = 100 * time.Millisecond
 			WaitTimeout = 150 * time.Millisecond
+			CVOMaxTimeout = 150 * time.Second
```
eranco74, line 1286: While must tests shouldn't reach this timeout, no test should take more than a second so I think you can set it to:
```suggestion
			CVOMaxTimeout = time.Second
```

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -1345,6 +1346,29 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
 		})
 
+		It("service fail to sync - maxTimeout applied", func() {
+			WaitTimeout = 150 * time.Second
+			CVOMaxTimeout = 150 * time.Millisecond
+			currentServiceCVOStatus := &models.MonitoredOperator{Status: models.OperatorStatusProgressing, StatusInfo: ""}
+			clusterVersionReport := &configv1.ClusterVersion{
+				Status: configv1.ClusterVersionStatus{
+					Conditions: []configv1.ClusterOperatorStatusCondition{
+						{Type: configv1.OperatorAvailable, Status: configv1.ConditionTrue, Message: ""},
```
eranco74, line 1356: I don't get it, this should pass no?
your clusterVersionReport available status is True

src/assisted_installer_controller/assisted_installer_controller_test.go
tsorya: i wanted to set it to something huge

src/assisted_installer_controller/assisted_installer_controller_test.go
tsorya: this will get the same message all the time, you can see that we fail on timeout

YuviGold (approved):
LGTM

PR #336, opened 2021-08-02
MGMT-7417: Change installer and installer controller to support infra…

yevgeny-shnaidman:
…-env for the host

Assisted-installer will get infra-env-id as part of install command aragument, and will use it
to update host progress and to download host ignition
Assisted-installer-controller will use each host's InfraEnvID field to update it progress in
assisted-service


src/config/config.go
```diff
@@ -48,6 +49,7 @@ func ProcessArgs() {
 	ret := &GlobalConfig
 	flag.StringVar(&ret.Role, "role", string(models.HostRoleMaster), "The node role")
 	flag.StringVar(&ret.ClusterID, "cluster-id", "", "The cluster id")
```
ravidbro, line 51: do we still need the 'ClusterID' ?

src/config/config.go
```diff
@@ -15,6 +15,7 @@ import (
 type Config struct {
 	Role                 string
 	ClusterID            string
```
ravidbro, line 17: do we still need the 'ClusterID' ?

src/config/config.go
```diff
@@ -96,4 +98,7 @@ func ProcessArgs() {
 			printHelpAndExit()
 		}
 	}
+	if ret.InfraEnvID == "" {
```
ravidbro, line 101: In which case will it happen?

src/inventory_client/inventory_client.go
```diff
@@ -200,15 +200,15 @@ func (c *inventoryClient) DownloadHostIgnition(ctx context.Context, hostID strin
 	}()
 
 	params := installer.DownloadHostIgnitionParams{
-		ClusterID: c.clusterId,
+		ClusterID: strfmt.UUID(infraEnvID),
```
ravidbro, line 203: Don't we have a new API that accepts InfraEnv ID as a parameter?

src/inventory_client/inventory_client.go
```diff
@@ -291,9 +291,9 @@ func (c *inventoryClient) createDownloadParams(filename string) *installer.Downl
 	}
 }
 
-func (c *inventoryClient) createUpdateHostInstallProgressParams(hostId string, newStage models.HostStage, info string) *installer.UpdateHostInstallProgressParams {
+func (c *inventoryClient) createUpdateHostInstallProgressParams(infraEnvId, hostId string, newStage models.HostStage, info string) *installer.UpdateHostInstallProgressParams {
 	return &installer.UpdateHostInstallProgressParams{
-		ClusterID: c.clusterId,
+		ClusterID: strfmt.UUID(infraEnvId),
```
ravidbro, line 296: same, I think that we should have an API that accepts InfraEnvID and then use that API

src/config/config.go
yevgeny-shnaidman: yes, it is used for access the cluster and getting its hosts by assisted-controller

src/config/config.go
yevgeny-shnaidman: yes, it is used for access the cluster and getting its hosts by assisted-controller

src/config/config.go
yevgeny-shnaidman: we cannot add https://github.com/openshift/assisted-service/pull/2326, until assisted-installer supports its. So, until that PR is pushed, we won't get the infra-env id in command line, and we are setting it to cluster id

src/inventory_client/inventory_client.go
yevgeny-shnaidman: not for downloading host ignitions

src/inventory_client/inventory_client.go
yevgeny-shnaidman: i think we will start moving to that API once we have late binding working, there will be quite a lot of them

src/config/config.go
ravidbro: when is the command line used?
I'm not familiar with that

src/inventory_client/inventory_client.go
ravidbro: I think that we should have, now that we are using InfraEnv to identify the hosts, all the APIs should be under /v2/infra-envs/<infraenvid>/hosts/<hostID>
we don't need new definitions in the swagger but if we want to introduce a new path in the swagger so the path params will be correct.

src/inventory_client/inventory_client.go
ravidbro: but then you will need to come back here and change again the code that you are just putting now,
I'm not sure how big is the task of duplicating host APIs (with the definitions as they are) from v1 to v2

src/config/config.go
eranco74: The command line @yevgeny-shnaidman refers to is the install-command that executes the assisted-installer on the node.

Will we have InfraEnvID when installing without kube-api e.g cloud.redhat.com?

src/installer/installer.go
```diff
@@ -612,7 +612,7 @@ func (i *installer) updateReadyMasters(nodes *v1.NodeList, readyMasters *[]strin
 				return fmt.Errorf("Node %s is not in inventory hosts", node.Name)
 			}
 			ctx = utils.GenerateRequestContext()
-			if err := i.inventoryClient.UpdateHostInstallProgress(ctx, host.Host.ID.String(), models.HostStageJoined, ""); err != nil {
+			if err := i.inventoryClient.UpdateHostInstallProgress(ctx, host.Host.InfraEnvID.String(), host.Host.ID.String(), models.HostStageJoined, ""); err != nil {
```
eranco74, line 615:
```suggestion
			if err := i.inventoryClient.UpdateHostInstallProgress(ctx, config.InfraEnvID, host.Host.ID.String(), models.HostStageJoined, ""); err != nil {
```

src/config/config.go
yevgeny-shnaidman: yes, install command is per host, and each host has its own infra-env id

src/inventory_client/inventory_client.go
yevgeny-shnaidman: adding another version: not changing anything in the download API, since it works at cluster level, changing UpdateHostInstallProgress to V2

src/inventory_client/inventory_client.go
yevgeny-shnaidman: changing UpdateHostInstallProgress to V2

src/installer/installer.go
yevgeny-shnaidman: This is not a correct change - we are updating the status of the other masters, while running on bootstrap node, so we need the other hosts InfraEnvId, and not the InfraEnv ID of the bootstrap node

src/inventory_client/inventory_client.go
```diff
@@ -350,9 +350,9 @@ func (c *inventoryClient) ClusterLogProgressReport(ctx context.Context, clusterI
 	}
 }
 
-func (c *inventoryClient) HostLogProgressReport(ctx context.Context, clusterId string, hostId string, progress models.LogsState) {
+func (c *inventoryClient) HostLogProgressReport(ctx context.Context, infraEnvId string, hostId string, progress models.LogsState) {
 	_, err := c.ai.Installer.UpdateHostLogsProgress(ctx, &installer.UpdateHostLogsProgressParams{
-		ClusterID: strfmt.UUID(clusterId),
+		ClusterID: strfmt.UUID(infraEnvId),
```
ravidbro, line 355: Why the name of the parameter is still ClusterID?
Don't we need an API with InfraEnvID for a case they are different?

yevgeny-shnaidman: fixed. reverted to previous implementation. I am gonna change it in a different ticket

ravidbro:
I see it's still on hold

yevgeny-shnaidman:
> I see it's still on hold

removed

src/installer/installer.go
eranco74: Is it possible that they will have a different infra env???

eranco74: Perhaps I'm missing what does InfraEnv stands for?

ravidbro: You can construct a cluster from hosts that are coming from different InfraEnvs (pools)

eranco74:
/hold
@yevgeny-shnaidman  can you please explain why the assisted installer needs to use infraEnv?
This code gets executed only on hosts that are bounded to a cluster, we definitely have clusterID.

ravidbro:
> /hold
> @yevgeny-shnaidman can you please explain why the assisted installer needs to use infraEnv?
> This code gets executed only on hosts that are bounded to a cluster, we definitely have clusterID.

When using v2 API InfraEnvID is part of the identifier of every host, this is the pathParam in the URL to identify the host

eranco74:
Note that once this merges this assisted-installer will get used by the CI and dev deployments of assisted-service.

PR #337, opened 2021-08-03
MGMT-3817: assisted-controller should always send logs

tsorya:
Right now in case we kube-api-server is not reachable we will not send
    any logs. This code is changing, from now in case kube-api error we will
    send this error as log


osherdp (approved):

osherdp:
/retitle MGMT-3817: assisted-controller should always send logs

PR #340, opened 2021-08-09
MGMT-7450: Removing pull secret token from failure logs

tsorya:
Adding function that will change token value to <SECRET>


eranco74 (commented):
/lgtm

PR #341, opened 2021-08-09
Remove installation timeout

eranco74:


filanov:
Maybe you should open a ticket for it, looks like a big change.

eranco74:
> Maybe you should open a ticket for it, looks like a big change.

Not really, this parameter is obsolete since Oct 19.
No one is using it, so basically just deleting dead code.
There is another [PR](https://github.com/openshift/assisted-service/pull/2355) in assisted-service to delete it from the service side (not used by default or by tests)

filanov (approved):

PR #343, opened 2021-08-11
MGMT-7178: Assisted-installer should not reboot worker till min number of masters have joined

ori-amizur:
/cc @tsorya

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-openapi/swag"
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


src/installer/installer.go
```diff
@@ -133,6 +133,11 @@ func (i *installer) InstallNode() error {
 		}
 		i.log.Info("Setting bootstrap node new role to master")
 
+	} else if i.Config.Role == string(models.HostRoleWorker) {
```
tsorya, line 136: why to wait for controller? i think you should just wait till 2 masters (not bootstrap) moved to Joined or Done

ori-amizur: OK

tsorya, line 136: i am not sure 100% but there is an option it will break day2 flow, as there will be none masters in the cluster

flaper87:
/retitle MGMT-7178: Assisted-installer should not reboot worker till min number of masters have joined

flaper87 (approved):
Please, always add a note when holding a PR

Why is this PR on hold?

ori-amizur:
> Please, always add a note when holding a PR
> 
> Why is this PR on hold?

Waited for https://github.com/openshift/assisted-service/pull/2399

PR #350, opened 2021-08-31
MGMT-7713 Worker won't get updated to 'Waiting for ignition' when installing with IPv6

eranco74:
Updated the regex to allow more chars between the host IP and 'Ignition'
this is required because in the MCS log the host IP is logged as scoped literal IPv6 address
e.g. [fe80::ff:fe9d:12ac%ens3]:42692
This should also allow master nodes to get updated to 'Configuring'

### DIFF
--- a/src/common/common.go
+++ b/src/common/common.go
@@ -45,7 +45,7 @@ func SetConfiguringStatusForHosts(client inventory_client.InventoryClient, inven
 			continue
 		}
 		log.Infof("Verifying if host %s pulled ignition", hostName)
-		pat := fmt.Sprintf("(%s).{1,20}(Ignition)", strings.Join(host.IPs, "|"))
+		pat := fmt.Sprintf("(%s).{1,40}(Ignition)", strings.Join(host.IPs, "|"))
 		pattern, err := regexp.Compile(pat)
 		if err != nil {
 			log.WithError(err).Errorf("Failed to compile regex from host %s ips list", hostName)
--- a/src/common/common_test.go
+++ b/src/common/common_test.go
@@ -41,7 +41,7 @@ var _ = Describe("verify common", func() {
 				IPs: []string{"192.168.126.10", "192.168.11.122", "fe80::5054:ff:fe9a:4738"}},
 				"node1": {Host: &models.Host{InfraEnvID: infraEnvId, ID: &node1Id, Progress: &models.HostProgressInfo{CurrentStage: models.HostStageRebooting}, Role: models.HostRoleMaster}, IPs: []string{"192.168.126.11", "192.168.11.123", "fe80::5054:ff:fe9a:4739"}},
 				"node2": {Host: &models.Host{InfraEnvID: infraEnvId, ID: &node2Id, Progress: &models.HostProgressInfo{CurrentStage: models.HostStageRebooting}, Role: models.HostRoleWorker}, IPs: []string{"192.168.126.12", "192.168.11.124", "fe80::5054:ff:fe9a:4740"}}}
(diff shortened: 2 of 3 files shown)


osherdp (approved):
it works on e2e?

eranco74:
> it works on e2e?

It won't necessarily fix the IPv6 e2e but it should give us better visibility about the worker actual status

osherdp:
> > it works on e2e?
> 
> It won't necessarily fix the IPv6 e2e but it should give us better visibility about the worker actual status

yeah, I just mean if it will go over the required stage of configuring

eranco74:
Doesn't seem to work.
We get the transition to "reached installation stage Waiting for ignition" in e2e-metal-assisted  but not in e2e-metal-assisted-ipv6.

eranco74:
My guess is that the hosts change IP post-reboot, but we need the must-gather logs to verify.

osherdp:
@eranco74 yes, on ipv6 the dhcp server always return different IPs

as for the must-gather failure, hmm, why should it fail to reach the image? I think it should use the proxy...
either way you can use the sosreports for investigation, no?

eranco74:
> @eranco74 yes, on ipv6 the dhcp server always return different IPs

I thought that for that reason we use --append-kargs ip=<inf>:dhcp6 
 
> as for the must-gather failure, hmm, why should it fail to reach the image? I think it should use the proxy...
> either way you can use the sosreports for investigation, no?

Thanks!!!
I managed to get the MCS logs in the sosreport:
We see here the 2 workers and the last master node (bootstrap)
```
2021-08-31T15:47:16.956367277+00:00 stderr F I0831 15:47:16.956062       1 start.go:38] Version: v4.8.0-202107291502.p0.git.d0db136.assembly.stream-dirty (d0db13627e8b48f7021cd4a44d964c55f15b65e1)
2021-08-31T15:47:16.956931531+00:00 stderr F I0831 15:47:16.956906       1 api.go:65] Launching server on :22624
2021-08-31T15:47:16.957157659+00:00 stderr F I0831 15:47:16.957037       1 api.go:65] Launching server on :22623
2021-08-31T15:50:44.452052214+00:00 stderr F I0831 15:50:44.451854       1 api.go:110] Pool worker requested by address:"**[1001:db9::56]:59006**" User-Agent:"Ignition/2.9.0" Accept-Header: "application/vnd.coreos.ignition+json;version=3.2.0, */*;q=0.1"
2021-08-31T15:50:44.888407176+00:00 stderr F I0831 15:50:44.886656       1 api.go:110] Pool worker requested by address:"**[1001:db9::5e]:45220**" User-Agent:"Ignition/2.9.0" Accept-Header: "application/vnd.coreos.ignition+json;version=3.2.0, */*;q=0.1"
2021-08-31T15:51:02.788291244+00:00 stderr F I0831 15:51:02.788224       1 api.go:110] Pool master requested by address:"**[fe80::ff:fee1:22e%br-ex]:59624**" User-Agent:"Ignition/2.9.0" Accept-Header: "application/vnd.coreos.ignition+json;version=3.2.0, */*;q=0.1"

```

And these are the hosts IPv6 addresses we have in the service:
test-infra-cluster-fac6faec-master-2 (bootstrap) - 1001:db9::59
test-infra-cluster-fac6faec-worker-0 - 1001:db9::45
test-infra-cluster-fac6faec-worker-1 -1001:db9::62



This is why the host status isn't updated after they pull the ignition

osherdp:
/unhold
It fixes part of the problem, we'll handle ip changes in another effort

PR #353, opened 2021-09-07
MGMT-7760: Move assisted-installer images to ubi8

flaper87:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 build
+**/Dockerfile.*
--- a/Dockerfile.assisted-installer
+++ b/Dockerfile.assisted-installer
@@ -11,7 +11,7 @@ RUN go mod download
 COPY . .
 RUN make installer
 
-FROM quay.io/centos/centos:centos8
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.4
 
 COPY --from=builder /go/src/github.com/openshift/assisted-installer/build/installer /usr/bin/installer
 COPY --from=builder /go/src/github.com/openshift/assisted-installer/deploy/assisted-installer-controller /assisted-installer-controller/deploy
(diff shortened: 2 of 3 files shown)


.dockerignore
```diff
@@ -1 +1,2 @@
 build
+**/Dockerfile.*
```
tsorya, line 2: why it should be ignored?

flaper87: If we don't, then working/modifying the Dockerfile will invalidated the build cache, since we are doing `COPY . .`

In fairness, this is a micro-optimization but it annoyed it and made my workflow slower. :)

Dockerfile.assisted-installer
```diff
@@ -11,7 +11,7 @@ RUN go mod download
 COPY . .
 RUN make installer
 
-FROM quay.io/centos/centos:centos8
+FROM registry.access.redhat.com/ubi8/ubi-minimal
```
osherdp, line 14: A question
Should we stick with ``latest`` tag? Or rather with something more stable (in the sense of non-changing) e.g. ``8.4``

flaper87: whoops, I actually meant to use 8.4, rather than latest. This reminds me that we should do the same in the image service too.

Dockerfile.assisted-installer-controller
```diff
@@ -13,9 +13,9 @@ RUN go mod download
 COPY . .
 RUN make controller
 
-FROM quay.io/centos/centos:centos8
+FROM registry.access.redhat.com/ubi8/ubi-minimal
```
osherdp, line 16:
```suggestion
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.4
```

osherdp (changes requested):

flaper87:
/close

We recently found out that we won't be able to use ubi8 for these repos. I will, instead, align all images to use `stream:8` so we can bring consistency to IC.

PR #360, opened 2021-09-22
MGMT-7760: Switch to stream8

flaper87:
The original plan was to move all images to ubi8. This is not possible due to the lack
of some packages that are needed for other projects. We are now going to switch all images
to stream8 with the hope that the consistency accross repos will prevent (or help) with
debugging current/future issues in CI.

The goal is to keep component's builds as consistent as possible in the channels we are
releasing them on

### DIFF
--- a/Dockerfile.assisted-installer
+++ b/Dockerfile.assisted-installer
@@ -11,7 +11,7 @@ RUN go mod download
 COPY . .
 RUN make installer
 
-FROM quay.io/centos/centos:centos8
+FROM quay.io/centos/centos:stream8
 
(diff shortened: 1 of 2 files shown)


mkowalski:
/approve

Don't have anything against, but may be missing some CI and build internals to give +2

flaper87:
> /approve
> 
> Don't have anything against, but may be missing some CI and build internals to give +2

FWIW, we are already using `stream:8` for assisted-service:  https://github.com/openshift/assisted-service/blob/be02b47c11302ccc927f26f7130c196504abafc7/Dockerfile.assisted-service#L28

The required images, and mirrors, should all be in place already

mkowalski:
/lgtm

I see, as other services already roll with this, let's just get it

PR #373, opened 2021-11-02
MGMT-7912: Controller: Add cache to the operators status API

tsorya:
There is no need to fetch operator status every couple of seconds,
that's why we are adding cache for operator status for 30 seconds

### DIFF
--- a/src/inventory_client/inventory_client.go
+++ b/src/inventory_client/inventory_client.go
@@ -21,6 +21,7 @@ import (
 	"github.com/thoas/go-funk"
 
 	"github.com/PuerkitoBio/rehttp"
+	ttlCache "github.com/ReneKroon/ttlcache/v2"
 	"github.com/go-openapi/strfmt"
 	"github.com/openshift/assisted-installer/src/utils"
 	"github.com/openshift/assisted-service/client"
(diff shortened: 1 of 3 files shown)


src/inventory_client/inventory_client.go
```diff
@@ -253,26 +258,46 @@ func (c *inventoryClient) GetCluster(ctx context.Context) (*models.Cluster, erro
 	return cluster.Payload, nil
 }
 
-func (c *inventoryClient) GetClusterMonitoredOperator(ctx context.Context, clusterId, operatorName string) (*models.MonitoredOperator, error) {
+func (c *inventoryClient) GetMonitoredOperator(ctx context.Context, clusterId string) (models.MonitoredOperatorsList, error) {
+	cacheKey := fmt.Sprintf("GetMonitoredOperator-%s", clusterId)
+	if val, err := c.cache.Get(cacheKey); err != ttlCache.ErrNotFound {
```
eranco74, line 263: What if err is a different error?

eranco74: NM, if it's good enough for the Go docs it's good enough here
https://pkg.go.dev/github.com/ReneKroon/ttlcache/v2#section-readme

eranco74 (commented):
/lgtm

PR #375, opened 2021-11-07
MGMT-7912: getmonitors operator cache should not return value on any

tsorya:
[MGMT-7912](https://issues.redhat.com/browse/MGMT-7912): getmonitors operator cache should not return value on any error


eranco74 (commented):
/lgtm

eranco74 (commented):
/lgtm

PR #377, opened 2021-11-15
Latest triage issues fixes

tsorya:
[MGMT-8240](https://issues.redhat.com/browse/MGMT-8240): Controller should validate that console and cvo statuses 
were both updated in service
In case we failed to send console operator status but succeeded cvo we will stuck in finalizing

[MGMT-8238](https://issues.redhat.com/browse/MGMT-8238): assisted-controller should send all logs and not just the last 2 hours


mkowalski (approved):

filanov:
@tsorya should it be backported for 2.4.1?

tsorya:
/ retitle Bug 2024159: Controller should validate console and cvo statuses were both updated in service and bring all controller logs

tsorya:
/retitle Bug 2024159: Controller should validate console and cvo statuses were both updated in service and bring all controller logs

PR #381, opened 2021-11-24
MGMT-8330: Minimize number of inventory client retries in assisted-co…

tsorya:
[MGMT-8330](https://issues.redhat.com/browse/MGMT-8330): Minimize number of inventory client retries in assisted-controller

everything in assisted-controller runs in loops, we prefer to fail early on error and to retry on the next loop
this will allow us to show service error more quickly and not to be blocked for half an hour on api call
Setting maximum 15 retries

### DIFF
--- a/src/inventory_client/inventory_client.go
+++ b/src/inventory_client/inventory_client.go
@@ -35,9 +35,9 @@ import (
 )
 
 const (
-	defaultRetryMinDelay = time.Duration(2) * time.Second
-	defaultRetryMaxDelay = time.Duration(10) * time.Second
-	defaultMinRetries    = 10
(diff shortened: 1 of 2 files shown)


omertuc:
Wow I think this is what's causing me tons of issues with swarm tests.... But I didn't realize we have this silent rehttp thing

omertuc:
Wouldn't we want to do the same thing with the installer? Why just the controller?

tsorya:
@omertuc nice question but there was a BZ to make add this retries , i just saw that they are not relevant for controller as controller will try to update stuff in a loop in anycase, though installer will just skip it and move forward

PR #385, opened 2022-01-10
MGMT-8321: try to clean disk even if there is no VG on it

michaellevy101:


src/ops/ops.go
```diff
@@ -504,13 +506,98 @@ func (o *ops) RemovePV(pvName string) error {
...
 
+func (o *ops) IsRaidDevice(device string) bool {
+	output, err := o.ExecPrivilegeCommand(o.logWriter, "mdadm", "-v", "--query", "--detail", "--scan")
+
+	if err != nil {
+		o.log.WithError(err).Errorf("Error occurred while trying to get list of raid devices - continue without cleaning")
+		return false
+	}
+
+	return strings.Contains(output, device)
```
omertuc, line 526: This might trip up on `/dev/sda1` and `/dev/sda12`, is it possible to do parsing that's a little more structured?

michaellevy101: That's ok, We want to do a quick check if the disk itself or one of its partitions are part of the raid array.

omertuc: https://github.com/torvalds/linux/blob/fe91c4725aeed35023ba4f7a1e1adfebb6878c23/drivers/scsi/sd.c#L3415-L3417

src/ops/ops.go
```diff
@@ -504,13 +507,123 @@ func (o *ops) RemovePV(pvName string) error {
...
+		o.log.WithError(err).Errorf("Error occurred while trying to get list of raid devices - continue without cleaning")
+		return false
+	}
+
+	// The device itself or one of its partitions
+	expression, _ := regexp.Compile(device + "[\\d]*")
+
+	for _, raidArrayMembers := range raidDevices {
+		for _, raidMember := range raidArrayMembers {
+			if raidMember == device || expression.MatchString(raidMember) {
```
omertuc, line 532: `raidMember == device` is redundant because the regex already uses `*` rather than `+`

src/ops/ops.go
```diff
@@ -504,13 +507,123 @@ func (o *ops) RemovePV(pvName string) error {
...
+	*/
+	if len(lines) < 2 {
+		return result, nil
+	}
+
+	for i := 0; i < len(lines); i += 2 {
+		fields := strings.Fields(lines[i])
+		raidDeviceName := fields[1]
+
+		raidArrayMembersStr := strings.TrimSpace(lines[i+1])
```
omertuc, line 584: This will cause a panic when `len(lines) % 2 == 1`

src/ops/ops.go
michaellevy101: sure,  but it's a simple and clear

src/ops/ops.go
michaellevy101: I checked before if we have 2 lines at least

src/ops/ops.go
omertuc: Fair enough. I personally find such redundancy confusing, it's makes the intent less clear - it made me question my understanding: "if the regex is capturing the device itself, why do we need this additional condition? what am I missing? what is it for?" and made me re-read a few times until I was sure it's just redundant

src/ops/ops.go
michaellevy101: anyway will make it more robust

src/ops/ops.go
michaellevy101: It's about code styling but not critical for me. will update

src/ops/ops.go
```diff
@@ -504,13 +507,132 @@ func (o *ops) RemovePV(pvName string) error {
...
+			continue
+		}
+
+		fields := strings.Fields(lines[i])
+		raidDeviceName := fields[1]
+		i++
+
+		// Ensuring that we have at least two lines per device.
+		if len(lines) == i {
+			break
```
omertuc, line 588: log warning "malformed mdadm output"

omertuc: in a few other places too

omertuc: Consider even returning a hard error, it's risky parsing free-form text like this already, it's probably not a good idea to keep going while the text clearly doesn't follow the format we assumed

michaellevy101: not sure that we always have 2 lines pre raid device, maybe the raid device has no devices (don't know if possible to remove all the devices from the raid) and has only a single line

src/ops/ops.go
```diff
@@ -504,13 +507,132 @@ func (o *ops) RemovePV(pvName string) error {
...
+
+	return result, nil
+}
+
+func (o *ops) cleanDeviceFromRaidDevice(deviceName string, raidDeviceName string, raidArrayMembers []string) error {
+	raidStopped := false
+
+	for _, raidMember := range raidArrayMembers {
+		// A partition or the device itself is part of the raid array.
+		if strings.HasPrefix(raidMember, deviceName) {
```
omertuc, line 611: Any particular reason `HasPrefix` is used here rather than regex like in `IsRaidDevice`?

src/ops/ops.go
omertuc: ack

src/ops/ops.go
```diff
@@ -504,13 +507,134 @@ func (o *ops) RemovePV(pvName string) error {
...
+				if err != nil {
+					return err
+				}
+
+				raidStopped = true
+			}
+
+			// Clean the raid superblock from the device
+			o.log.Infof("Cleaning raid member %s superblock", raidMember)
+			_, err := o.ExecPrivilegeCommand(o.logWriter, "mdadm", "--zero-superblock", raidMember)
```
omertuc, line 628: Does this have the potential to erase partitions on devices that are not the installation device?

eranco74:
cc: @yevgeny-shnaidman

src/ops/ops.go
omertuc: If yes, we need to bring this up in the UI. We have a warning for all bootable devices being formatted, we'll need to add a warning for devices belonging to a raid array also getting formatted

michaellevy101: no
if expression.MatchString(raidMember)

omertuc: I mean if `/dev/md0` for example is made up of `/dev/sdb1` and `/dev/sdc1` but the installation device is `/dev/sdb`, will `/dev/sdc1` also have its superblock removed?

yevgeny-shnaidman: it will format all the disk participating in the raid. Maybe this is something that should be moved to the discovery stage?
A warning that a picked up boot device is part of a RAID?

michaellevy101: @omertuc no it didn't only the installation device.
@yevgeny-shnaidman already asked @avishayt and @romfreiman about adding a warning and they said that no warning needed

omertuc: So the answer to "Does this have the potential to delete the superblock on devices that are not the installation device?" is "yes" and "We won't add a warning"?

Because "No" and "we won't add a warning" doesn't make much sense, since if the answer is no, there's no need for a warning?

I'm confused

omertuc: > it will format all the disk participating in the raid. Maybe this is something that should be moved to the discovery stage? A warning that a picked up boot device is part of a RAID?

So that sounds like a "yes".

Why do we need to delete all devices in the RAID? Why not just the device we're interested in installing to?

avishayt: I don't see how this would affect other disks - RAID doesn't work that way.  RAID will replicate changes to the RAID device to each disk, but won't track changes to each disk and sync with each other.

avishayt: Agree with Omer.  If we damage the RAID as a byproduct of cleaning the installation disk, that's fine.  But why go out of our way to clean disks that the user didn't ask us to touch?

omertuc: I mean given

```go
if expression.MatchString(raidMember)
```

It does look like the answer is "no" but looks like this then contradicts @yevgeny-shnaidman 's expectations

michaellevy101: I don't understand what you mean, @omertuc @avishayt. The other disks are not touched. 
Asked @avishayt and @romfreiman about creating a warning like "Your installation disk is part of a raid array," which isn't related to affecting other disks but to damaging the raid device.

michaellevy101: As for the raid device itself, I'm not certain if it will be damaged, but even though the user requested us to install on this device

yevgeny-shnaidman: i have missed that part of the code where only the installation disk's superblock is removed , @michaellevy101  is right, only the installation disk is reconfigured.

omertuc: Okay so that's a completely different warning, which was not discussed in this thread, you never mentioned you're talking about a different warning. This is the source of the confusion.

Also it seems that only deleting the RAID from the installation device (which is what both me and @avishayt think is the correct thing to do) is NOT what @yevgeny-shnaidman expected, according to his response in this thread, so maybe he has good reasons of why we do need to wipe all devices after all

yevgeny-shnaidman: @omertuc see my comment above, code is ok

omertuc: Okay so everyone's aligned, conversation can be resolved

PR #413, opened 2022-02-09
NO-ISSUE: Fix error log printing the wrong error variable

omertuc:
`err` didn't get re-assigned so it always held a previously assigned
`nil` leading to an uninformative error message


eranco74 (commented):
/lgtm

PR #423, opened 2022-03-15
MGMT-9358 update assisted-service to master

filanov:
update controller docker image to 1.17
update go mod to 1.17
reduce replace to only what needed, i had to replace
github.com/openshift/api because of the import in assited-service, will
try to resolve it first in the service

Had to change k8s_client.go to use &unstructured.Unstructured{} for the
get call beceause UnstructuredList don't implements all the required
functions for a get call


src/k8s_client/k8s_client.go
```diff
@@ -485,7 +485,7 @@ func (c *k8sClient) GetPodLogsAsBuffer(namespace string, podName string, sinceSe
 }
 
 func (c *k8sClient) IsMetalProvisioningExists() (bool, error) {
-	u := &unstructured.UnstructuredList{}
+	u := &unstructured.Unstructured{}
```
filanov, line 488: @yevgeny-shnaidman i'm not sure about this part, do you know if it's tested somewhere?

filanov:
/hold 
until i clarify https://github.com/openshift/assisted-installer/pull/423#discussion_r826997525

src/k8s_client/k8s_client.go
yevgeny-shnaidman: it was added for 4.7, and tested with 4.7 on real baremetal cluster (3 lab machines). Since then i don't think that we touched it.

PR #427, opened 2022-03-24
MGMT-9709: Modify assisted installer, to use more efficient queries when accessing the assisted service

ori-amizur:
- When worker is waiting for 2 ready master nodes, ask only for masters
  instead of all hosts
- Change locations that only need cluster without hosts, to request only
  cluster.

/assign @filanov 
/assign @avishayt 
/assign @tsorya

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller_test.go
+++ b/src/assisted_installer_controller/assisted_installer_controller_test.go
@@ -189,7 +189,7 @@ var _ = Describe("installer HostRoleMaster role", func() {
 
 	setClusterAsFinalizing := func() {
 		finalizing := models.ClusterStatusFinalizing
-		mockbmclient.EXPECT().GetCluster(gomock.Any()).Return(&models.Cluster{Status: &finalizing}, nil).Times(1)
+		mockbmclient.EXPECT().GetCluster(gomock.Any(), false).Return(&models.Cluster{Status: &finalizing}, nil).Times(1)
 	}
 
 	uploadIngressCert := func(clusterID string) {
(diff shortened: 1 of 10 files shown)


src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -248,8 +248,8 @@ var _ = Describe("installer HostRoleMaster role", func() {
 	Context("Waiting for 3 nodes", func() {
 		It("Set ready event", func() {
 			// fail to connect to assisted and then succeed
-			mockbmclient.EXPECT().GetCluster(gomock.Any()).Return(nil, fmt.Errorf("dummy")).Times(1)
-			mockbmclient.EXPECT().GetCluster(gomock.Any()).Return(nil, nil).Times(3)
+			mockbmclient.EXPECT().GetCluster(gomock.Any(), gomock.Any()).Return(nil, fmt.Errorf("dummy")).Times(1)
```
tsorya, line 251: why everything is Any? don't we know if it will send true or false?

tsorya: thx

src/inventory_client/inventory_client.go
```diff
@@ -250,15 +252,23 @@ func (c *inventoryClient) UploadIngressCa(ctx context.Context, ingressCA string,
 	return aserror.GetAssistedError(err)
 }
 
-func (c *inventoryClient) GetCluster(ctx context.Context) (*models.Cluster, error) {
-	cluster, err := c.ai.Installer.V2GetCluster(ctx, &installer.V2GetClusterParams{ClusterID: c.clusterId})
+func (c *inventoryClient) GetCluster(ctx context.Context, withHosts bool) (*models.Cluster, error) {
```
filanov, line 255: I suggest to turn the boolean into a type, otherwise it's very hard to understand the code without going into the declaration of the function every time

PR #436, opened 2022-04-08
AGENT-105: Add Dockerfiles for OCP release

lranjbar:
This is more of a subtask to [AGENT-80](https://issues.redhat.com//browse/AGENT-80) by createing the Dockerfile.*.ocp for these two components. I tested the builds locally and they build. Will need to test in Openshift CI after merge.

Most of these file changes came in with the `go mod vendor` the changes I actually made are:

- Adding Dockerfile.assisted-installer.ocp and Dockerfile.assisted-installer-controller.ocp
- Adding the `vendor-diff` command to the Makefile. This should be added to the CI as a test to check for diff in running `go mod vendor`
- Removing vendor/ from .gitgnore


omertuc:
Can you add the `vendor-diff` target as a prerequisite to the `ci-lint` target?

`ci-lint` is currently indirectly called by [this](https://github.com/openshift/release/blob/48edf323d5fa2a6914773a6d578efea555982fbb/ci-operator/config/openshift/assisted-installer/openshift-assisted-installer-master.yaml#L102-L108) job  since `ci-lint` is a prerequisite of `lint`, it would be nice to have `vendor-diff` run as part of that job as well

PR #438, opened 2022-04-11
MGMT-9971: removing timeout from waiting for 2 master.

tsorya:
[MGMT-9971](https://issues.redhat.com//browse/MGMT-9971): removing timeout from waiting for 2 master.
Moving back to the logic where timeout is decision is taken only by
service

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -39,7 +39,6 @@ const (
 
 var generalWaitTimeout = 30 * time.Second
 var generalWaitInterval = 5 * time.Second
-var waitForMastersTimeout = 60 * time.Minute
 
 // Installer will run the install operations on the node
 type Installer interface {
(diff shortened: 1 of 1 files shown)


filanov:
@tsorya lint issue `waitForMastersTimeout` is not used

src/installer/installer.go
```diff
@@ -425,40 +424,28 @@ func (i *installer) workerWaitFor2ReadyMasters(ctx context.Context) error {
...
 	i.log.Info("Waiting for 2 ready masters")
 	i.UpdateHostInstallProgress(models.HostStageWaitingForControlPlane, "")
-	start := time.Now()
-	err := utils.WaitForPredicate(waitForMastersTimeout, generalWaitInterval, func() bool {
-		var callErr error
-		cluster, callErr = i.inventoryClient.GetCluster(ctx, false)
-		if callErr != nil {
-			i.log.WithError(callErr).Errorf("Getting cluster %s", i.ClusterID)
-			return false
+	_ = utils.WaitForPredicate(waitForeverTimeout, generalWaitInterval, func() bool {
```
ori-amizur, line 427: We need to bail out if the cluster or host is in error state.  Otherwise, it will loop forever

tsorya: we don't, service will send installation stop command in case cluster is in error

ori-amizur: /lgtm

tsorya:
/retitle Bug 2078213: removing timeout from waiting for 2 master

PR #439, opened 2022-04-12
MGMT-9985: Eliminate use of global variables in assisted-installer

ori-amizur:
In order to enable running the assisted installer in the combined agent,
use of global variables have to be eliminated,

[MGMT-9991](https://issues.redhat.com//browse/MGMT-9991): Separate installer main function to contain only flag
processing and environment variables handling.

Move the rest of the functionality to separate function.

/cc @omertuc

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -3,11 +3,14 @@ package installer
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"path/filepath"
 	"time"
 
 	"github.com/go-openapi/swag"
 	"github.com/google/uuid"
+	"github.com/openshift/assisted-installer/src/main/drymock"
+	"github.com/openshift/assisted-service/pkg/secretdump"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
@@ -23,6 +26,10 @@ import (
 	"github.com/openshift/assisted-installer/src/utils"
 	"github.com/openshift/assisted-service/models"
 )
+
+// In dry run mode we prefer to get quick feedback about errors rather
+// than keep retrying many times.
+const dryRunMaximumInventoryClientRetries = 3
 
 const (
 	InstallDir                   = "/opt/install-dir"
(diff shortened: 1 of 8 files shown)


src/k8s_client/k8s_client.go
```diff
@@ -204,10 +203,6 @@ func (c *k8sClient) ListMachines() (*machinev1beta1.MachineList, error) {
 }
 
 func (c *k8sClient) PatchEtcd() error {
```
omertuc, line 205: This is called by the installer, not the controller as you may have suggested in your Slack messages, so I think you'd have to put that back

omertuc: https://github.com/openshift/assisted-installer/blob/c9c39a06fdcd172a26cd61590cb96bffa031c5b1/src/installer/installer.go#L393

ori-amizur: But isn't the installer client mocked here https://github.com/openshift/assisted-installer/blob/c9c39a06fdcd172a26cd61590cb96bffa031c5b1/src/main/main.go#L69 ?

So this code is not actually called by the installer

ori-amizur: Anyhow, it is never called because Etcd patch is never required (only version less than 4.7)

https://github.com/openshift/assisted-installer/blob/c9c39a06fdcd172a26cd61590cb96bffa031c5b1/src/installer/installer.go#L387

eranco74: We can remove it once we don't need to support 4.6 installations.
Currently, we do:
https://console.redhat.com/openshift/assisted-installer/clusters/~new

ori-amizur: I basically was talking about dry mode flow

src/installer/installer.go
```diff
@@ -825,3 +834,64 @@ func (i *installer) checkLocalhostName() error {
...
+		k8sClientBuilder = k8s_client.NewK8SClient
+	} else {
+		k8sClientBuilder = func(configPath string, logger *logrus.Logger) (k8s_client.K8SClient, error) {
+			var kc k8s_client.K8SClient
+			mockController := gomock.NewController(ginkgo.GinkgoT())
+			kc = k8s_client.NewMockK8SClient(mockController)
+			mock, _ := kc.(*k8s_client.MockK8SClient)
+			drymock.PrepareInstallerDryK8sMock(mock, logger, o, installerConfig.ParsedClusterHosts)
+			return kc, nil
+		}
```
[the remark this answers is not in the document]
ori-amizur: OK

PR #440, opened 2022-04-18
MGMT-7340: add labels to the v2 api

dvachhan:
This PR applies the labels available in the host database to installed nodes.

This PR depends on [openshift/assisted-service#3358](https://github.com/openshift/assisted-service/pull/3358)


src/k8s_client/k8s_client.go
```diff
@@ -584,3 +585,9 @@ func (c *k8sClient) GetCSVFromSubscription(namespace string, name string) (strin
 	}
 	return result.Status.CurrentCSV, nil
 }
+
+func (c *k8sClient) PatchNodeLabels(node *v1.Node, node_labels string) error {
+	data := []byte(`{"metadata": {"labels": ` + node_labels + `}}`)
```
omertuc, line 590: To avoid weird injection / string escaping bugs, please assemble a Metadata struct and serialize it properly rather than using naive string concatenation

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -177,6 +177,43 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
...
+			continue
+		}
+
+		err = c.kc.PatchNodeLabels(&node, hostData.Host.NodeLabels)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to patch %s node with labels %s", node.Name, hostData.Host.NodeLabels)
+			continue
+		}
+	}
+	return ExitWaiting
```
omertuc, line 214: How do you know you've successfully labeled all the nodes that are going to join?

Some nodes join much later than other nodes.

e.g. workers join after control plane nodes - in that case, "ListNodes" at first would give you just the control plane nodes, and only later once the workers join you'll see the worker nodes in that list.

I think your method should loop over `assistedNodesMap`, not over `nodes`, and only once it successfully patched all of them, it should ExitWaiting.

You should also stop ignoring errors - you should keep trying until you've successfully patched all the nodes with the right labels.

dvachhan: @omertuc This func is under PostInstallConfigs which is after WaitAndUpdateNodesStatus. I might be mistaken here but according to my understanding, all the nodes would have already joined by the time we get to PostInstallConfigs. So this should not be an issue. Thoughts?

omertuc: The issue is, is that the finalizing state doesn't imply that all nodes have already joined. Here is why:

The transition between installing and finalizing:
https://github.com/openshift/assisted-service/blob/e251185323c16862f4898380895e0954e86180b0/internal/cluster/statemachine.go#L254-L264

Is fired when the IsFinalizing method is true:

https://github.com/openshift/assisted-service/blob/b365b892b2288e4c4914224e3e3a18db7e84f24f/internal/cluster/transition.go#L328-L339

omertuc: Also don't forget this part of my comment, it's also important:

>You should also stop ignoring errors (i.e. continue) - you should keep trying until you've successfully patched all the nodes with the right labels.

dvachhan: Understood. And yes, I have taken into account the "stop ignoring errors" part that you commented.

src/k8s_client/k8s_client.go
dvachhan: I am not exactly clear on this. It would be helpful if you can point to an example in the code itself that someone has implemented. Thanks.

omertuc: Oh I now see that node_labels is already a JSON blob, so nevermind. Not too important

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -177,6 +177,31 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
...
+	ctxReq := utils.GenerateRequestContext()
+	log := utils.RequestIDLogger(ctxReq, c.log)
+
+	assistedNodesMap, err := c.ic.GetHosts(ctxReq, log, ignoreStatuses)
+	if err != nil {
+		log.WithError(err).Error("Failed to get node map from the assisted service")
+		return KeepWaiting
+	}
+
+	for hostname, hostData := range assistedNodesMap {
```
omertuc, line 191: At some point the assisted installer may give up on some worker nodes, so you probably want to only loop over the hosts that are not yet in the "Error" state.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -177,6 +177,31 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
...
+		if len(hostData.Host.NodeLabels) == 0 {
+			continue
+		}
+		for ok := true; ok; ok = err != nil {
+			err = c.kc.PatchNodeLabels(hostname, hostData.Host.NodeLabels)
+			if err != nil {
+				log.WithError(err).Errorf("Failed to patch node %s with node labels %s", hostname, hostData.Host.NodeLabels)
```
omertuc, line 198: You'll still falling through to `ExitWaiting`, there's no retrying going on here

dvachhan: This is a do-while loop, so, when this statement `err!=nil` is true i.e we encounter an error, the patch function would be executed again.

omertuc: Completely missed that `for` loop. Please use the KeepWaiting mechanism to try again if any of the hosts fail a label update, and only do ExitWaiting if all of them succeeded, this retry-loop inside an already retrying function is redundant.

Also it's a bit weird to fixate on one host, I think it's more neat to go through all of them, skip the ones that fail, and then retry all of the ones that failed in the next run. This way if there's just one bad host it would be more obvious that it's the only bad one, because we can see that rest got their labels correctly.

Also `for ok := true; ok; ok = err != nil` gave me a headache, I don't even want to wrap my head around it. Shouldn't it be `for needRetry := true; needRetry; needRetry = err != nil`? ok is a bit misleading, because it sounds like it means everything is fine while in practice is should mean "please keep looping, there's an error"

src/assisted_installer_controller/assisted_installer_controller.go
dvachhan: Understood. So this check should occur when the patch fails and before we retry, correct?

omertuc: Not sure I understand the question, but the gist is that the map should have any hosts in error state removed from it

omertuc: You can also use `ListClusterHostsParams` like in they do in [here](https://github.com/openshift/assisted-installer/blob/89ae514be1699f8b6f5d2e6d6d1ed4ce4e14c2ef/src/inventory_client/inventory_client.go#L265) for the role, but for the "status" of the host instead, to get the service to do the filtering for you

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -973,6 +978,63 @@ func (c controller) waitingForClusterOperators(ctx context.Context) error {
...
+	if err != nil {
+		log.WithError(err).Error("Failed to get node map from the assisted service")
+		return KeepWaiting
+	}
+
+	nodes, err := c.kc.ListNodes()
+	if err != nil {
+		log.WithError(err).Error("Failed to get list of nodes from k8s client")
+		return KeepWaiting
+	} else if len(nodes.Items) != len(assistedNodesMap) {
```
omertuc, line 1010: This is not good enough, (assisted) hosts that are in error may or may not appear as nodes in the nodes list, which will cause a discrepancy in the list lengths and fail this condition forever. Why have you decided to undo the "loop over the assistedNodesMap" that you had before?

dvachhan: Can we not just remove the nodes from ListNodes which are in an error state before comparing the lengths?

omertuc: You could, but it's still feels like the wrong way to do it.

For example, another (admittedly obscure) case where you would still go wrong in your code even after the fix you just proposed, is the following (very stupid) DOS attack:

1. User starts installation
2. Installation is finalizing
3. Attacker uses the (public) worker.ign machine config to create nodes that try to join the cluster
4. This, if I recall correctly, creates a node resource with unapproved CSRs
5. Your code gets stuck because of length mismatch

The bottom line is that your goal is to look at the (non-error) hosts in the assisted API, find their corresponding nodes, and add the labels to those nodes. You're doing it the other way around (listing the nodes and finding their matching AI api hosts) it and makes these weird edge cases possible (needing to check for length match, having to remove error nodes from ListNodes, having to ignore external non-AI nodes that someone might add).

dvachhan:
@omertuc Any chance the tests that are failing are flaky?

PR #448, opened 2022-04-26
MGMT-9244: Prevent assisted-agent from rebooting the node when managed by IPA

eranco74:
In case ironic-agent.service exists on the node the assisted-installer will not reboot the node. 
Instead, it will stop the assisted-agent (agent.service) in order to let the ironic-agent know that the assisted-agent served its purpose, and let the ironic agent reboot the node

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/go-openapi/swag"
@@ -165,11 +166,36 @@ func (i *installer) InstallNode() error {
 	if err != nil {
 		i.log.Errorf("upload installation logs %s", err)
 	}
+	return i.finalize()
+}
+
+func (i *installer) finalize() error {
 	//update installation progress
 	i.UpdateHostInstallProgress(models.HostStageRebooting, "")
-	//reboot
-	if err = i.ops.Reboot(); err != nil {
+	if i.DryRunEnabled {
+		_, err := i.ops.ExecPrivilegeCommand(nil, "touch", i.Config.FakeRebootMarkerPath)
+		return errors.Wrap(err, "failed to touch fake reboot marker")
+	}
+
+	// in case ironic-agent exists on the host we should stop the assisted-agent service instead of rebooting the node.
+	// the assisted agent service stop will signal the ironic agent that we are done so that IPA can continue with its flow.
+	// regardless, we update the host install progress to `rebooting` since the node will get rebooted shortly after.
+	ironicAgentServiceName := "ironic-agent.service"
+	out, err := i.ops.ExecPrivilegeCommand(nil, "systemctl", "list-units", "--no-legend", ironicAgentServiceName)
+	if err != nil {
+		i.log.Errorf("Failed to check if ironic agent service exists on the node %s", err)
 		return err
+	}
+	if strings.Contains(out, ironicAgentServiceName) {
+		// in case the ironic-agent exists the installer should stop the assisted agent.service
+		if err = i.ops.SystemctlAction("stop", "agent.service"); err != nil {
+			return err
+		}
+	} else {
+		//reboot
+		if err = i.ops.Reboot(); err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/src/installer/installer_test.go
+++ b/src/installer/installer_test.go
@@ -130,6 +130,9 @@ var _ = Describe("installer HostRoleMaster role", func() {
 	rebootSuccess := func() {
 		mockops.EXPECT().Reboot().Return(nil).Times(1)
 	}
+	ironicAgentDoesntExist := func() {
+		mockops.EXPECT().ExecPrivilegeCommand(gomock.Any(), "systemctl", "list-units", "--no-legend", "ironic-agent.service").Return("", nil).Times(1)
+	}
 	BeforeEach(func() {
 		ctrl = gomock.NewController(GinkgoT())
 		mockops = ops.NewMockOps(ctrl)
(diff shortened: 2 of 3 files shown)


src/installer/installer.go
```diff
@@ -167,10 +168,25 @@ func (i *installer) InstallNode() error {
 	}
 	//update installation progress
 	i.UpdateHostInstallProgress(models.HostStageRebooting, "")
-	//reboot
-	if err = i.ops.Reboot(); err != nil {
+	// if ironic-agent exists on the host stop the assisted-agent service and exit, else reboot
```
omertuc, line 171: what's happening is pretty obvious from the code, could be nice instead to have a comment explaining _why_ we'd want to do that

src/installer/installer.go
```diff
@@ -167,10 +168,25 @@ func (i *installer) InstallNode() error {
...
+	ironicAgentServiceName := "ironic-agent.service"
+
+	out, err := i.ops.ExecPrivilegeCommand(nil, "systemctl", "list-units", "--no-legend", ironicAgentServiceName)
+	if err != nil {
+		i.log.Errorf("Failed to check if ironic agent service exists on the node %s", err)
 		return err
 	}
+	if strings.Contains(out, ironicAgentServiceName) {
```
omertuc, line 179: Can we move this logic to the systemd unit file level, and have this logic be a simple configuration at the agent level? Something like "AVOID_REBOOT: true, SELF_SERVICE_NAME: agent.service"

omertuc: Having very specific logic like this makes it harder to run the agent in "unnatural" environments, like the swarm.

eranco74: This means someone else will need to set this configuration.
With the current implementation, the only place that cares about the ironic service existence and how to handle it is here. 
Side note, the assisted-installer isn't a service, in case ironic agent also exists on the host the installer will stop the assisted service agent service: `agent.service`

omertuc: >This means someone else will need to set this configuration.

Exactly, I want the unit file that runs the agent to:

1. Check if there's an ironic service defined in the system
2. If yes, configure the agent to AVOID_REBOOT
3. Tell the agent its own service name through SELF_SERVICE_NAME, set it to `agent.service`

But when you're putting all this logic inside the agent itself, you're assuming the agent:
1. Runs in a systemd environment
2. Assuming the agent itself runs as a systemd service
3. Assuming the agent runs in a systemd service called "agent.service"

All these assumptions make it harder to run the agent in different environments - for example, as a library, like @ori-amizur  is working on

omertuc: >Side note, the assisted-installer isn't a service, in case ironic agent also exists on the host the installer will stop the assisted service agent service: agent.service

Not sure I understand this part, can you elaborate?

eranco74: > Not sure I understand this part, can you elaborate?
Yes, and I think it will explain why I prefer putting it all here.
This code isn't the agent.service, this code get compiled and shipped as the assisted-installer image.
The assisted-installer image get executed (podman run... ) when the assisted-agent get the install command.

omertuc: :man_facepalming: ugh I got confused and then thought we're in the agent repo lol.

eranco74: > logic like this makes it harder to run the agent in "unnatural" environments, like the swarm.
So I can add:
```
if i.DryRunEnabled {
		return nil
	}
```
As we do for all other systemctl operations

omertuc: My point still stands though - the agent should be configured through the systemd unit to know its own systemd service name, and it should be configured to *tell the installer* (through the installer CLI args) to avoid a reboot and close the agent instead

eranco74: What will we gain from it?
Except from splitting the logic to 2 different code bases?
Why does the agent service needs to care about the ironic agent?
It's not different than saying the assisted-agent should check the hostname and pass it in a cli option instead of checking for it in the installer code

tsorya: hmm, can't it be done from service side? i mean don't we have some configuration saying we are running with this specific mode? 
I am not sure regarding stopping agent in installer as installer is ran from agent , sort of killing your own father?
As installer command is known to agent we can add params to it easy from agent itself and then all logic will be in agent and i think this logic should be part of agent and not installer.

eranco74: We pass a flag from from the service side, we can also add API for the user to control it...
But why would we want to?
The only place that cares about and knows how to perform the installation is here.

src/installer/installer.go
```diff
@@ -165,12 +166,37 @@ func (i *installer) InstallNode() error {
...
+	if i.DryRunEnabled {
+		_, err := i.ops.ExecPrivilegeCommand(nil, "touch", i.Config.FakeRebootMarkerPath)
+		return errors.Wrap(err, "failed to touch fake reboot marker")
+	}
+
+	// in case ironic-agent exists on the host we should stop the assisted-agent service instead of rebooting the node.
+	// the assisted agent service stop will signal the ironic agent that we are done so that IPA can continue with its flow.
+	// regardless, we update the host install progress to `rebooting` since the node will get rebooted shortly after.
+	ironicAgentServiceName := "ironic-agent.service"
+	out, err := i.ops.ExecPrivilegeCommand(nil, "systemctl", "list-units", "--no-legend", ironicAgentServiceName)
```
flaper87, line 184: This check assumes that we will *always* have the socked mounted so that we can use `systemctl`. I think we already have it but just wanted to be explicit about it as I don't know if we already depend on it. I think.... we do

eranco74: The installer runs multiple other systemctl commands (it doesn't use a socket but have another trick instead :shushing_face: ), so no need to worry, this didn't add any dependency.

PR #451, opened 2022-05-05
MGMT-9379: Disregard HA mode for worker.

paul-maidment:
When adding a worker to a 2nd day cluster, HA mode is irrelevant, so
we shall not be passing it here.

It used to be the case that a day 2 worker could only be added to an HA
cluster. We are making this change to allow the addition of a day 2
worker to a non HA cluster.

The condition [1] was blocking this.

That validation ensures that Day 1 SNO clusters will only accept Master
or Bootstrap roles for the host being installed.

This change will validate that the installer has not been called with
high availability mode set for a worker, if this has occurred, the
installer will reject the request with an error.

[1] https://github.com/openshift/assisted-installer/blob/d67d2843d0e757ef5643e6c873b49e61ef83a0ea/src/config/config.go#L93-L99

### DIFF
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -5,10 +5,7 @@ import (
 	"flag"
 	"os"
 
-	"github.com/thoas/go-funk"
-
 	"github.com/openshift/assisted-installer/src/utils"
-
 	"github.com/openshift/assisted-service/models"
 )
 
@@ -39,7 +36,10 @@ type Config struct {
 	DisksToFormat        ArrayFlags
 }
 
-func printHelpAndExit() {
+func printHelpAndExit(err error) {
+	if err != nil {
+		println(err)
+	}
 	flag.CommandLine.Usage()
 	os.Exit(0)
 }
@@ -65,40 +65,46 @@ func ProcessArgs() *Config {
 	flag.StringVar(&ret.HTTPSProxy, "https-proxy", "", "A proxy URL to use for creating HTTPS connections outside the cluster")
 	flag.StringVar(&ret.NoProxy, "no-proxy", "", "A comma-separated list of destination domain names, domains, IP addresses, or other network CIDRs to exclude proxying")
 	flag.StringVar(&ret.ServiceIPs, "service-ips", "", "All IPs of assisted service node")
-	flag.StringVar(&ret.HighAvailabilityMode, "high-availability-mode", "Full", "high-availability expectations. default is 'Full', which represents the behavior in a \"normal\" cluster. Use 'None' for single-node deployment")
+	flag.StringVar(&ret.HighAvailabilityMode, "high-availability-mode", "", "high-availability expectations, \"Full\" which represents the behavior in a \"normal\" cluster. Use 'None' for single-node deployment. Leave this value as \"\" for workers as we do not care about HA mode for workers.")
 	flag.BoolVar(&ret.CheckClusterVersion, "check-cluster-version", false, "Do not monitor CVO")
 	flag.StringVar(&ret.MustGatherImage, "must-gather-image", "", "Custom must-gather image")
 	flag.Var(&ret.DisksToFormat, "format-disk", "Disk to format. Can be specified multiple times")
 
 	var installerArgs string
 	flag.StringVar(&installerArgs, "installer-args", "", "JSON array of additional coreos-installer arguments")
-
 	h := flag.Bool("help", false, "Help message")
 	flag.Parse()
+
+	if err := ret.SetInstallerArgs(installerArgs); err != nil {
+		printHelpAndExit(err)
+	}
+
+	if h != nil && *h {
+		printHelpAndExit(nil)
+	}
 
 	if ret.NoProxy != "" {
 		utils.SetNoProxyEnv(ret.NoProxy)
 	}
-	if h != nil && *h {
-		printHelpAndExit()
+
+	ret.SetDefaults()
+	return ret
+}
+
+func (c *Config) SetInstallerArgs(installerArgs string) error {
+	if installerArgs != "" {
+		return json.Unmarshal([]byte(installerArgs), &c.InstallerArgs)
+	}
+	return nil
+}
+
+func (c *Config) SetDefaults() {
+	if c.Role == string(models.HostRoleWorker) {
+		//High availability mode is not relevant to workers, so make sure we clear this.
+		c.HighAvailabilityMode = ""
 	}
 
-	if installerArgs != "" {
-		err := json.Unmarshal([]byte(installerArgs), &ret.InstallerArgs)
-		if err != nil {
-			println(err.Error())
-			printHelpAndExit()
-		}
+	if c.InfraEnvID == "" {
+		c.InfraEnvID = c.ClusterID
 	}
-	if ret.HighAvailabilityMode == models.ClusterHighAvailabilityModeNone {
-		validRoles := []string{string(models.HostRoleMaster), string(models.HostRoleBootstrap)}
-		if !funk.ContainsString(validRoles, ret.Role) {
-			println("high-availability-mode is set to None, but host role is %s. should be one of: %s", ret.Role, validRoles)
-			printHelpAndExit()
-		}
-	}
-	if ret.InfraEnvID == "" {
-		ret.InfraEnvID = ret.ClusterID
-	}
-	return ret
 }
--- a/src/config/config_test.go
+++ b/src/config/config_test.go
@@ -1,0 +1,89 @@
+package config
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/assisted-service/models"
+)
+
+func TestConfig(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "config_test")
+}
+
+var _ = Describe("SetInstallerArgs", func() {
+
+	It("Should not deserialize installer args when they are not supplied.", func() {
+		config := &Config{}
+		Expect(config.SetInstallerArgs("")).To(BeNil())
+		Expect(len(config.InstallerArgs)).To(BeZero())
+	})
+
+	It("Should deserialize installer args correctly when they are supplied correctly.", func() {
+		config := &Config{}
+		Expect(config.SetInstallerArgs("[\"arg1=foo\", \"arg2=bar\"]")).To(BeNil())
+		Expect(len(config.InstallerArgs)).To(Equal(2))
+		Expect(config.InstallerArgs[0]).To(Equal("arg1=foo"))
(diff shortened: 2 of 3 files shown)


omertuc:
There's no point in retesting this while the service PR hasn't merged, as you're just constantly hitting:
```
May 10 19:07:24 test-infra-cluster-34a7fdfa-worker-1 next_step_runne[1822]: time="10-05-2022 19:07:24" level=info msg="Sending step <install-3506586e> reply output <> error <high-availability-mode was specified when adding a worker, this should be empty (\"\") as it is not relevant for workers.\nUsage of /usr/bin/installer:\n  -agent-image string\n    \tAssisted Installer Agent image URL that will be used to send logs on successful installation (default \"quay.io/ocpmetal/assisted-installer-agent:latest\")\n  -alsologtostderr\n    \tlog to standard error as well as files\n  -boot-device string\n    \tThe boot device\n  -cacert string\n    \tPath to custom CA certificate in PEM format\n  -check-cluster-version\n    \tDo not monitor CVO\n  -cluster-id string\n    \tThe cluster id\n  -controller-image string\n    \tAssisted Installer Controller image URL (default \"quay.io/ocpmetal/assisted-installer-controller:latest\")\n  -format-disk value\n    \tDisk to format. Can be specified multiple times\n  -ginkgo.debug\n    \tIf set, ginkgo will emit node output to files when running in parallel.\n  -ginkgo.dryRun\n    \tIf set, ginkgo will walk the test hierarchy without actually running anything.  Best paired with -v.\n  -ginkgo.failFast\n    \tIf set, ginkgo will stop running a test suite after a failure occurs.\n  -ginkgo.failOnPending\n    \tIf set, ginkgo will mark the test suite as failed if any specs are pending.\n  -ginkgo.flakeAttempts int\n    \tMake up to this many attempts to run each spec. Please note that if any of the attempts succeed, the suite will not be failed. But any failures will still be recorded. (default 1)\n  -ginkgo.focus value\n    \tIf set, ginkgo will only run specs that match this regular expression. Can be specified multiple times, values are ORed.\n  -ginkgo.noColor\n    \tIf set, suppress color output in default reporter.\n  -ginkgo.noisyPendings\n    \tIf set, default reporter will shout about pending tests. (default true)\n  -ginkgo.noisySkippings\n    \tIf set, default reporter will shout about skipping tests. (default true)\n  -ginkgo.parallel.node int\n    \tThis worker node's (one-indexed) node number.  For running specs in parallel. (default 1)\n  -ginkgo.parallel.streamhost string\n    \tThe address for the server that the running nodes should stream data to.\n  -ginkgo.parallel.synchost string\n    \tThe address for the server that will synchronize the running nodes.\n  -ginkgo.parallel.total int\n    \tThe total number of worker nodes.  For running specs in parallel. (default 1)\n  -ginkgo.progress\n    \tIf set, ginkgo will emit progress information as each spec runs to the GinkgoWriter.\n  -ginkgo.randomizeAllSpecs\n    \tIf set, ginkgo will randomize all specs together.  By default, ginkgo only randomizes the top level Describe, Context and When groups.\n  -ginkgo.regexScansFilePath\n    \tIf set, ginkgo regex matching also will look at the file path (code location).\n  -ginkgo.reportFile string\n    \tOverride the default reporter output file path.\n  -ginkgo.reportPassed\n    \tIf set, default reporter prints out captured output of passed tests.\n  -ginkgo.seed int\n    \tThe seed used to randomize the spec suite. (default 1652209644)\n  -ginkgo.skip value\n    \tIf set, ginkgo will only run specs that do not match this regular expression. Can be specified multiple times, values are ORed.\n  -ginkgo.skipMeasurements\n    \tIf set, ginkgo will skip any measurement specs.\n  -ginkgo.slowSpecThreshold float\n    \t(in seconds) Specs that take longer to run than this threshold are flagged as slow by the default reporter. (default 5)\n  -ginkgo.succinct\n    \tIf set, default reporter prints out a very succinct report\n  -ginkgo.trace\n    \tIf set, default reporter prints out the full stack trace when a failure occurs\n  -ginkgo.v\n    \tIf set, default reporter print out all specs as they begin.\n  -help\n    \tHelp message\n  -high-availability-mode string\n    \thigh-availability expectations, \"Full\" which represents the behavior in a \"normal\" cluster. Use 'None' for single-node deployment. Leave this value as \"\" for workers as we do not care about HA mode for workers.\n  -host-id string\n    \tThis host id\n  -http-proxy string\n    \tA proxy URL to use for creating HTTP connections outside the cluster\n  -https-proxy string\n    \tA proxy URL to use for creating HTTPS connections outside the cluster\n  -infra-env-id string\n    \tThis host infra env id\n  -insecure\n    \tDo not validate TLS certificate\n  -installer-args string\n    \tJSON array of additional coreos-installer arguments\n  -kubeconfig string\n    \tPaths to a kubeconfig. Only required if out-of-cluster.\n  -log_backtrace_at value\n    \twhen logging hits line file:N, emit a stack trace\n  -log_dir string\n    \tIf non-empty, write log files in this directory\n  -logtostderr\n    \tlog to standard error instead of files\n  -mco-image string\n    \tMCO image to install\n  -must-gather-image string\n    \tCustom must-gather image\n  -no-proxy string\n    \tA comma-separated list of destination domain names, domains, IP addresses, or other network CIDRs to exclude proxying\n  -openshift-version string\n    \tOpenshift version to install (default \"4.4\")\n  -role string\n    \tThe node role (default \"master\")\n  -service-ips string\n    \tAll IPs of assisted service node\n  -stderrthreshold value\n    \tlogs at or above this threshold go to stderr\n  -url string\n    \tThe BM inventory URL, including a scheme and optionally a port (overrides the host and port arguments\n  -v value\n    \tlog level for V logs\n  -verbose\n    \tIncrease verbosity, set log level to debug\n  -vmodule value\n    \tcomma-separated list of pattern=N settings for file-filtered logging\n> exit-code <0>" file="step_processor.go:70" request_id=384db051-dd7d-4672-964a-6f1bbd1bac82
```

src/config/config.go
```diff
@@ -65,40 +69,59 @@ func ProcessArgs() *Config {
...
+			return fmt.Errorf("high-availability-mode is set to None, but host role is %s. should be one of: %s", c.Role, validRoles)
 		}
 	}
-	if ret.InfraEnvID == "" {
-		ret.InfraEnvID = ret.ClusterID
+	return nil
+}
+
+func (c *Config) SetDefaults() {
+	if c.HighAvailabilityMode != "" && c.Role == string(models.HostRoleWorker) {
```
tsorya, line 119: not sure i understand why we are doing it here. Change agent code and don't send it in case it is worker

src/config/config.go
```diff
@@ -65,40 +69,59 @@ func ProcessArgs() *Config {
...
+	}
+
 	if ret.NoProxy != "" {
 		utils.SetNoProxyEnv(ret.NoProxy)
 	}
-	if h != nil && *h {
-		printHelpAndExit()
+
+	ret.SetDefaults()
+	if err := ret.Validate(); err != nil {
```
eranco74, line 95: What's the point?
The only thing that r.Validate() is validating is role == worker  and HighAvailabilityMode = none, and that can't happen anymore due to SetDefaults.

eranco74 (commented):
Just pass an empty string for  HighAvailabilityMode when adding a worker (as you explain in the StringVar description) I don't thing we need any other code changes.
/hold

src/config/config.go
tsorya: after discussion i understand why it is needed. 
As for me there is no need to check c.HighAvailabilityMode != ""

src/config/config_test.go
```diff
@@ -0,0 +1,94 @@
...
+			config := &Config{
+				ClusterID:            "0ae63135-5f7c-431e-9c72-0efaf2cb83b8",
+				Role:                 string(models.HostRoleMaster),
+				HighAvailabilityMode: models.ClusterHighAvailabilityModeFull,
+			}
+			config.SetDefaults()
+			Expect(config.HighAvailabilityMode).To(Equal(string(models.ClusterHighAvailabilityModeFull)))
+		})
+
+		It("HighAvailabilityMode should not be normalised to empty string if the host role is bootstrap.", func() {
```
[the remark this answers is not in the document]
paul-maidment: this is fair

src/config/config_test.go
```diff
@@ -0,0 +1,94 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "config_test")
+}
+
+var _ = Describe("Config tests", func() {
+	BeforeEach(func() {
+
+	})
+
+	Context("Fetching installer args", func() {
```
carbonin, line 21: A few things here.
`BeforeEach` shouldn't be present if it's empty.

While `Describe` and `Context` function identically they have slightly different semantics.
Typically `Describe` would be used with a specific function or method name. In BDD testing frameworks the idea is that you're _describing_ how the object or method should function.

A context would be used with shared setup code to define the _context_ for the individual examples (tests). So if you wanted to get really pedantic you could do something like:

```go
var _ = Describe("SetInstallerArgs", func() {
  var args string
  Context("with valid args", func() {
    BeforeEach(func() {
      args = "[\"arg1=foo\", \"arg2=bar\"]"
    })
    It("sets the value correctly", func() {
      ...
    })
  })
})
var _ = Describe("SetDefaults", func() {
...
```

The idea is to have the whole thing read like a sentence describing what kind of behavior the test is expecting so if this test fails you would see `SetInstallerArgs with valid args sets the value correctly`. You then immediately know what case isn't working for what function.

You don't have to go quite that far in this PR, I generally will put the setup code in the example when there would only be one case in a context, but I do think using top-level `Describe` blocks for the different functions and getting rid of the context and empty `BeforeEach` block would be an improvement.

src/config/config.go
```diff
@@ -65,40 +65,46 @@ func ProcessArgs() *Config {
...
-		validRoles := []string{string(models.HostRoleMaster), string(models.HostRoleBootstrap)}
-		if !funk.ContainsString(validRoles, ret.Role) {
-			println("high-availability-mode is set to None, but host role is %s. should be one of: %s", ret.Role, validRoles)
-			printHelpAndExit()
-		}
+	return nil
+}
+
+func (c *Config) SetDefaults() {
+	if c.Role == string(models.HostRoleWorker) {
```
eranco74, line 102: Why do we need this if it's already taken care of by the service?
https://github.com/openshift/assisted-service/pull/3718/files
Can we keep the role validation instead?

paul-maidment: The main issue here is that UserManagedNetworking can be set to "None" when you are adding a working.
This occurs when attempting to add a Day2 worker to an existing SNO cluster.

The original solution was to take care of this earlier in the process by not passing the parameter to the installer, however this has proved to be problematic as it will require a breaking change to the service and there is presently no good mechanism for ensuring that all users are running the correct version of the agent.

After discussion with Avishay and Omer, we decided to deal with the issue here instead.

eranco74 (commented):
/lgtm

PR #457, opened 2022-05-11
NO-ISSUE: remove dependabot bumps of docker images

osherdp:
Since it requires D/S changes and changes in openshift/release repo, it isn't really something that we'd like to go in automatically.

In addition, we no longer need the ok-to-test label, as dependabot is a trusted-app.


eliorerz (approved):
LGTM

PR #459, opened 2022-05-11
NO-ISSUE: Reverting removal of dependabot image bumps

osherdp:
Reverting most of https://github.com/openshift/assisted-installer/pull/457 because we would still want to have reminders for new docker images.

For more context: https://github.com/openshift/assisted-installer-agent/pull/352#issuecomment-1123333744

/cc @omertuc @eliorerz

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,3 +16,10 @@ updates:
       - "dependencies"
     commit-message:
       prefix: "NO-ISSUE"
+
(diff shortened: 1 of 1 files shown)


osherdp:
/test images
(seems like a known flake now being discussed on slack)

osherdp:
/override ci/prow/images
will not compromise CI coverage as the change has nothing to do with this job

PR #460, opened 2022-05-11
Bug 2086273: Cleanup disk which is part of a raid and have lvm on it

michaellevy101:
Cleanup Raid PVs and VGs before stopping it and removing the installation disk membership.
Solving "mdadm: Cannot get exclusive access to /dev/md0" When trying to stop a raid with LVM on it.

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -734,7 +734,42 @@ func (i *installer) cleanupInstallDevice() error {
 		return nil
 	}
 
-	vgName, err := i.ops.GetVGByPV(i.Device)
+	err := i.cleanupDevice(i.Device)
+
+	if err != nil {
+		return err
+	}
+
+	if i.ops.IsRaidMember(i.Device) {
+		i.log.Infof("A raid was detected on the device (%s) - cleaning", i.Device)
+		var devices []string
+		devices, err = i.ops.GetRaidDevices(i.Device)
+
+		if err != nil {
+			return err
+		}
+
+		for _, device := range devices {
+			// Cleaning the raid device itself before removing membership.
+			err = i.cleanupDevice(device)
+
+			if err != nil {
+				return err
+			}
+		}
+
+		err = i.ops.CleanRaidMembership(i.Device)
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return i.ops.Wipefs(i.Device)
+}
+
+func (i *installer) cleanupDevice(device string) error {
+	vgName, err := i.ops.GetVGByPV(device)
 	if err != nil {
 		return err
 	}
@@ -747,23 +782,14 @@ func (i *installer) cleanupInstallDevice() error {
 			return err
 		}
 
-		err = i.ops.RemovePV(i.Device)
+		err = i.ops.RemovePV(device)
 
 		if err != nil {
 			return err
 		}
 	}
 
-	if i.ops.IsRaidDevice(i.Device) {
-		i.log.Infof("A raid was detected on the installation device (%s) - cleaning", i.Device)
-		err = i.ops.CleanRaidDevice(i.Device)
-
-		if err != nil {
-			return err
-		}
-	}
-
-	return i.ops.Wipefs(i.Device)
+	return nil
 }
 
 func (i *installer) verifyHostCanMoveToConfigurationStatus(inventoryHostsMapWithIp map[string]inventory_client.HostData) {
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -46,8 +46,9 @@ type Ops interface {
 	RemoveLV(lvName, vgName string) error
 	RemovePV(pvName string) error
 	Wipefs(device string) error
-	IsRaidDevice(device string) bool
-	CleanRaidDevice(device string) error
+	IsRaidMember(device string) bool
+	GetRaidDevices(device string) ([]string, error)
+	CleanRaidMembership(device string) error
 	GetMCSLogs() (string, error)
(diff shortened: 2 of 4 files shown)


omertuc:
Ticket says "can you please open a BZ so we can backport to ACM2.5?", so this PR should also be using a BZ so it can be backported

src/installer/installer.go
```diff
@@ -734,7 +734,38 @@ func (i *installer) cleanupInstallDevice() error {
 		return nil
 	}
 
-	vgName, err := i.ops.GetVGByPV(i.Device)
+	error := i.cleanupDevice(i.Device)
```
omertuc, line 737: Rename to `err` as you're shadowing the `error` type

src/installer/installer.go
```diff
@@ -734,7 +734,38 @@ func (i *installer) cleanupInstallDevice() error {
...
+
+		devices, error := i.ops.GetRaidDevices(i.Device)
+
+		if error != nil {
+			return error
+		}
+
+		for _, device := range devices {
+			// Cleaning the raid device itself before removing membership.
+			i.cleanupDevice(device)
```
omertuc, line 754: So now we *do* wipe all devices in the array, regardless of whether they're the installation disk or not? As discussed in your previous PR, this would have to come with a UI warning, and we have an entire epic (MGMT-7409) about disk preservation that would have to address this

omertuc: Or maybe I misunderstand. Can you please explain how the new behavior in this PR differs from the existing behavior?

michaellevy101: Raid device consists by multiple devices. 
In case of PV on the raid device we remove it (the PV but not wipe the raid device). 
Before this PR we stopped the raid device and remove the installation device from it which means destroying the data.
We don't wipe or remove the other disks from the raid array.
IMO destroying the RAID justifies a warning @avishayt @romfreiman @yevgeny-shnaidman  WDYT?

michaellevy101: Last time we agreed that if we don't touch the other disks no warning needed. @avishayt and @romfreiman had strong argument against the warning

michaellevy101: @omertuc anyway the warning is from the assisted-service not the installer. Let's approve this PR and wait for @avishayt response

michaellevy101:
https://bugzilla.redhat.com/show_bug.cgi?id=2086273

gamli75:
We will add a new epic for all the cleaning issues that we have + add a warning message in the UI + need QE to test those tasks

osherdp:
/retitle Bug 2086273: Cleanup disk which is part of a raid and have lvm on it

src/installer/installer.go
```diff
@@ -716,7 +716,7 @@ func (i *installer) updateReadyMasters(nodes *v1.NodeList, readyMasters *[]strin
 
 			host, ok := common.HostMatchByNameOrIPAddress(node, inventoryHostsMap, knownIpAddresses)
 			if !ok {
-				return fmt.Errorf("Node %s is not in inventory hosts", node.Name)
+				return fmt.Errorf("node %s is not in inventory hosts", node.Name)
```
omertuc, line 719: PRs that are backport candidates should strive for minimal changes, so it's best to avoid small irrelevant cleanups such as this

src/installer/installer.go
```diff
@@ -734,36 +734,62 @@ func (i *installer) cleanupInstallDevice() error {
...
 		}
 	}
 
-	if i.ops.IsRaidDevice(i.Device) {
-		i.log.Infof("A raid was detected on the installation device (%s) - cleaning", i.Device)
-		err = i.ops.CleanRaidDevice(i.Device)
+	return i.ops.Wipefs(i.Device)
+}
+
+func (i *installer) cleanupDevice(device string) error {
```
omertuc, line 771: This name is a bit too generic, can you change it to something along the lines of CleanLVM?

src/ops/ops.go
```diff
@@ -568,6 +569,29 @@ func (o *ops) CleanRaidDevice(device string) error {
 	return nil
 }
 
+func (o *ops) GetRaidDevices(deviceName string) ([]string, error) {
```
omertuc, line 572: The distinction between getRaidDevices and GetRaidDevices is not immediately clear from the function names, one has to read the code to tell the difference. The former should be something along the lines of `getAllRaidDevices` and the latter should maybe be `getDeviceRaidDevices`. The latter could really use some documentation to explain its purpose though, because I don't think there's a good name to explain it well enough

omertuc:
Other than superficial naming, looks fine

src/installer/installer.go
michaellevy101: right.

src/installer/installer.go
michaellevy101: This method could be extended to include not only lvm but every nested cleanup (relevant to installation devices and raid devices). Currently, this bug is regarding lvm, but it could also apply to any other nested cleanup.

src/ops/ops.go
michaellevy101: please note that only one of them is public and should used externally but i will add documentation

src/installer/installer.go
omertuc: Such general function could simply call CleanLVM, and ideally wouldn't have all the cleanup code simply mixed in the same method body. 

Personally it confused me because I still don't really understand why you would consider RAID removal to not be "cleanup" but LVM to be "cleanup" .  There's clearly some distinction and one has to happen before the other so I don't think it's fair to give this a general name like "cleanupDevice" when it currently has a very specific cleanup purpose

PR #461, opened 2022-05-11
MGMT-10348: backofflimit in assisted controller is set in a wrong place and does nothing

tsorya:
[MGMT-10348](https://issues.redhat.com//browse/MGMT-10348): backofflimit in assisted controller is set in a wrong place and does nothing


omertuc:
Nice catch, where did you encounter this? it's 6 by default, shouldn't that be way more than enough?

PR #469, opened 2022-05-25
MGMT-10003: save-partlabel and save-partindex coreos-installer args are not honored because the partition is formatted previously

tsorya:
[MGMT-10003](https://issues.redhat.com//browse/MGMT-10003): save-partlabel and save-partindex coreos-installer args are not honored because the partition is formatted previously

If skip-installation-disk-cleanup is set don't cleanup installation disk

### DIFF
--- a/src/installer/installer_test.go
+++ b/src/installer/installer_test.go
@@ -536,6 +536,32 @@ var _ = Describe("installer HostRoleMaster role", func() {
 			ret := installerObj.InstallNode()
 			Expect(ret).Should(BeNil())
 		})
+
+		It("HostRoleMaster role happy flow with skipping disk cleanup", func() {
+			installerObj.Config.SkipInstallationDiskCleanup = true
+			// verify none of cleanup function runs
+			mockops.EXPECT().GetVGByPV(device).Return("vg1", nil).Times(0)
+			mockops.EXPECT().RemoveVG("vg1").Return(nil).Times(0)
+			mockops.EXPECT().IsRaidMember(device).Return(false).Times(0)
+			mockops.EXPECT().Wipefs(device).Return(nil).Times(0)
+			mockops.EXPECT().RemovePV(device).Return(nil).Times(0)
+			updateProgressSuccess([][]string{{string(models.HostStageStartingInstallation), conf.Role},
+				{string(models.HostStageInstalling), conf.Role},
+				{string(models.HostStageWritingImageToDisk)},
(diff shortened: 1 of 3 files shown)


src/installer/installer_test.go
```diff
@@ -593,6 +593,23 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
 		})
+		It("HostRoleMaster role happy flow with skipping disk cleanup", func() {
+			installerObj.Config.SkipInstallationDiskCleanup = true
+			// verify none of cleanup function runs
+			cleanInstallDeviceClean := func() {
+				mockops.EXPECT().GetVGByPV(device).Return("vg1", nil).Times(0)
+				mockops.EXPECT().RemoveVG("vg1").Return(nil).Times(0)
+				mockops.EXPECT().IsRaidMember(device).Return(false).Times(0)
+				mockops.EXPECT().Wipefs(device).Return(nil).Times(0)
+				mockops.EXPECT().RemovePV(device).Return(nil).Times(0)
```
carbonin, line 604: Why put this in a function?

src/installer/installer_test.go
```diff
@@ -593,6 +593,20 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
+			mockops.EXPECT().GetVGByPV(device).Return("vg1", nil).Times(0)
+			mockops.EXPECT().RemoveVG("vg1").Return(nil).Times(0)
+			mockops.EXPECT().IsRaidMember(device).Return(false).Times(0)
+			mockops.EXPECT().Wipefs(device).Return(nil).Times(0)
+			mockops.EXPECT().RemovePV(device).Return(nil).Times(0)
+			updateProgressSuccess([][]string{{string(models.HostStageStartingInstallation), conf.Role}})
+			err := fmt.Errorf("failed to create dir")
+			mockops.EXPECT().Mkdir(InstallDir).Return(err).Times(1)
+			ret := installerObj.InstallNode()
+			Expect(ret).Should(Equal(err))
```
carbonin, line 608: Why are we expecting an error in a happy flow test?

I see this is the case for all if the disk cleanup tests ... that doesn't seem right.
It looks like it was written to assume that a call to `Mkdir` would come right after the cleanup so it was mocked to avoid having to deal with the rest of the function? Why even bother? Just test the cleanup function then.

carbonin (approved):

PR #475, opened 2022-05-26
MGMT-9379: Fix arguments to make sure that HighAvailabilityMode is updated correctly to ensure it is unset for workers.

paul-maidment:
After merging [MGMT-9379](https://issues.redhat.com//browse/MGMT-9379) it was found that the parameter changes that we
applied to HighAvailabilityMode were ineffective in the installer.

After tracking down the root cause, the problem was that the command
line arguments were being parsed twice, once for the main arguments and
once again for the "dry run arguments". On the second parse, the changes
we made to the first parse were overwritten, rendering them ineffective.

This PR fixes that by merging together the code that performs the parse
on main parameters and dry run arguments.

I also removed the static way of parsing arguments and replaced it with
a more "instance based" approach, this is cleaner, causes less confusion
and is also more testable as we can pass an args array to it.

As a result, we can now test the whole method ProcessArgs(..) and I have
added tests to ensure that specific command line parameters achieve the
desired effect.

### DIFF
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -5,6 +5,9 @@ import (
 	"flag"
 	"os"
 
+	"fmt"
+
+	"github.com/kelseyhightower/envconfig"
 	"github.com/openshift/assisted-installer/src/utils"
 	"github.com/openshift/assisted-service/models"
 )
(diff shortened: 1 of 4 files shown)


carbonin:
Do you think it would be good to make `ProcessDryRunArgs` private?

Also maybe make the same change to use a local `FlagSet` in `ProcessArgs` for consistency?

Either way I think you need some more words in the commit an PR description describing exactly why the args were getting overwritten as this behavior around `flag.Parse()` wasn't particularly obvious to me at least.

ori-amizur:
Maybe **_flag.Parse()_** can be called once?

carbonin:
> Maybe flag.Parse() can be called once?

It _should_ only be called once, but was being called twice.
The second call overwrote the defaults we set after the first call even though we didn't register those flags because the `flag` methods use a global `FlagSet` by default.

This is exactly the kind of explanation I'm asking @paul-maidment to add to the PR an commit message :wink:

carbonin:
After looking into this I would prefer we always use a locally defined `FlagSet` object.

It seems like using the global is a really easy way to introduce issues if your argument parsing gets any more complicated than a single function.

src/config/config.go
```diff
@@ -47,35 +50,58 @@ func printHelpAndExit(err error) {
...
-	flag.StringVar(&ret.ClusterID, "cluster-id", "", "The cluster id")
-	flag.StringVar(&ret.InfraEnvID, "infra-env-id", "", "This host infra env id")
-	flag.StringVar(&ret.HostID, "host-id", "", "This host id")
-	flag.StringVar(&ret.Device, "boot-device", "", "The boot device")
-	flag.StringVar(&ret.URL, "url", "", "The BM inventory URL, including a scheme and optionally a port (overrides the host and port arguments")
-	flag.StringVar(&ret.OpenshiftVersion, "openshift-version", "4.4", "Openshift version to install")
-	flag.StringVar(&ret.MCOImage, "mco-image", "", "MCO image to install")
-	flag.BoolVar(&ret.Verbose, "verbose", false, "Increase verbosity, set log level to debug")
-	flag.StringVar(&ret.ControllerImage, "controller-image", "quay.io/ocpmetal/assisted-installer-controller:latest",
+	flagSet := flag.NewFlagSet("flagset", flag.ContinueOnError)
```
carbonin, line 53: Why `ContinueOnError` here?
The [previous implementation used `ExitOnError`](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/flag/flag.go;drc=7791e934c882fd103357448aee0fd577b20013ce;l=1147)
Maybe just use that definition as a reference and do `flagSet := flag.NewFlagSet(os.Args[0], flag.ExitOnError)`?

Also, you don't even need to define the flagSet now as we're only doing this handling once, but I personally prefer less global state.

src/config/config.go
```diff
@@ -47,35 +50,58 @@ func printHelpAndExit(err error) {
...
+			os.Exit(1)
+		}
+	}
 
 	var installerArgs string
-	flag.StringVar(&installerArgs, "installer-args", "", "JSON array of additional coreos-installer arguments")
-	h := flag.Bool("help", false, "Help message")
-	flag.Parse()
+	flagSet.StringVar(&installerArgs, "installer-args", "", "JSON array of additional coreos-installer arguments")
+	h := flagSet.Bool("help", false, "Help message")
```
carbonin, line 104: These need to be set up before the call to `flagSet.Parse`

filanov: +1

src/config/config_test.go
```diff
@@ -13,6 +13,61 @@ func TestConfig(t *testing.T) {
 	RunSpecs(t, "config_test")
 }
 
+var _ = Describe("ProcessArgs", func() {
+
+	It("Should not deserialize installer args when they are not supplied.", func() {
```
carbonin, line 18: This seems like a duplicate of the test in `SetInstallerArgs`.
I think you'd be safe making that method private and removing those tests I don't expect that method to be useful outside of the package.

Same for all the other duplicated tests here.

carbonin (approved):

PR #478, opened 2022-06-02
MGMT-10450: Timeout writing to disk reported from bootstrap node when failing to start bootkube

tsorya:
[MGMT-10450](https://issues.redhat.com//browse/MGMT-10450): Timeout writing to disk reported from bootstrap node when failing to start bootkube

After writing to disk we will update status to waiting for control plane
with info relevant to the state


eranco74 (commented):
/lgtm

PR #479, opened 2022-06-15
AGENT-75: Add assisted installer images into OCP playload for Agent Work

pawanpinjarkar:
Once the images are building successfully in OSBS (and only then), they can be added to the release payload image by setting a label in the Dockerfile:

`LABEL io.openshift.release.operator=true`


pawanpinjarkar:
hold until https://github.com/openshift/ocp-build-data/pull/1582 and https://github.com/openshift/ocp-build-data/pull/1583 are merged

PR #487, opened 2022-06-26
NO-ISSUE: In case must-gather upload failed but logs were gathered already we should retry only upload part

tsorya:
NO-ISSUE: In case must-gather upload failed but logs were gathered already we should retry only upload part

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -62,6 +62,7 @@ var (
 	GeneralWaitInterval      = generalWaitTimeoutInt * time.Second
 	GeneralProgressUpdateInt = 60 * time.Second
 	LogsUploadPeriod         = 5 * time.Minute
+	SummaryLogsPeriod        = 30 * time.Second
 	WaitTimeout              = 70 * time.Minute
 	CompleteTimeout          = 30 * time.Minute
 	DNSAddressRetryInterval  = 20 * time.Second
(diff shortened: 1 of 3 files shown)


omertuc (commented):
/lgtm

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1234,10 +1240,10 @@ func (c *controller) UploadLogs(ctx context.Context, wg *sync.WaitGroup) {
 			if podName != "" {
 				c.log.Infof("Upload final controller and cluster logs before exit")
 				c.ic.ClusterLogProgressReport(progressCtx, c.ClusterID, models.LogsStateRequested)
-				_ = utils.WaitForPredicate(WaitTimeout, LogsUploadPeriod, func() bool {
+				_ = utils.WaitForPredicate(WaitTimeout, SummaryLogsPeriod, func() bool {
 					err := c.uploadSummaryLogs(podName, c.Namespace, controllerLogsSecondsAgo)
```
slaviered, line 1244: why uploading every 30 seconds when the process itself can take more then an hour? plus with regular upload we still wait 5 minutes between one upload to another.

Here there will be no must-gather logs anyhow until the final output is tarred. so I do not see any advantage in this change. I think that the problem in triage is the collecting scripts and not the code here. We see a lot of must-gathers that are not reflected in the ticket

tsorya: we will upload every 30 seconds only in case something gone wrong.
This ticket is fixing the problem when for unknown reason upload didn't work and it will require rerunning must gather logs one more time though those logs already exists and were gathered.

slaviered: Yes, but note that you are retrying the summary log upload all the time (in every case) every 30 seconds. it make sense only if you continuous upload the partial results of must gather like I am attempting at the bootstrap. otherwise it just wait for a very long time for the gather to end

slaviered: OK. I agree

PR #488, opened 2022-06-26
MGMT-10660: Reformat assisted-installer ops code to allow writing unitests

tsorya:
MGMT-10660: Reformat assisted-installer ops code to allow writing unitests


src/ops/execute/execute_test.go
```diff
@@ -66,19 +67,19 @@ var _ = Describe("installerArgs", func() {
...
-		args := installerArgs(ignitionPath, device, []string{})
+		args := utils.InstallerArgs(ignitionPath, device, []string{})
 		expected := []string{"install", "--insecure", "-i", "/tmp/ignition.ign", "/dev/sda"}
 		Expect(args).To(Equal(expected))
 	})
 
 	It("Returns the correct list with extra args", func() {
-		args := installerArgs(ignitionPath, device, []string{"-n", "--append-karg", "nameserver=8.8.8.8"})
+		args := utils.InstallerArgs(ignitionPath, device, []string{"-n", "--append-karg", "nameserver=8.8.8.8"})
 		expected := []string{"install", "--insecure", "-i", "/tmp/ignition.ign", "-n", "--append-karg", "nameserver=8.8.8.8", "/dev/sda"}
```
eranco74, line 83: These tests don't belong here, they have nothing to do with execute.
Consider moving to utils_test/ops_test

src/utils/utils.go
```diff
@@ -310,3 +310,11 @@ func ClusterOperatorConditionsToMonitoredOperatorStatus(conditions []configv1.Cl
 
 	return models.OperatorStatusProgressing, ""
 }
+
+func InstallerArgs(ignitionPath string, device string, extra []string) []string {
```
eranco74, line 314: Unsure why you moved it here :confused:

eranco74 (commented):
Looks good

omertuc:
Can you make sure this doesn't break https://github.com/openshift-assisted/assisted-swarm ?

tsorya:
@omertuc not sure i know how but i don't see how it can break at least, especially if you don't update installer package in swarn

omertuc:
>@omertuc not sure i know how but i don't see how it can break at least, especially if you don't update installer package in swarn

But we do want to occasionally update installer package in swarm, so just make sure it still compiles

tsorya:
@omertuc in case swarm can be broken by installer changes we should have test for it.

tsorya:
@omertuc please share how i should check it, from the code i don't see any issue it can create

omertuc:
> @omertuc in case swarm can be broken by installer changes we should have test for it.

True, but in this case it seems like a refactor that seems like it has a huge potential to break swarm, so I thought it should be manually tested

PR #493, opened 2022-07-06
NO-ISSUE: add stage info in case converged flow is enabled

eranco74:
This should allow us to know if the assisted installer triggred the reboot or not

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -177,16 +177,14 @@ func (i *installer) InstallNode() error {
 }
 
 func (i *installer) finalize() error {
-	//update installation progress
-	i.UpdateHostInstallProgress(models.HostStageRebooting, "")
 	if i.DryRunEnabled {
+		i.UpdateHostInstallProgress(models.HostStageRebooting, "")
 		_, err := i.ops.ExecPrivilegeCommand(nil, "touch", i.Config.FakeRebootMarkerPath)
 		return errors.Wrap(err, "failed to touch fake reboot marker")
 	}
 
 	// in case ironic-agent exists on the host we should stop the assisted-agent service instead of rebooting the node.
 	// the assisted agent service stop will signal the ironic agent that we are done so that IPA can continue with its flow.
-	// regardless, we update the host install progress to `rebooting` since the node will get rebooted shortly after.
 	ironicAgentServiceName := "ironic-agent.service"
 	out, err := i.ops.ExecPrivilegeCommand(nil, "systemctl", "list-units", "--no-legend", ironicAgentServiceName)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


src/installer/installer.go
```diff
@@ -194,12 +194,15 @@ func (i *installer) finalize() error {
 		return err
 	}
 	if strings.Contains(out, ironicAgentServiceName) {
+		//update installation progress
+		i.UpdateHostInstallProgress(models.HostStageRebooting, "converged flow enabled, ironic will reboot the node shortly")
```
omertuc, line 198:
```suggestion
		i.UpdateHostInstallProgress(models.HostStageRebooting, "Converged flow enabled, ironic will reboot the node shortly")
```

omertuc: This is user facing. I don't think most users (or even most people in our team) know what the hell converged flow means, it's not that interesting, so perhaps we should have a simpler message here?

eranco74: I lose the "converged flow enabled," part.

src/installer/installer.go
```diff
@@ -177,9 +177,9 @@ func (i *installer) InstallNode() error {
 }
 
 func (i *installer) finalize() error {
-	//update installation progress
-	i.UpdateHostInstallProgress(models.HostStageRebooting, "")
 	if i.DryRunEnabled {
+		//update installation progress
```
omertuc, line 181:
```suggestion
```

src/installer/installer.go
```diff
@@ -194,12 +194,15 @@ func (i *installer) finalize() error {
 		return err
 	}
 	if strings.Contains(out, ironicAgentServiceName) {
+		//update installation progress
```
omertuc, line 197:
```suggestion
```

src/installer/installer.go
```diff
@@ -194,12 +194,15 @@ func (i *installer) finalize() error {
...
+		i.UpdateHostInstallProgress(models.HostStageRebooting, "Ironic will reboot the node shortly")
 		// in case the ironic-agent exists the installer should stop the assisted agent.service
 		if err = i.ops.SystemctlAction("stop", "agent.service"); err != nil {
 			return err
 		}
 	} else {
 		//reboot
```
omertuc, line 204:
```suggestion
```

omertuc:
Can you elaborate on why this is needed / in which situations it helps?

omertuc:
The old:
```
// regardless, we update the host install progress to `rebooting` since the node will get rebooted shortly after.
```

comment can probably removed now, as the stage message makes it obviously apparent

eranco74:
> Can you elaborate on why this is needed / in which situations it helps?

This helps in cases like [this](https://bugzilla.redhat.com/show_bug.cgi?id=2104117) where the assisted-service thinks the installation is done but ironic thinks that the assisted-agent didn't complete the installation.
Since this progress update is a bit misleading (someone else will reboot the node) adding this info should make it easier to understand that the assisted0installer didn't reboot it

PR #497, opened 2022-07-10
MGMT-11100: Add router get router status logic in case of sno failure

tsorya:
MGMT-11100: Adding get router status logic in case of sno failure.
    Currently we have known issue in sno flow where router healthcheck fails
    from pods network. This pr add verification that router is reachable
    from host network and enables router access log.
    This change is temporary till we will not find an issue we encountered

### DIFF
--- a/deploy/assisted-installer-controller/assisted-installer-controller-role.yaml
+++ b/deploy/assisted-installer-controller/assisted-installer-controller-role.yaml
@@ -99,6 +99,7 @@ rules:
       - operator.openshift.io
     resources:
       - etcds
+      - ingresscontrollers
     verbs:
       - patch
   - apiGroups:
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -2,6 +2,7 @@ package assisted_installer_controller
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -9,6 +10,7 @@ import (
 	"io/ioutil"
 	"math"
 	"net"
+	"net/http"
 	"os"
 	"path"
 	"strings"
@@ -70,6 +72,7 @@ var (
 	FetchRetryInterval       = 10 * time.Second
 	LongWaitTimeout          = 10 * time.Hour
 	CVOMaxTimeout            = 3 * time.Hour
+	dnsValidationTimeout     = 10 * time.Second
 )
 
 // assisted installer controller is added to control installation process after  bootstrap pivot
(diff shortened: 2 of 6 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1075,6 +1078,67 @@ func (c controller) logClusterOperatorsStatus() {
 	}
 }
 
+// This is temporary function that should validate router state only in case of failure
+// It will patch router to add access logs
+// It will run http router health check to see if router is healthy on host network
+func (c controller) logRouterStatus() {
+	if (!c.Status.HasError() && !c.Status.HasOperatorError()) || c.HighAvailabilityMode != models.ClusterHighAvailabilityModeNone {
+		return
```
eranco74, line 1086: add log

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1075,6 +1078,67 @@ func (c controller) logClusterOperatorsStatus() {
 	}
 }
 
+// This is temporary function that should validate router state only in case of failure
+// It will patch router to add access logs
+// It will run http router health check to see if router is healthy on host network
+func (c controller) logRouterStatus() {
+	if (!c.Status.HasError() && !c.Status.HasOperatorError()) || c.HighAvailabilityMode != models.ClusterHighAvailabilityModeNone {
```
eranco74: Why do we need to check both?
(!c.Status.HasError() && !c.Status.HasOperatorError())

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1075,6 +1078,67 @@ func (c controller) logClusterOperatorsStatus() {
...
+
+		url := fmt.Sprintf("http://%s/_______internal_router_healthz", localIp)
+		_, err = client.Get(url)
+		if err != nil {
+			c.log.WithError(err).Warning("Failed to reach internal router health")
+		} else {
+			c.log.Infof("route internal health status %s", r.Status)
+		}
+
+		if !patched {
```
eranco74, line 1128: Why have the patch in the predicate? why not do it once in case the predicate failed

tsorya: cause i want to check status before the check and in that case i need to move check from predicate.

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: I don't think it is actually needed. Don't want anybody to be confused by it

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: so it doesn't actually matter. Don't see reason to move it to blocking part of the function

filanov:
Can't understand anything from the commit message

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1075,6 +1078,67 @@ func (c controller) logClusterOperatorsStatus() {
 	}
 }
 
+// This is temporary function that should validate router state only in case of failure
```
filanov, line 1081: temporary until?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1075,6 +1078,67 @@ func (c controller) logClusterOperatorsStatus() {
...
+		consoleUrl = fmt.Sprintf("https://console-openshift-console.apps.%s.%s/health", cl.Name, cl.BaseDNSDomain)
+		r, err := client.Get(consoleUrl)
+		if err != nil {
+			c.log.WithError(err).Warning("Failed to reach console")
+		} else {
+			c.log.Infof("console url status %s", r.Status)
+		}
+
+		url := fmt.Sprintf("http://%s/_______internal_router_healthz", localIp)
+		_, err = client.Get(url)
```
filanov, line 1121: Is this call just for logs?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1075,6 +1078,67 @@ func (c controller) logClusterOperatorsStatus() {
...
+	}
+	client := &http.Client{Transport: tr, Timeout: dnsValidationTimeout}
+	localIp, _ := utils.GetOutboundIP()
+	if localIp == "" {
+		localIp = "localhost"
+	}
+
+	var consoleUrl string
+	// run every 30 seconds router check and log the result
+	go utils.WaitForPredicate(WaitTimeout, SummaryLogsPeriod, func() bool {
```
filanov, line 1103: Does it return true at some point?

src/k8s_client/k8s_client.go
```diff
@@ -226,6 +227,17 @@ func (c *k8sClient) UnPatchEtcd() error {
 	return nil
 }
 
+func (c *k8sClient) EnableRouterAccessLogs() error {
+	c.log.Info("Enabling router logs")
+	data := []byte(`{"spec":{"logging":{"access":{"destination":{"type":"Container"}}}}}`)
+	result, err := c.ocClient.OperatorV1().IngressControllers("openshift-ingress-operator").Patch(context.Background(), "default", types.MergePatchType, data, metav1.PatchOptions{})
```
filanov, line 233: please break long lines

src/utils/utils.go
```diff
@@ -310,3 +310,16 @@ func ClusterOperatorConditionsToMonitoredOperatorStatus(conditions []configv1.Cl
...
+
+// Get preferred outbound ip of this machine
+func GetOutboundIP() (string, error) {
+	conn, err := net.Dial("udp", "8.8.8.8:80")
+	if err != nil {
+		return "", err
+	}
+	defer conn.Close()
+
+	localAddr := conn.LocalAddr().(*net.UDPAddr)
```
filanov, line 322: Can it be nil?

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: nope, it will exit when controller will exit

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: this call is just for status

deploy/assisted-installer-controller/assisted-installer-controller-role.yaml
```diff
@@ -176,3 +177,4 @@ rules:
       - securitycontextconstraints
     verbs:
       - use
+
```
eranco74, line 180: remove

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1075,6 +1078,71 @@ func (c controller) logClusterOperatorsStatus() {
...
+
+			consoleUrl = fmt.Sprintf("https://console-openshift-console.apps.%s.%s/health", cl.Name, cl.BaseDNSDomain)
+			r, err := client.Get(consoleUrl)
+			if err != nil {
+				c.log.WithError(err).Warning("Failed to reach console")
+			} else {
+				c.log.Infof("console url status %s", r.Status)
+			}
+
+			url := fmt.Sprintf("http://%s/_______internal_router_healthz", localIp)
```
eranco74, line 1123: Is this right?
All the _____?
Why not user /healthz ?

eranco74 (commented):
Just make sure to revert it once the issue is resolved

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: this is what we were asked to set. Yep it is doing what it should

PR #506, opened 2022-07-18
MGMT-11106: Writing controller logs to file in order to be able to send

tsorya:
[MGMT-11106](https://issues.redhat.com//browse/MGMT-11106): Writing controller logs to file in order to be able to send them in case of kube-api failure

Adding file handler to assisted-controller logger. Upload logs will send logs from log file in case we failed to reach kube-api to read those logs. In case something is wrong with file and kube-api we will send error message.
File will be created in /tmp in order not to handle user premissions in k8s

### DIFF
--- a/src/common/common.go
+++ b/src/common/common.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


src/common/common.go
```diff
@@ -93,14 +95,28 @@ func GetPodInStatus(k8Client k8s_client.K8SClient, podNamePrefix string, namespa
...
+		if err != nil {
+			log.WithError(err).Error("Failed to get logs from kube-api, reading from file")
+		}
 	}
+	if podName == "" || err != nil {
+		log.Infof("Reading logs from file")
+		logs, errF := ioutil.ReadFile(ControllerLogFIle)
+		if errF != nil {
+			log.WithError(errF).Warnf("Failed to read %s", ControllerLogFIle)
+			logs = []byte(fmt.Sprintf("Failed to get logs from kube-api and from file. Read file err %e, kube-api err %e", errF, err))
```
omertuc, line 115: Maybe let's return instead? Uploading this line can only damage existing logs...

Although it would probably never happen so no big deal

omertuc: Feel free to unhold after reading this

tsorya: To be sure we will get something if file read will not work for any unknown strange reason

nmagnezi (approved):
/lgtm

Issue #513, opened 2022-08-02
assisted-installer-controller Job does not apply Additional Root CA Trust Bundle


kenmoini:
**ISSUE:**

When installing an OpenShift cluster that has a combination of:

- Outbound Proxy
- SSL Re-encryption with another trusted root CA (Cisco/McAfee/Squid proxy SSL MitM basically)

...the installation halts with the bootstrap node at `reached installation stage Waiting for controller: waiting for controller pod ready event` and the other two control plane nodes at a `Joined` status.

**CAUSE:**

Once the assisted-installer-controller Job is created in the assisted-installer namespace, it correctly passes in the Outbound Proxy but the Additional Trust Bundles are not added as a mounted volume.  If the Outbound Proxy performs SSL re-encryption then the Pod will fail with the following, even if the CA that is performing the re-encryption is applied as an additionalTrustBundle certificate:

```
time="2022-08-01T23:39:07Z" level=info msg="Start running Assisted-Controller. Configuration is:\n struct ControllerConfig {\n\tClusterID: \"6e08c390-3b8c-4009-9a41-605c7ff40f25\",\n\tURL: \"https://api.openshift.com\",\n\tPullSecretToken: <SECRET>,\n\tSkipCertVerification: false,\n\tCACertPath: \"\",\n\tNamespace: \"assisted-installer\",\n\tOpenshiftVersion: \"4.10.18\",\n\tHighAvailabilityMode: \"Full\",\n\tWaitForClusterVersion: true,\n\tMustGatherImage: \"quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:03a06499c3efae948535eb61c340efb8511d3ac45db1ca9fccfe5515e49a70ac\",\n\tDryRunEnabled: false,\n\tDryFakeRebootMarkerPath: \"\",\n\tDryRunClusterHostsPath: \"\",\n\tParsedClusterHosts: config.DryClusterHosts(nil),\n}"
W0801 23:39:07.295522       1 client_config.go:617] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0801 23:39:09.236233       1 request.go:601] Waited for 1.045203989s due to client-side throttling, not priority and fairness, request: GET:https://172.30.0.1:443/apis/node.k8s.io/v1?timeout=32s
[...]

PR #514, opened 2022-08-02
Bug Fix: Add additionalTrustBundle defined CA Certificates from RHCOS CA Trust as a default

kenmoini:
Addresses Issue #513 

When additionalTrustBundle Root CA certificates are added to the RHCOS trusted store the assisted-installer-controller Job Pod needs to have them mounted - add the default `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` for CACertPath.

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -84,7 +84,7 @@ type ControllerConfig struct {
 	URL                     string `envconfig:"INVENTORY_URL" required:"true"`
 	PullSecretToken         string `envconfig:"PULL_SECRET_TOKEN" required:"true" secret:"true"`
 	SkipCertVerification    bool   `envconfig:"SKIP_CERT_VERIFICATION" required:"false" default:"false"`
-	CACertPath              string `envconfig:"CA_CERT_PATH" required:"false" default:"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"`
+	CACertPath              string `envconfig:"CA_CERT_PATH" required:"false" default:""`
 	Namespace               string `envconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
 	OpenshiftVersion        string `envconfig:"OPENSHIFT_VERSION" required:"true"`
 	HighAvailabilityMode    string `envconfig:"HIGH_AVAILABILITY_MODE" required:"false" default:"Full"`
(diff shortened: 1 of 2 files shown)


mkowalski (changes requested):
Some test please? What if `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` does not exist and we try to mount it?

kenmoini:
> Some test please? What if `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` does not exist and we try to mount it?

Is there a configuration in which the assisted-installer-controller is not run in a Linux container or on a Linux host?

The `ca-certificates` package is a standard and required component of both the base images and RHCOS.  Curious if supportability of something like Mac OS X is targeted for out-of-cluster binary execution?

Anywho, happy to add a test or two for a discount-double-check...I can add a file/path check in `func NewExecutor` that resets that config variable to `""` if not found?  This would also provide a check and graceful failback for when a user specifies a different path that doesn't exist. 

Is there another test you'd like added otherwise?

mkowalski:
By no means I want to overengineer anything here especially that I see the use case as valid. I'm just worried about "what if", e.g. we have already had a case in the past where RHCOS didn't have the path you added - https://access.redhat.com/solutions/5582001.

> Curious if supportability of something like Mac OS X is targeted for out-of-cluster binary execution?
I wouldn't worry about this at all now. This is about running on the node that's part of OCP cluster, so as long as we install OCP on RHCOS this is the only thing we should care about. 

>I can add a file/path check in func NewExecutor that resets that config variable to "" if not found?

Yes please, something like this would make it already safer. I only want to make sure that we gracefully handle scenario when the file does not have the content you would expect it to have

eranco74:
@kenmoini please remove the default from the controllerConfig struct (just so we will have a single source of truth).

mkowalski:
/hold

Until I see something that ensures existence of `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`, I don't feel like this should land. RHCOS has this file **today**. Is it going to have it in the next release? Does FCOS have this file? Is it going to have it in the next release?

kenmoini:
> /hold
> 
> Until I see something that ensures existence of `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`, I don't feel like this should land. RHCOS has this file **today**. Is it going to have it in the next release? Does FCOS have this file? Is it going to have it in the next release?

Ok, where would you think the best place for the host filesystem check would be?  Leveraging *installer.ops.ExecPrivilegeCommand similar to how the finalize() func does?

tsorya:
will it be enough to validate this path exists on discovery iso?
actually i think we can always mount all /etc/pki as we do in agent and installer

tsorya:
@kenmoini i think we want to mount all the folder https://github.com/openshift/assisted-installer/pull/551

kenmoini:
That should work too - thanks gang!

PR #516, opened 2022-08-15
MGMT-11170: Log thread ID and caller information for each log entry.

paul-maidment:
Every time an item is logged, we should show the thread ID and any caller information to aid debugging.


src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -51,7 +52,10 @@ func DryRebootComplete() bool {
 }
 
 func main() {
+	threadId := syscall.Gettid()
```
omertuc, line 55: This seems pretty useless:
1. OS threads don't mean much from a go perspective, a thread may be executing several different go routines and go routines often pause/resume from and to different OS threads.
2. Even if we switched to some go routine ID instead of thread ID, what would that ID actually mean? How can I associate that ID with a particular go routine just from looking at the logs?

PR #517, opened 2022-08-18
MGMT-10085: Get real filepath of RAID device

javipolo:
When having LVM on top of software raid, the name of the device reported by mdadm is a symlink to the real block device in /dev, causing problems identifying which LVM volume groups need to be removed before cleaning the RAID

This should fix that issue

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -510,7 +510,12 @@ func (o *ops) getRaidDevices2Members() (map[string][]string, error) {
 		}
 
 		fields := strings.Fields(lines[i])
-		raidDeviceName := fields[1]
+		// In case of symlink, get real file path
+		raidDeviceName, err := filepath.EvalSymlinks(fields[1])
+		if err != nil {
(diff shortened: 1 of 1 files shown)


src/ops/ops.go
```diff
@@ -510,7 +510,12 @@ func (o *ops) getRaidDevices2Members() (map[string][]string, error) {
 		}
 
 		fields := strings.Fields(lines[i])
-		raidDeviceName := fields[1]
+		// In case of symlink, get real file path
+		raidDeviceName, err := filepath.EvalSymlinks(fields[1])
+		if err != nil {
+			return nil, err
```
omertuc, line 516: There should be much more of this in the file but let's at-least make sure new code has it - errors should be given context, otherwise it's hard to follow where they come from

```suggestion
			return nil, errors.Wrapf(err, "failed to evaluate raid symlink")
```

javipolo: sure!
Just added your suggestion :)

PR #518, opened 2022-08-19
MGMT-11695: Remove all LVM volume groups for installer disk

javipolo:
If you have more than one phisical volume assigned to a volume group in disk that we need to wipe, only the first volume group is deleted during cleanup, this causes problems later because there are still existing volume groups in use

This commit addresses that by ensuring that all volume groups are considered.

/cc @omertuc
/cc @paul-maidment

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -39,7 +39,7 @@ type Ops interface {
 	ExtractFromIgnition(ignitionPath string, fileToExtract string) error
 	SystemctlAction(action string, args ...string) error
 	PrepareController() error
-	GetVGByPV(pvName string) (string, error)
+	GetVolumeGroupsByDisk(diskName string) ([]string, error)
 	RemoveVG(vgName string) error
 	RemoveLV(lvName, vgName string) error
 	RemovePV(pvName string) error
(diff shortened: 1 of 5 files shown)


src/ops/ops.go
```diff
@@ -368,11 +369,13 @@ func (o *ops) renderDeploymentFiles(srcTemplate string, params map[string]interf
 	return nil
 }
 
-func (o *ops) GetVGByPV(pvName string) (string, error) {
+func (o *ops) GetVolumeGroupsByDisk(diskName string) ([]string, error) {
+	var VGs []string
```
tsorya, line 373: lowcase?

src/ops/ops.go
```diff
@@ -368,11 +369,13 @@ func (o *ops) renderDeploymentFiles(srcTemplate string, params map[string]interf
 	return nil
 }
 
-func (o *ops) GetVGByPV(pvName string) (string, error) {
+func (o *ops) GetVolumeGroupsByDisk(diskName string) ([]string, error) {
```
tsorya, line 372: maybe some test? Now we have option to create unitest for ops, i think we should add small test for this logic

src/ops/ops.go
paul-maidment: I'm not sure it should be? every other method in the interface (many of these only used internally also) are Uppercase.

See 

https://github.com/openshift/assisted-installer/blob/master/src/ops/ops.go#L34-L62

tsorya: i mean "var VGs []string" should be "var vgs []string"

paul-maidment: agreed, and fixed

osherdp:
/hold
waiting for reviews and tests to finish

osherdp:
> /test all

no need to ``/test all`` after commits once the PR is not in draft :smiley:

eranco74 (commented):
Looks good just nit

osherdp (approved):

PR #519, opened 2022-08-21
Bug 2119308: save-partlabel and save-partindex coreos-installer args …

tsorya:
…are not honored because the partition is formatted previously


tsorya:
original pr https://github.com/openshift/assisted-installer/pull/469

filanov (approved):

PR #526, opened 2022-08-22
MGMT-11565: Add handleOLMEarlySetupBug when waiting for CSV

omertuc:
2 commits, 1 general cleanup and 1 actual feature

- Waiting for ClusterServiceVersion for operators now actually waits for
a real CSV and not just CSV name in the subscription status, because
the latter doesn't seem to mean much

- When we wait for CSV we also check if the subscription is manual, if
it is, we change it to automatic and delete its install plans so they
they get re-created as automatic


omertuc:
Superseded by #531

PR #531, opened 2022-08-25
MGMT-11565: Add handleOLMEarlySetupBug when waiting for CSV

tsorya:
[MGMT-11565](https://issues.redhat.com//browse/MGMT-11565): Add handleOLMEarlySetupBug when waiting for CSV
- Waiting for ClusterServiceVersion for operators now actually waits for
  a real CSV and not just CSV name in the subscription status, because
  the latter doesn't seem to mean much

- When we wait for CSV and it doesn't exist yet, we also check if there are failed olm jobs
  if they are, we delete those jobs and delete install plan.
  It will force olm to recreate failed jobs and restart installation

### DIFF
--- a/src/assisted_installer_controller/operator_handler.go
+++ b/src/assisted_installer_controller/operator_handler.go
@@ -4,6 +4,13 @@ import (
 	"context"
 	"time"
 
+	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/sirupsen/logrus"
+	batchV1 "k8s.io/api/batch/v1"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+
 	"github.com/openshift/assisted-installer/src/k8s_client"
 	"github.com/openshift/assisted-installer/src/utils"
 	"github.com/openshift/assisted-service/models"
@@ -12,6 +19,7 @@ import (
 const (
 	cvoOperatorName    = "cvo"
 	clusterVersionName = "version"
+	olmNamespace       = "openshift-marketplace"
 )
 
 type OperatorHandler interface {
@@ -96,10 +104,11 @@ func (handler ClusterOperatorHandler) OnChange(_ models.OperatorStatus) bool { r
 type ClusterVersionHandler struct {
 	kc    k8s_client.K8SClient
 	timer *time.Timer
+	log   *logrus.Logger
 }
 
-func NewClusterVersionHandler(kc k8s_client.K8SClient, timer *time.Timer) *ClusterVersionHandler {
-	return &ClusterVersionHandler{kc: kc, timer: timer}
+func NewClusterVersionHandler(log *logrus.Logger, kc k8s_client.K8SClient, timer *time.Timer) *ClusterVersionHandler {
+	return &ClusterVersionHandler{kc: kc, timer: timer, log: log}
 }
 
 func (handler ClusterVersionHandler) GetName() string { return cvoOperatorName }
@@ -132,10 +141,11 @@ type ClusterServiceVersionHandler struct {
 	operator *models.MonitoredOperator
 	status   *ControllerStatus
 	retries  int
+	log      *logrus.Logger
 }
 
-func NewClusterServiceVersionHandler(kc k8s_client.K8SClient, operator *models.MonitoredOperator, status *ControllerStatus) *ClusterServiceVersionHandler {
-	return &ClusterServiceVersionHandler{kc: kc, operator: operator, status: status, retries: 0}
+func NewClusterServiceVersionHandler(log *logrus.Logger, kc k8s_client.K8SClient, operator *models.MonitoredOperator, status *ControllerStatus) *ClusterServiceVersionHandler {
+	return &ClusterServiceVersionHandler{kc: kc, operator: operator, status: status, retries: 0, log: log}
 }
 
 func (handler ClusterServiceVersionHandler) GetName() string { return handler.operator.Name }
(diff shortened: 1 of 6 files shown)


gamli75:
cc @omertuc

tsorya:
cnv is failing on test issue

omertuc:
Which OLM jobs are failing? The service is supposed to set subscription to manual, I assume that should prevent any jobs from even appearing

tsorya:
Creating subscription will generate the job, it doesn't matter if it is manual or automatic. 
What is causing the failure is actually this failing job that exists per each future csv.
I left automatic subscription as it allows to pass installation in 50% of the cases and changed controller logic to remove jobs that failed on subscription creation, then remove failed install plans only then jobs are rerun and everything starts working

omertuc:
> Creating subscription will generate the job, it doesn't matter if it is manual or automatic. What is causing the failure is actually this failing job that exists per each future csv. I left automatic subscription as it allows to pass installation in 50% of the cases and changed controller logic to remove jobs that failed on subscription creation, then remove failed install plans only then jobs are rerun and everything starts working

Huh. I see. Then there's no point in even doing manual service subs then. I didn't encounter those failed jobs in my testing, must have gotten lucky

src/assisted_installer_controller/operator_handler.go
```diff
@@ -145,14 +155,97 @@ func (handler ClusterServiceVersionHandler) IsInitialized() bool {
...
+// them.
+func (handler ClusterServiceVersionHandler) handleOLMEarlySetupBug() error {
+	handler.log.Infof("Check if there are failed olm jobs and delete them in case they exists")
+	err := handler.deleteFailedOlmJobs()
+	if err != nil {
+		handler.log.WithError(err).Warnf("Failed to delete olm jobs")
+		return err
+	}
+	// OLM for some reason doesn't reconcile the subscription becoming
+	// automatic, so we have to delete all the install plans to force OLM
```
omertuc, line 192: This comment no longer seems relevant

tsorya:
yeah, no need for service changes. Currenlty it looks working, we ran today 6 tests and all passed. I assume it should be good

src/assisted_installer_controller/operator_handler.go
```diff
@@ -145,14 +155,97 @@ func (handler ClusterServiceVersionHandler) IsInitialized() bool {
...
 		return false
 	}
 
 	return true
 }
 
+// handleOLMEarlySetupBug cleans failed olm jobs as they will rerun by olm and
+// deletes install plans to force olm re-conciliation and job recreation.
+// Assisted Controller to take care of it once the cluster is up. The service
+// does this because subscriptions created too early in the installation
```
omertuc, line 181: This comment is outdated as well

tsorya: why? i updated it.

src/assisted_installer_controller/operator_handler.go
tsorya: it is relevant to , it doesn't reconcile it :(

omertuc: But now we delete the job because we want it to recreate, not because we changed it to automatic as the current comment implies

src/assisted_installer_controller/operator_handler.go
omertuc: Because it talks about the service doing things which it no longer does

tsorya: fixed

src/assisted_installer_controller/operator_handler.go
tsorya: fixed

src/k8s_client/k8s_client.go
```diff
@@ -595,11 +614,55 @@ func (c *k8sClient) CreateEvent(namespace, name, message, component string) (*v1
...
+			if owner.Name == subscription.Name {
+				installPlansOwnedBySubscription = append(installPlansOwnedBySubscription, plan)
+			}
+		}
+	}
+
+	return installPlansOwnedBySubscription, nil
+}
+
+func (c *k8sClient) MakeSubscriptionAutomatic(subscription types.NamespacedName) error {
```
omertuc, line 649: This function is unused now, so it can be removed I think

src/k8s_client/k8s_client.go
```diff
@@ -63,6 +64,10 @@ type K8SClient interface {
 	GetPods(namespace string, labelMatch map[string]string, fieldSelector string) ([]v1.Pod, error)
 	GetCSV(namespace string, name string) (*olmv1alpha1.ClusterServiceVersion, error)
 	GetCSVFromSubscription(namespace string, name string) (string, error)
+	GetSubscription(subscription types.NamespacedName) (*olmv1alpha1.Subscription, error)
+	GetAllInstallPlansOfSubscription(subscription types.NamespacedName) ([]olmv1alpha1.InstallPlan, error)
+	MakeSubscriptionAutomatic(subscription types.NamespacedName) error
```
omertuc, line 69: Don't forget to also remove the interface / re-generate mocks

tsorya: missed it

osherdp:
hmm ``OpenShift Virtualization is supported only for x86_64 CPU architecture``
Maybe it's related to the multi-arch change

osherdp:
> OpenShift Virtualization is supported

meaning this condition: https://github.com/openshift/assisted-service/blob/5c9745d2283726f48d2d25c859b2ec5730d863f9/internal/operators/cnv/cnv_operator.go#L79
needs to include ``MultiCPUArchitecture`` as a supported option
@tsorya ^

osherdp:
/override ci/prow/edge-e2e-metal-assisted-cnv
I opened https://issues.redhat.com/browse/MGMT-11826 to handle this problem

PR #532, opened 2022-08-28
Bug 2122012: assisted-installer isn't sending the host logs in case th…e installation was successful and cacert was  added

tsorya:
Bug 2122012: assisted-installer isn't sending the host logs in case th…e installation was successful and cacert was  added


tsorya:
https://github.com/openshift/assisted-installer/pull/495

filanov:
please link the original PR

filanov (approved):

PR #536, opened 2022-09-05
MGMT-11348: Changes for upstream multi-arch support

ori-amizur:
Let Dockerfile and Makefile respond to TARGETPLATFORM setting which is
set by docker buildx when building multi-arch images.

/cc @tsorya

### DIFF
--- a/Dockerfile.assisted-installer-controller
+++ b/Dockerfile.assisted-installer-controller
@@ -1,6 +1,10 @@
-FROM quay.io/openshift/origin-cli:4.11.0 AS cli
+FROM quay.io/openshift/origin-cli-artifacts:latest as cli-artifacts
+ARG TARGETPLATFORM
+RUN mkdir -p /usr/share/openshift/assisted
+RUN case $TARGETPLATFORM in "") dir=linux_amd64;; *) dir=`echo $TARGETPLATFORM | sed 's@/@_@'` ;; esac ;  ln /usr/share/openshift/${dir}/oc /usr/share/openshift/assisted
 
 FROM registry.ci.openshift.org/openshift/release:golang-1.17 AS builder
+ARG TARGETPLATFORM
 ENV GOFLAGS=-mod=mod
 WORKDIR /go/src/github.com/openshift/assisted-installer
 
@@ -11,12 +15,12 @@ COPY go.sum go.sum
 RUN go mod download
 
 COPY . .
-RUN make controller
+RUN TARGETPLATFORM=$TARGETPLATFORM make controller
 
 FROM quay.io/centos/centos:stream8
 
 RUN yum -y install make gcc unzip wget curl rsync && yum clean all
+COPY --from=cli-artifacts /usr/share/openshift/assisted/oc /usr/bin/oc
 COPY --from=builder /go/src/github.com/openshift/assisted-installer/build/assisted-installer-controller /usr/bin/assisted-installer-controller
-COPY --from=cli /usr/bin/oc /usr/bin/oc
 
 ENTRYPOINT ["/usr/bin/assisted-installer-controller"]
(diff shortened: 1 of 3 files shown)


tsorya:
looks good to me, lets one more person look at and lgtm

Dockerfile.assisted-installer-controller
```diff
@@ -11,12 +10,14 @@ COPY go.sum go.sum
...
 COPY . .
-RUN make controller
+RUN TARGETPLATFORM=$TARGETPLATFORM make controller
 
 FROM quay.io/centos/centos:stream8
+ARG TARGETPLATFORM
 
 RUN yum -y install make gcc unzip wget curl rsync && yum clean all
+RUN case $TARGETPLATFORM in linux/amd64 | "") arch=x86_64 ;; linux/arm64) arch=aarch64 ;; *) echo Invalid platform $TARGETPLATFORM && exit 1 ;; esac ; \
+   curl -k https://mirror.openshift.com/pub/openshift-v4/$arch/clients/ocp/4.11.3/openshift-client-linux.tar.gz | tar xfz - -C /usr/bin oc && chmod +x /usr/bin/oc
```
adriengentil, line 20: don't we have multi arch images for oc cli? Otherwise, we will end up with the same issues described in https://github.com/openshift/assisted-service/pull/4286

ori-amizur: Good point.
Unfortunately, that repo **origin-cli** is not multi arch.  Didn't find another that is multi-arch.

tsorya: Can you ask in multi-arch channel? I think we need to open ticket for them

romfreiman: +1 to @tsorya. Lets not reinvent the wheel here - lets ask on slack

tsorya: @romfreiman but i think we can push it as is and open relevant tickets wdyt?

romfreiman: i think that asking takes exactly 10 seconds

ori-amizur: BTW, the problem in the PR https://github.com/openshift/assisted-service/pull/4286 was that we used dev-preview

adriengentil, line 20:
```suggestion
   curl --retry 5 --silent https://mirror.openshift.com/pub/openshift-v4/$arch/clients/ocp/4.11.3/openshift-client-linux.tar.gz | tar xfz - -C /usr/bin oc && chmod +x /usr/bin/oc
```
removed `-k` because I'm not sure we want insecure connection.

adriengentil (changes requested):
/cc @osherdp 
FYI, `quay.io/openshift/origin-cli` is not available as a multi-arch image, I don't see any alternatives to what @ori-amizur proposed

Dockerfile.assisted-installer-controller
ori-amizur: Changed to use [quay.io/openshift/origin-cli-artifacts](http://quay.io/openshift/origin-cli-artifacts)  according to suggestion from https://coreos.slack.com/archives/CFFJUNP6C/p1662470502219889?thread_ts=1662449204.007179&cid=CFFJUNP6C

tsorya:
I assume that we should run test of it no?

Dockerfile.assisted-installer-controller
```diff
@@ -11,12 +15,12 @@ COPY go.sum go.sum
 RUN go mod download
 
 COPY . .
-RUN make controller
+RUN TARGETPLATFORM=$TARGETPLATFORM make controller
 
 FROM quay.io/centos/centos:stream8
 
 RUN yum -y install make gcc unzip wget curl rsync && yum clean all
+COPY --from=cli-artifacts /usr/share/openshift/assisted/oc /usr/bin/oc
```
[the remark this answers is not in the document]
ori-amizur: Please look at line 4 of the Dockerfile.  It creates hard link from the right oc.
Here we take the contents of the file.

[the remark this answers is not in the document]
tsorya: sorry, missed it

PR #540, opened 2022-09-13
MGMT-12038: Controller should upload logs from file on summary logs

tsorya:


src/common/common.go
```diff
@@ -109,29 +102,41 @@ func UploadPodLogs(kc k8s_client.K8SClient, ic inventory_client.InventoryClient,
 	}
 	if podName == "" || err != nil {
 		log.Infof("Reading logs from file")
-		logs, errF := ioutil.ReadFile(ControllerLogFIle)
+		logs, errF := ioutil.ReadFile(ControllerLogFile)
```
eranco74, line 105: Move the read logs from file to another function and call that function when you want to get the controller logs.
And rename this function to `GetPodLogs` (it's getting any pod logs not just the controller)

eranco74 (commented):
/lgtm

PR #542, opened 2022-09-14
MGMT-12057: Remove enablement of router access logs in controller, not needed anymore

tsorya:
[MGMT-12057](https://issues.redhat.com//browse/MGMT-12057): Remove enablement of router access logs in controller, not needed anymore


omertuc:
Why is it not needed anymore? Was the issue it was looking for resolved? How was it resolved? Any links / references?

tsorya:
We understood that issue is not in router and there is no point to do it. 
It is enough to curl router status
https://bugzilla.redhat.com/show_bug.cgi?id=2097041
see my comment

PR #545, opened 2022-09-15
MGMT-12055: Assisted controller should log if one of the nodes changed it's ip after reboot

tsorya:
[MGMT-12055](https://issues.redhat.com//browse/MGMT-12055): Assisted controller should log if one of the nodes changed it's ip after reboot

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -241,6 +241,7 @@ func (c *controller) waitAndUpdateNodesStatus() bool {
 			log.Warnf("Node %s is not in inventory hosts", strings.ToLower(node.Name))
 			continue
 		}
+		common.LogIfHostIpChanged(c.log, node, knownIpAddresses)
 
 		if host.Host.Progress.CurrentStage == models.HostStageConfiguring {
 			log.Infof("Found new joined node %s with inventory id %s, kubernetes id %s, updating its status to %s",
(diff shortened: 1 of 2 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -241,6 +241,7 @@ func (c *controller) waitAndUpdateNodesStatus() bool {
 			log.Warnf("Node %s is not in inventory hosts", strings.ToLower(node.Name))
 			continue
 		}
+		common.LogIfHostIpWasChanged(c.log, node, knownIpAddresses)
```
eranco74, line 244:
```suggestion
		common.LogIfHostIpChanged(c.log, node, knownIpAddresses)
```

src/common/common.go
```diff
@@ -168,6 +168,15 @@ func BuildHostsMapIPAddressBased(inventoryHostsMap map[string]inventory_client.H
 	return knownIpAddresses
 }
 
+func LogIfHostIpWasChanged(log logrus.FieldLogger, node v1.Node, IPAddressMap map[string]inventory_client.HostData) {
```
eranco74, line 171:
```suggestion
func LogIfHostIpChanged(log logrus.FieldLogger, node v1.Node, IPAddressMap map[string]inventory_client.HostData) {
```

src/common/common.go
```diff
@@ -168,6 +168,15 @@ func BuildHostsMapIPAddressBased(inventoryHostsMap map[string]inventory_client.H
...
+	for _, ip := range node.Status.Addresses {
+		_, exists := IPAddressMap[ip.Address]
+		if !exists && ip.Type == v1.NodeInternalIP {
```
eranco74, line 174: Seems that you should check the map only for the nodeInternalIP.
Also, consider passing to this function just the node addresses or just the nodeInternalIP

tsorya: Will require for loop outside the function to search for internal ip. Don't want to handle it outside actually.

eranco74:
```suggestion
	for _, ip := range node.Status.Addresses {
		if ip.Type == v1.NodeInternalIP {
			_, exists := IPAddressMap[ip.Address]
			if !exists {
```

PR #567, opened 2022-10-31
NO-ISSUE: improvement on gathering controller logs, validating that buffer is not empty

tsorya:
NO-ISSUE: improvement on gathering controller logs, validating that buffer is not empty


eranco74 (commented):
/lgtm

PR #572, opened 2022-11-07
MGMT-12469: Add resolv.conf content to controller logs in order to improve dns issue visibility

tsorya:
MGMT-12469: Add resolv.conf content to controller logs in order to improve dns issue visibility

Currently when controller has an issue with dns pointing to wrong ip, we don't know why. We should add resolv.conf to controller logs to be able to see that it was managed successfully


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1123,7 +1134,11 @@ func (c controller) logRouterStatus() {
 			if err != nil {
 				c.log.WithError(err).Warning("Failed to reach console")
 			} else {
-				c.log.Infof("canary url status %s", r.Status)
+				response, errR := ioutil.ReadAll(r.Body)
```
omertuc, line 1137:
```suggestion
				defer r.Body.Close()
				response, errR := ioutil.ReadAll(r.Body)
```

omertuc: Also use `io.ReadAll` since `ioutil` is deprecated

src/ops/ops.go
```diff
@@ -832,6 +833,10 @@ func (o *ops) ExecPrivilegeCommand(liveLogger io.Writer, command string, args ..
 	return o.executor.ExecCommand(liveLogger, commandBase, arguments...)
 }
 
+func (o *ops) ReadFile(filePath string) ([]byte, error) {
+	return ioutil.ReadFile(filePath)
```
omertuc, line 837:
```suggestion
	return os.ReadFile(filePath)
```

PR #573, opened 2022-11-07
NO-ISSUE: Log informative error in case subscription doesn't have CSV

eranco74:


src/k8s_client/k8s_client.go
```diff
@@ -598,12 +598,16 @@ func (c *k8sClient) CreateEvent(namespace, name, message, component string) (*v1
...
 func (c *k8sClient) GetCSVFromSubscription(namespace string, name string) (string, error) {
-	result, err := c.olmClient.Subscriptions(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	subscription, err := c.olmClient.Subscriptions(namespace).Get(context.TODO(), name, metav1.GetOptions{})
 	if err != nil {
 		c.log.WithError(err).Warnf("Failed to get subscription %s in %s", name, namespace)
 		return "", err
 	}
-	return result.Status.CurrentCSV, nil
+	if subscription.Status.CurrentCSV == "" {
+		c.log.WithError(err).Warnf("subscription CurrentCSV is empty, subscription status: %+v", subscription.Status)
```
tsorya, line 607: return error, it should not happen i assume ?

omertuc: It does happen, while we wait

tsorya: so not better to return error in that case?

omertuc: hmm not sure. Depends on how the users of this function behave

eranco74: It get called in operatorHandler `IsInitialized` and `GetStatus` in both cases it's eventually wrapped in a loop.
I don't want to change behavior here. just add logging so it's clear that the CSV is empty and whats the subscription status

eranco74:
devscript setup failed
/test edge-e2e-ai-operator-ztp

src/k8s_client/k8s_client.go
```diff
@@ -598,12 +598,17 @@ func (c *k8sClient) CreateEvent(namespace, name, message, component string) (*v1
...
```
rccrdpccl, line 607: variable `err` passed to `WithError` will always be nil. We could reduce noise by not passing error context

eranco74:
packet issue
/retest-required

eranco74:
devscripts-setup failed
/retest-required

PR #574, opened 2022-11-08
MGMT-12471: Don't wait for console if it is disabled

jhernand:
This patch changes the controller so that it doesn't wait for the console operator when the console capability is explicitly disabled.

Related: https://issues.redhat.com/browse/MGMT-12471


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -45,6 +45,7 @@ const (
 	// zero means -> bring all logs
 	controllerLogsSecondsAgo  = 0
 	consoleOperatorName       = "console"
+	consoleCapabilityName     = "Console"
```
omertuc, line 48: Can we:
1. Stop our weird `go.mod` `replace` directive for openshift/api and bring it as a proper dependency?
2. Use [the actual constant](https://github.com/openshift/api/blob/7287964c76316b1d6636d31d7df7f82ee1e33d45/config/v1/types_cluster_version.go#L251) for this

Let's not hold back this PR though and follow up with a separate PR

omertuc (approved):
Looks perfect, just one small remark

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -992,12 +993,31 @@ func (c controller) waitingForClusterOperators(ctx context.Context) error {
 	ctxWithTimeout, cancel := context.WithTimeout(ctx, CVOMaxTimeout)
 	defer cancel()
 	isClusterVersionAvailable := func(timer *time.Timer) bool {
-		result := c.isOperatorAvailable(NewClusterOperatorHandler(c.kc, consoleOperatorName))
-
+		clusterOperatorHandler := NewClusterOperatorHandler(c.kc, consoleOperatorName)
+		clusterVersionHandler := NewClusterVersionHandler(c.log, c.kc, timer)
+		disabledCapabilities, err := clusterVersionHandler.GetDisabledCapabilities()
```
romfreiman, line 998: @wking ptal

src/assisted_installer_controller/assisted_installer_controller.go
jhernand: This would be the first step, I think, just updating to Go 1.19: #575.

src/k8s_client/k8s_client.go
```diff
@@ -615,3 +617,14 @@ func (c *k8sClient) PatchNodeLabels(nodeName string, nodeLabels string) error {
...
+func (c *k8sClient) IsClusterCapabilityEnabled(capability configv1.ClusterVersionCapability) (result bool, err error) {
+	version, err := c.GetClusterVersion()
+	if err != nil {
+		return
+	}
+	isExplicitlyEnabled := funk.Contains(version.Status.Capabilities.EnabledCapabilities, capability)
+	isKnown := funk.Contains(version.Status.Capabilities.KnownCapabilities, capability)
+	result = isExplicitlyEnabled || !isKnown
+	return
+}
```
omertuc, line 630: nit: Named return values seem unnecessary in this simple function

```suggestion
func (c *k8sClient) IsClusterCapabilityEnabled(capability configv1.ClusterVersionCapability) (bool, error) {
	version, err := c.GetClusterVersion()
	if err != nil {
		return false, err
	}
	isExplicitlyEnabled := funk.Contains(version.Status.Capabilities.EnabledCapabilities, capability)
	isKnown := funk.Contains(version.Status.Capabilities.KnownCapabilities, capability)
 	return isExplicitlyEnabled || !isKnown
}
```

PR #580, opened 2022-11-14
change all 'go get' commands to 'go install'

osherdp:
``go get`` might result in ``go.mod`` changes meaning we might use different dependencies versions than intended.
In addition, this kind of usage of ``go get`` is deprecated in go 1.17 and is blocked in go version >= 1.19.


osherdp:
/hold
waiting for tests

omertuc:
>will likely get blocked in newer go versions.

It's not likely, it's already blocked since (I think) 1.19

osherdp:
/unhold
@omertuc can you re-lgtm?

PR #583, opened 2022-12-01
MGMT-12339: Add extra step to force clean disks with LVM volumes

javipolo:
Forcing the removal of affected DeviceMapper devices should fix the problems when properly running disk_speed_check before disk cleanup

/cc @vrutkovs 
/cc @paul-maidment

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -95,8 +95,8 @@ func (i *installer) InstallNode() error {
 	i.Config.Device = i.ops.EvaluateDiskSymlink(i.Config.Device)
 	err := i.cleanupInstallDevice()
 	if err != nil {
+		// Do not change the phrasing of this error message, as we rely on it in a triage signature
 		i.log.Errorf("failed to prepare install device %s, err %s", i.Device, err)
-		return err
 	}
 
 	if err = i.ops.Mkdir(InstallDir); err != nil {
@@ -659,11 +659,12 @@ func (i *installer) cleanupInstallDevice() error {
 		return nil
 	}
 
+	var ret error
 	i.log.Infof("Start cleaning up device %s", i.Device)
 	err := i.cleanupDevice(i.Device)
 
 	if err != nil {
-		return err
+		ret = utils.CombineErrors(ret, err)
 	}
 
 	if i.ops.IsRaidMember(i.Device) {
@@ -672,7 +673,7 @@ func (i *installer) cleanupInstallDevice() error {
 		devices, err = i.ops.GetRaidDevices(i.Device)
 
 		if err != nil {
-			return err
+			ret = utils.CombineErrors(ret, err)
 		}
 
 		for _, device := range devices {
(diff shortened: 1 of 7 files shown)


src/installer/installer.go
```diff
@@ -722,7 +722,12 @@ func (i *installer) cleanupDevice(device string) error {
 		}
 	}
 
-	return i.ops.RemoveAllPVsOnDevice(device)
+	err = i.ops.RemoveAllPVsOnDevice(device)
```
vrutkovs, line 725:
```suggestion
	if err = i.ops.RemoveAllPVsOnDevice(device); err != nil {
```

javipolo: Changed :+1:

src/installer/installer.go
```diff
@@ -722,7 +722,11 @@ func (i *installer) cleanupDevice(device string) error {
 		}
 	}
 
-	return i.ops.RemoveAllPVsOnDevice(device)
+	if err = i.ops.RemoveAllPVsOnDevice(device); err != nil {
+		return err
```
omertuc, line 726: Do we really want a hard error here? Let's keep it best-effort

omertuc: Or let's just remove [this](https://github.com/openshift/assisted-installer/blob/8027b73de9c36396f90530eb2b5481dc4b3532db/src/installer/installer.go#L99) line

omertuc: hmm yes

src/utils/utils_test.go
```diff
@@ -116,3 +118,23 @@ var _ = Describe("Verify_utils", func() {
 		})
 	})
 })
+
+var _ = FDescribe("CombineErrors", func() {
```
omertuc, line 122:
```suggestion
var _ = Describe("CombineErrors", func() {
```

src/installer/installer.go
```diff
@@ -672,57 +672,72 @@ func (i *installer) cleanupInstallDevice() error {
 		devices, err = i.ops.GetRaidDevices(i.Device)
 
 		if err != nil {
-			return err
+			ret = utils.CombineErrors(ret, err)
+			return ret
```
omertuc, line 676: Is this `return` on purpose?

javipolo: Wops, that should not be there :)

src/installer/installer.go
```diff
@@ -96,7 +96,6 @@ func (i *installer) InstallNode() error {
 	err := i.cleanupInstallDevice()
 	if err != nil {
 		i.log.Errorf("failed to prepare install device %s, err %s", i.Device, err)
-		return err
```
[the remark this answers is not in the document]
javipolo: Ok, I'll do that. I guess that only looking for 'failed to prepare install device' is good enough, with a message to look for lvm/raid cleanup messages in the logs, right?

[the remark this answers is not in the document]
omertuc: yeah it's rather distinct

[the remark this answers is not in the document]
omertuc: But leave a note on top of it saying "Do not change the phrasing of this message as we rely on it in a triage signature"

vrutkovs (approved):
/approve
/lgtm

src/installer/installer.go
```diff
@@ -672,57 +673,71 @@ func (i *installer) cleanupInstallDevice() error {
...
-		err = i.removeVolumeGroupsFromDevice(vgNames, device)
-		if err != nil {
-			return err
+		if err = i.removeVolumeGroupsFromDevice(vgNames, device); err != nil {
+			ret = utils.CombineErrors(ret, err)
 		}
 	}
 
-	return i.ops.RemoveAllPVsOnDevice(device)
+	if err = i.ops.RemoveAllPVsOnDevice(device); err != nil {
```
eranco74, line 732: If this is successful we can ignore the previous error while removing the VG, no?

eranco74:
This change the behavior of the disk cleanup to best effort?
if we fail during disk cleanup we will still try to install and might fail later. 
This means that installation failures caused due to unsuccessful disk cleanup will be a lot harder to debug.
/hold

src/installer/installer.go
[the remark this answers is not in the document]
eranco74: Perhaps we can also add an event for this case.

src/installer/installer.go
```diff
@@ -95,8 +95,8 @@ func (i *installer) InstallNode() error {
 	i.Config.Device = i.ops.EvaluateDiskSymlink(i.Config.Device)
 	err := i.cleanupInstallDevice()
 	if err != nil {
+		// Do not change the phrasing of this error message, as we rely on it in a triage signature
 		i.log.Errorf("failed to prepare install device %s, err %s", i.Device, err)
-		return err
```
vrutkovs, line 99: Lets emit an event here too - self-hosted users don't get the benefits of triage signatures so it would be useful too (thanks @eranco74 for the suggestion)

vrutkovs (changes requested):

omertuc:
> This change the behavior of the disk cleanup to best effort? if we fail during disk cleanup we will still try to install and might fail later. This means that installation failures caused due to unsuccessful disk cleanup will be a lot harder to debug. /hold

https://github.com/openshift/assisted-installer/pull/583?notification_referrer_id=NT_kwDOAKYMDrM0OTc0NTI5MzAzOjEwODgyMDYy#discussion_r1042055874 
https://github.com/openshift-assisted/assisted-installer-deployment/pull/316

src/installer/installer.go
javipolo: We could, but I wanted to give all the information possible in case of a failure

Anyways, it would be pretty weird that pvremove works if the VG is not removed before ... I guess in that case I'd like to know anyways, because that's something funky

src/installer/installer.go
javipolo: sure, can someone point me to some code emitting an event? O:)

javipolo:
> This change the behavior of the disk cleanup to best effort? if we fail during disk cleanup we will still try to install and might fail later. This means that installation failures caused due to unsuccessful disk cleanup will be a lot harder to debug. /hold

As @omertuc stated, we're already notifying of this via signature, and also errors happening after cleanup not doing its thing are always the same (partition table busy or similar) so they're not that hard to spot

src/installer/installer.go
eranco74: So, emitting an event is tricky.
The assisted installer can't create an event, what it can do is update the installation progress and in response to some of these updates the assisted-service generates events.
So I suggest you add a call to UpdateHostInstallProgress with the same stage and info message saying that the cleanup failed.
Later we can add the code to the service that identifies the string in the progress info and emits an event. (edited)

PR #597, opened 2022-12-13
MGMT-12339: Update host install progress on cleanup error

javipolo:
Update the host install progress with a message if disk cleanup failed. This way we can catch it in assisted-service and emit an event that will ease debugging

/cc @eranco74

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -95,6 +95,7 @@ func (i *installer) InstallNode() error {
 	i.Config.Device = i.ops.EvaluateDiskSymlink(i.Config.Device)
 	err := i.cleanupInstallDevice()
 	if err != nil {
+		i.UpdateHostInstallProgress(models.HostStageStartingInstallation, fmt.Sprintf("Could not clean install device %s. The installation will continue. If the installation fails, clean the disk and try again", i.Device))
 		// Do not change the phrasing of this error message, as we rely on it in a triage signature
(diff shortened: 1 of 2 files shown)


eranco74 (commented):
/lgtm

eranco74:
Though I suspect a few UT might fail...

src/installer/installer.go
```diff
@@ -95,6 +95,7 @@ func (i *installer) InstallNode() error {
 	i.Config.Device = i.ops.EvaluateDiskSymlink(i.Config.Device)
 	err := i.cleanupInstallDevice()
 	if err != nil {
+		i.UpdateHostInstallProgress(models.HostStageStartingInstallation, fmt.Sprintf("Failed to cleanup install device %s", i.Device))
```
omertuc, line 98: This is user facing. We should soften the error

javipolo: What about "Non fatal error: could not clean install device %s, installation will continue" ?

eranco74: Agree, @wolfeallison PTAL

wolfeallison: For error messages, we try to follow the pattern description, reason, resolution.  
"Could not clean install device
The installation will continue. If the installation fails, clean the disk and try again."

What do you think of this? @javipolo @eranco74

javipolo: I like it :+1: , let's see how @eranco74 feels about it

PR #601, opened 2022-12-22
MGMT-12908: Labeling nodes as quick as possible and fixing hostname p…

tsorya:
[MGMT-12908](https://issues.redhat.com//browse/MGMT-12908): Labeling nodes as quick as possible and fixing hostname possible issue
Moving labeling part to run in parallel with waiting for hosts and order to set labels as fast as possible

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -449,16 +449,24 @@ func (c controller) PostInstallConfigs(ctx context.Context, wg *sync.WaitGroup) 
 	c.sendCompleteInstallation(ctx, success, errMessage)
 }
 
+func (c controller) UpdateNodeLabels(ctx context.Context, wg *sync.WaitGroup) {
+	defer func() {
+		c.log.Infof("Finished UpdateNodeLabels")
+		wg.Done()
+	}()
+
+	c.log.Infof("Updating host labels if required")
+	err := utils.WaitForPredicateWithContext(ctx, LongWaitTimeout, GeneralWaitInterval, c.updateNodesLabels)
+	if err != nil {
+		c.log.Warn("Failed to label the nodes")
+	}
+}
+
 func (c controller) postInstallConfigs(ctx context.Context) error {
 	var err error
 
 	if err = c.waitingForClusterOperators(ctx); err != nil {
 		return errors.Wrapf(err, "Timeout while waiting for cluster operators to be available")
-	}
-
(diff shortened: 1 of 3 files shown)


omertuc:
Why

tsorya:
There is a need to apply it before cluster is actually installed. We want to set it as node joins the cluster in order to minimise time host will be without label.

tsorya:
It was asked by kaboom sometime ago to halp in management cactus deployment on infra nodes

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1022,35 +1030,53 @@ func (c *controller) updateNodesLabels() bool {
...
 
-		err = c.kc.PatchNodeLabels(node.Name, hostData.Host.NodeLabels)
+		err = c.kc.PatchNodeLabels(node.Name, host.Host.NodeLabels)
 		if err != nil {
-			log.WithError(err).Errorf("Failed to patch node %s with node labels %s", node.Name, hostData.Host.NodeLabels)
-			retry = true
+			log.WithError(err).Errorf("Failed to patch node %s with node labels %s", node.Name, host.Host.NodeLabels)
+			continue
 		}
+		log.Infof("Successfully patched label for host %s", node.Name)
```
eranco74, line 1076:
```suggestion
		log.Infof("Successfully patched labels for host %s", node.Name)
```

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1022,35 +1030,53 @@ func (c *controller) updateNodesLabels() bool {
 		return KeepWaiting
 	}
 
-	retry := false
-
+	hostsWithLabels := make(map[string]inventory_client.HostData)
```
eranco74, line 1033: All you need is the labels not all the HostData
```suggestion
	hostsWithLabels := make(map[string]string)
```

tsorya: not right, i need host data in order to correlate k8s node with assisted host and hostname is not enough

eranco74: OK

eranco74 (commented):
/lgtm

romfreiman:
not sure what does it mean 'as quick as possible'. I assume it's still racy, no?

tsorya:
for sure but at least it will be maximum 30 seconds after host joined
Another option is to add node listener and react on events. Though it will be something new in controller logic and will take sometime to develop

PR #610, opened 2023-01-05
MGMT-13141: assisted-installer-controller shouldn't try to update BMH if the cluster platform is not BareMetal

tsorya:
[MGMT-13141](https://issues.redhat.com//browse/MGMT-13141): assisted-installer-controller shouldn't try to update BMH if the cluster platform is not BareMetal

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -671,7 +671,7 @@ func (c controller) UpdateBMHs(ctx context.Context, wg *sync.WaitGroup) {
 		bmhs, err := c.kc.ListBMHs()
 		if err != nil {
 			c.log.WithError(err).Errorf("Failed to list BMH hosts")
-			return false
+			return KeepWaiting
 		}
 
 		c.log.Infof("Number of BMHs is %d", len(bmhs.Items))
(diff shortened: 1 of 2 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -667,29 +667,46 @@ func (c controller) UpdateBMHs(ctx context.Context, wg *sync.WaitGroup) {
 		c.log.Infof("Finished UpdateBMHs")
 		wg.Done()
 	}()
+
+	var cluster *models.Cluster
+	var err error
```
eranco74, line 672: We get the cluster in the `SetReadyState` you can get it from there and just not start this goroutine.

tsorya: Optional yes

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -144,8 +144,12 @@ func main() {
 	wg.Add(1)
 	go assistedController.PostInstallConfigs(mainContext, &wg)
 	wg.Add(1)
-	go assistedController.UpdateBMHs(mainContext, &wg)
-	wg.Add(1)
+
+	if *cluster.Platform.Type != models.PlatformTypeBaremetal {
+		logger.Infof("Cluster platform is not %s, skipping BMH", models.PlatformTypeBaremetal)
+		go assistedController.UpdateBMHs(mainContext, &wg)
```
omertuc, line 150: shouldn't this be outside the `if`?

tsorya: right :(

tsorya: fixed

PR #613, opened 2023-01-09
MGMT-13229: SNO: Start controller when node is not ready, right after joined

tsorya:
[MGMT-13229](https://issues.redhat.com//browse/MGMT-13229): SNO: Start controller when node is not ready, right after joined
     We want to start controller as fast as possible after reboot and to minimize time when we don't know what is going on.
    This will allow us to start when kubelet will join to itself and we will be able
    to send joined stage to service and wait till node will become ready while controller is running already.
    In order to start running while control plane kube-api doesn't exists yet , going to use bootstrap kubeconfig as it is part of filesystem.

### DIFF
--- a/deploy/assisted-installer-controller/assisted-installer-controller-pod.yaml.template
+++ b/deploy/assisted-installer-controller/assisted-installer-controller-pod.yaml.template
@@ -75,6 +75,10 @@ spec:
               mountPath: /tmp/var-run-resolv.conf
             - name: host-resolv-conf
               mountPath: /tmp/host-resolv.conf
+          {{if .SNO}}
+            - name: sno-bootstrap-files
+              mountPath: /tmp/bootstrap-secrets
+          {{end}}
           {{if .CACertPath}}
             - name: service-ca-cert-config
               mountPath: {{.CACertPath}}
@@ -88,6 +92,14 @@ spec:
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
         operator: Exists
+      {{if .SNO}}
+      - effect: NoSchedule
+        key: node.kubernetes.io/not-ready
+        operator: Exists
+      - effect: NoSchedule
+        key: node.kubernetes.io/network-unavailable
+        operator: Exists
+      {{end}}
       volumes:
         - name: host-ca-bundle
           hostPath:
@@ -98,6 +110,11 @@ spec:
         - name: host-resolv-conf
           hostPath:
             path: /etc/resolv.conf
+        {{if .SNO}}
+        - name: sno-bootstrap-files
+          hostPath:
+            path: /etc/kubernetes/bootstrap-secrets
+        {{end}}
         {{if .CACertPath}}
         - name: service-ca-cert-config
           hostPath:
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -18,6 +18,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/thoas/go-funk"
+
 	"github.com/hashicorp/go-version"
 	metal3v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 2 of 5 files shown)


src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -83,7 +84,13 @@ func main() {
 
 	var kc k8s_client.K8SClient
 	if !Options.ControllerConfig.DryRunEnabled {
-		kc, err = k8s_client.NewK8SClient("", logger)
+		// in case of sno we want controller to start as fast as possible,
+		//in that case we want to use kubeconfig from filesystem and not pulling it from api
```
eranco74, line 88: Note that the bootstrap kubeconfig get deleted once the bootstrap process is complete

tsorya: Yep though object is already created and everything looks fine

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -246,7 +248,7 @@ func (c *controller) waitAndUpdateNodesStatus() bool {
 		}
 		common.LogIfHostIpChanged(c.log, node, knownIpAddresses)
 
-		if host.Host.Progress.CurrentStage == models.HostStageConfiguring {
+		if funk.Contains([]models.HostStage{models.HostStageConfiguring, models.HostStageRebooting}, host.Host.Progress.CurrentStage) {
```
omertuc, line 251: See https://issues.redhat.com/browse/MGMT-12303

omertuc: In particular:

>Also we want to introduce new stages with the new reporter service, and so being strict with requiring Configuring will interfere with that 

i.e. I think we should remove this condition altogether, and not be strict about the particular stage, as we might introduce the reporter soon and that means adding new stages which would have to be added to this `if` condition

tsorya: Not sure what is reporter and why we are going to add new stages, we still need to know when to send joined and when to send done, I don't want to send something each loop run.
So it will fix [MGMT-12303](https://issues.redhat.com/browse/MGMT-12303)

omertuc: >Not sure what is reporter

https://github.com/openshift/assisted-service/blob/master/docs/enhancements/boot-status-reporter.md

tsorya: It is not going to add any status or update installation flow, will be used for logs upload for now at least

omertuc: The proposal suggests adding new progress stages

omertuc: Anyway we'll just have to remember to update this condition when those new stages are added, that's all. Or find some way to get rid of this condition without having to list stages

deploy/assisted-installer-controller/assisted-installer-controller-pod.yaml.template
```diff
@@ -98,6 +110,11 @@ spec:
         - name: host-resolv-conf
           hostPath:
             path: /etc/resolv.conf
+        {{if .SNO}}
+        - name: sno-bootstrap-kubeconfig
+          hostPath:
+            path: /etc/kubernetes/bootstrap-secrets/kubeconfig
```
eranco74, line 116: mount the directory instead.
It should allow the cornet case of starting the controller without the bootstrap kubeconfig (it should fall back to the default kubeconfig).

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -329,6 +329,31 @@ var _ = Describe("installer HostRoleMaster role", func() {
 			Expect(exit).Should(Equal(false))
 		})
 
+		It("WaitAndUpdateNodesStatus including joined state from reboot", func() {
+			joined := []models.HostStage{models.HostStageJoined,
+				models.HostStageJoined,
+				models.HostStageJoined}
+
+			hosts := create3Hosts(models.HostStatusInstalling, models.HostStageRebooting, "")
```
eranco74, line 337: It's not clear to me why a PR that changes the behavior in the case of SNO is adding this single test with 3 nodes.
Though it shouldn't happen, I guess it is valid, yet please add a test for the actual flow this code tries to solve: SNO transitioning to `Joined` from `Rebooting`

tsorya: cause it is not specific for sno, this change allows transition from reboot to joined.

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -329,6 +329,58 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
+			for _, node := range nodes.Items {
+				for i, cond := range node.Status.Conditions {
+					if cond.Type == v1.NodeReady {
+						node.Status.Conditions[i].Status = v1.ConditionFalse
+					}
+				}
+			}
+			mockk8sclient.EXPECT().ListNodes().Return(nodes, nil).Times(1)
+			updateProgressSuccess(joined, inventoryNamesIds)
+			configuringSuccess()
```
eranco74, line 378: I thought that we don't have configuring in SNO?

eranco74 (commented):
/lgtm

PR #615, opened 2023-01-12
MGMT-13131: shorten SNO installation duration

eranco74:
Execute shutdown -r now instead of 1 minute from now this should save 1 minute during the installation
The side effect of this change is that the agent will not report back the install command result upon success
This side effect can be mitigated by adding inhibit lock on the agent side that will prevent the node from rebooting until the agent allow it.

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -166,7 +166,7 @@ func (o *ops) FormatDisk(disk string) error {
 
 func (o *ops) Reboot() error {
 	o.log.Info("Rebooting node")
-	_, err := o.ExecPrivilegeCommand(o.logWriter, "shutdown", "-r", "+1", "'Installation completed, server is going to reboot.'")
+	_, err := o.ExecPrivilegeCommand(o.logWriter, "shutdown", "-r", "now", "'Installation completed, server is going to reboot.'")
 	if err != nil {
 		o.log.Errorf("Failed to reboot node, err: %s", err)
 		return err


rccrdpccl:
Why was 1 minute wait there to start with?

romfreiman:
I think it's more important for us to be sure that we get logs rather than saving 1 min?

omertuc:
> by adding inhibit lock on the agent side 

/hold Shouldn't we do that before merging this PR?

eranco74:
> /hold Shouldn't we do that before merging this PR?
Nope, I find it useless, (unless we find that we need to wait for the agent to complete something which doesn't seem to be the case).

eranco74:
> I think it's more important for us to be sure that we get logs rather than saving 1 min?

Agree, I'm sure :wink:

eranco74:
Seems like an issue with the nightly release image:
```
INFO     root:operators_utils.py:30 Asked operators to be in one of the statuses from ['available'] and currently operators statuses are [('console', 'available', 'All is well'), ('cvo', 'progressing', 'Unable to apply 4.12.0-0.nightly-2023-01-10-062211: Cluster operator etcd is degraded')]
```

/retest-required

eranco74:
we get the logs:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-installer/615/pull-ci-openshift-assisted-installer-master-edge-e2e-metal-assisted/1613509467943473152/artifacts/e2e-metal-assisted/assisted-common-gather/artifacts/2023-01-12_12-45-44_2bb3480e-44f6-4d46-8a63-465ac1a833ce/

tsorya:
We don't need to wait for the logs as they will be sent earlier and it is blocking function. We had this minute only in order to handle reply to the service but we don't need it at all in case of success

PR #617, opened 2023-01-17
MGMT-10785: report failing cluster operators controller -> service

slaviered:
The controller part of PR https://github.com/openshift/assisted-service/pull/4864


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1094,15 +1097,29 @@ func (c controller) sendCompleteInstallation(ctx context.Context, isSuccess bool
 
 // logClusterOperatorsStatus logging cluster operators status
 func (c controller) logClusterOperatorsStatus() {
+	_, _ = c.operatorReport(true)
```
tsorya, line 1100: can we log it here? and remove this boolean?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -434,7 +435,8 @@ func (c controller) PostInstallConfigs(ctx context.Context, wg *sync.WaitGroup)
 	}
 
 	errMessage := ""
-	err = c.postInstallConfigs(ctx)
+	var data map[string]interface{}
+	err, data = c.postInstallConfigs(ctx)
```
omertuc, line 439:
```suggestion
	err, data := c.postInstallConfigs(ctx)
```

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1094,15 +1097,29 @@ func (c controller) sendCompleteInstallation(ctx context.Context, isSuccess bool
...
+		return nil, err
 	}
 
+	report := make([]models.OperatorMonitorReport, 0)
 	for _, operator := range operators.Items {
-		c.log.Infof("Operator %s, statuses: %v", operator.Name, operator.Status.Conditions)
+		if withLog {
+			c.log.Infof("Operator %s, statuses: %v", operator.Name, operator.Status.Conditions)
+		}
+		status, _, reason := utils.MonitoredOperatorStatus(operator.Status.Conditions)
```
omertuc, line 1115: Why do we drop the message? It contains a lot of detailed information, usually

slaviered: The log will be issued from the logger thread. In the postInstallConfig function we just need to collect the raw data to send to the service. So the log is printed just as before.

omertuc: Your response seems to imply you think I was talking about the `if withLog` block, but my comment is for the line immediately below that block, line 1115

omertuc:
>/lgtm

Please don't lgtm with unresolved conversations

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: Done

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: Done

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: I did not drop anything really. The previous code just printed out the log in line 1113 and the line was kept as is. In addition to that I am assembling the report and the message was not used because it is very verbose. The requirements, as I got them from Avishay, were to keep the information as short as possible. In the BE I further omit extra info and prints only the operator names (so actually neither the reason nor the message is actually used).

omertuc: >I did not drop anything really. The previous code just printed out the log in line 1113 and the line was kept as is
I am not talking about the log line, I'm talking about the second return value of `MonitoredOperatorStatus`, that has the value `condition.Message`.

>In addition to that I am assembling the report and the message was not used because it is very verbose.

It's just 2-20 lines of text, what's wrong with it being verbose? It's very informative, why are we aiming to keep it short? Shouldn't keeping it short be a concern of the UI? It would be nice to have those messages / reasons saved in the backend and maybe even displayed to the user in the UI in the future

But if you've explicitly decided against including them then that's fine, you can unhold

slaviered: The point of the ticket was to send an event which is quite limited in what it can present in a readable fashion. After discussion it was decided to reduce the data presented to the user even further and print only the names of the failed operators. The field "reason" and "message" are quite interchangeable (you do not need them both). If you think that we should send the message instead I can certainly change the code here. In any case, it will not be presented in the UI in the current implementation - but may serve future endeavors.

omertuc: >The field "reason" and "message" are quite interchangeable

Interchangeable in what sense? They're very different

The "reason" is a very very short camel case "ID" of the error, it's almost meaningless for the user

The message is a few very informative lines of text that explain the issue in detail, and often include complete error messages which often contain clues to help the cluster owner understand what they can do to fix the problem

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1094,17 +1102,30 @@ func (c controller) sendCompleteInstallation(ctx context.Context, isSuccess bool
...
-		return
+		return nil
 	}
 
+	report := make([]models.OperatorMonitorReport, 0)
 	for _, operator := range operators.Items {
+		//Note: always keep this log so we have all the information
+		//in the controller's log
 		c.log.Infof("Operator %s, statuses: %v", operator.Name, operator.Status.Conditions)
+		status, message, _ := utils.MonitoredOperatorStatus(operator.Status.Conditions)
```
omertuc, line 1121: I just noticed this only returns a single condition. And it would only return the first of `Available`/`Failed`, depending on the order they appear in the conditions list... This is very odd behavior. What's the reasoning behind it?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1094,17 +1102,30 @@ func (c controller) sendCompleteInstallation(ctx context.Context, isSuccess bool
...
 	if err != nil {
 		c.log.WithError(err).Warning("Failed to list cluster operators")
-		return
+		return nil
 	}
 
+	report := make([]models.OperatorMonitorReport, 0)
 	for _, operator := range operators.Items {
+		//Note: always keep this log so we have all the information
+		//in the controller's log
```
omertuc, line 1119:
```suggestion
		// Note: always keep this log so we have all the information
		// in the controller's log
```

omertuc (changes requested):

src/assisted_installer_controller/assisted_installer_controller.go
slaviered: Done.

PR #618, opened 2023-01-18
MGMT-13385: Installation fail due to race condition in etcd bootstrap strategy

eranco74:
reboot now only when installing single node


eranco74:
/cc @osherdp 
MGMT-13385 is a blocker

eranco74:
/override ci/prow/edge-e2e-ai-operator-ztp
 LSO issue

PR #619, opened 2023-01-24
MGMT-13463: Assisted-controller can fail to send summary logs and we  will not have any logs

tsorya:
[MGMT-13463](https://issues.redhat.com//browse/MGMT-13463): Assisted-controller can fail to send summary logs and we will not have any logs

Sometimes controller can fail to stream tar.gz of must-gather and no files will be send as we will close write pipe and will not try to stream another tar entry.
This change will log the error but will try to stream all other tar entries we pass to the stream function


eranco74 (commented):
/lgtm

PR #620, opened 2023-01-28
MGMT-13507: Apply node labels only when node is Ready

tsorya:
[MGMT-13507](https://issues.redhat.com//browse/MGMT-13507): Apply node labels only when node is Ready
Juniper have a problem with their deployment in case we apply node labels before node is read.
In order to fix it we should apply them only in case node is ready

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller_test.go
+++ b/src/assisted_installer_controller/assisted_installer_controller_test.go
@@ -1103,24 +1103,35 @@ var _ = Describe("installer HostRoleMaster role", func() {
 			// 3. Set last node as ready and patch labels for it
 			It("Set label only in when nodes are ready", func() {
 				nodeLabels := `{"node.ocs.openshift.io/storage":""}`
-				notReadyNodeName := ""
-				k8sNodesWith1NotReady := GetKubeNodes(kubeNamesIds)
-				for i, cond := range k8sNodesWith1NotReady.Items[0].Status.Conditions {
+				k8sReadyNodes := GetKubeNodes(kubeNamesIds)
+				k8sNodesWith1NotReady := k8sReadyNodes.DeepCopy()
+				conds := k8sNodesWith1NotReady.Items[0].Status.Conditions
+				for i, cond := range conds {
 					if cond.Type == v1.NodeReady {
-						k8sNodesWith1NotReady.Items[0].Status.Conditions[i].Status = v1.ConditionFalse
+						conds[i].Status = v1.ConditionFalse
 					}
-					notReadyNodeName = k8sNodesWith1NotReady.Items[0].Name
 				}
+				k8sNodesWith1NotReady.Items[0].Status.Conditions = conds
+				notReadyNodeName := k8sNodesWith1NotReady.Items[0].Name
 
 				hosts := create3Hosts(models.HostStatusInstalled, models.HostStageDone, nodeLabels)
 				mockbmclient.EXPECT().GetHosts(gomock.Any(), gomock.Any(), []string{models.HostStatusDisabled, models.HostStatusError}).
 					Return(hosts, nil).Times(2)
+
+				// set node labels to k8s object for 2 already patched nodes
+				k8sReadyNodes.Items[1].ObjectMeta.Labels["node.ocs.openshift.io/storage"] = ""
+				k8sReadyNodes.Items[2].ObjectMeta.Labels["node.ocs.openshift.io/storage"] = ""
+
 				// first run with 2 ready nodes
-				mockk8sclient.EXPECT().ListNodes().Return(k8sNodesWith1NotReady, nil).Times(1)
-				// Set 3rd node as ready
-				mockk8sclient.EXPECT().ListNodes().Return(GetKubeNodes(kubeNamesIds), nil).Times(1)
-				mockk8sclient.EXPECT().PatchNodeLabels(gomock.Any(), nodeLabels).Return(nil).Times(2)
-				mockk8sclient.EXPECT().PatchNodeLabels(notReadyNodeName, nodeLabels).Return(nil).Times(1)
+				gomock.InOrder(
+					mockk8sclient.EXPECT().ListNodes().Return(k8sNodesWith1NotReady, nil).Times(1),
+					mockk8sclient.EXPECT().ListNodes().Return(k8sReadyNodes, nil).Times(1),
+				)
+
+				gomock.InOrder(
+					mockk8sclient.EXPECT().PatchNodeLabels(gomock.Any(), nodeLabels).Return(nil).Times(2),
+					mockk8sclient.EXPECT().PatchNodeLabels(notReadyNodeName, nodeLabels).Return(nil).Times(1),
+				)
 
 				wg.Add(1)
 				assistedController.UpdateNodeLabels(context.TODO(), &wg)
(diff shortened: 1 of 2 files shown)


src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -1098,7 +1098,27 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
+					}
+				}
+
+				hosts := create3Hosts(models.HostStatusInstalled, models.HostStageDone, nodeLabels)
+				mockbmclient.EXPECT().GetHosts(gomock.Any(), gomock.Any(), []string{models.HostStatusDisabled, models.HostStatusError}).
+					Return(hosts, nil).Times(2)
+				mockk8sclient.EXPECT().ListNodes().Return(k8sNodes, nil).Times(1)
+				mockk8sclient.EXPECT().ListNodes().Return(GetKubeNodes(kubeNamesIds), nil).Times(1)
+				mockk8sclient.EXPECT().PatchNodeLabels(gomock.Any(), nodeLabels).Return(nil).Times(2)
+				mockk8sclient.EXPECT().PatchNodeLabels(gomock.Any(), nodeLabels).Return(nil).Times(1)
```
eranco74, line 1116:
```suggestion
				mockk8sclient.EXPECT().PatchNodeLabels(gomock.Any(), nodeLabels).Return(nil).Times(3)
```
?

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -1098,7 +1098,27 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
+				nodeLabels := `{"node.ocs.openshift.io/storage":""}`
+				k8sNodes := GetKubeNodes(kubeNamesIds)
+				for i, cond := range k8sNodes.Items[0].Status.Conditions {
+					if cond.Type == v1.NodeReady {
+						k8sNodes.Items[0].Status.Conditions[i].Status = v1.ConditionFalse
+					}
+				}
+
+				hosts := create3Hosts(models.HostStatusInstalled, models.HostStageDone, nodeLabels)
+				mockbmclient.EXPECT().GetHosts(gomock.Any(), gomock.Any(), []string{models.HostStatusDisabled, models.HostStatusError}).
```
eranco74, line 1111: `models.HostStatusDisabled, models.HostStatusError` ??
Why use these host status, should we even add labels to hosts with such status?

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -1098,7 +1098,27 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				assistedController.UpdateNodeLabels(context.TODO(), &wg)
 				wg.Wait()
 			})
+			It("Set label only in Ready node", func() {
+				nodeLabels := `{"node.ocs.openshift.io/storage":""}`
+				k8sNodes := GetKubeNodes(kubeNamesIds)
```
eranco74, line 1103:
```suggestion
				notReadyk8sNodes := GetKubeNodes(kubeNamesIds)
```

src/assisted_installer_controller/assisted_installer_controller_test.go
eranco74, line 1116: I don't get it, in the first call  `ListNodes` return not ready nodes, in the second time it get called  `ListNodes` return ready nodes.
Why `PatchNodeLabels` get called 3 times?

src/assisted_installer_controller/assisted_installer_controller_test.go
tsorya: Those are statuses the we skip(ignore) so we should not bring those hosts at all

src/assisted_installer_controller/assisted_installer_controller_test.go
tsorya: it runs in a loop till all nodes get there labels.
1. Set labels of 2 hosts as they are ready and don't set for the third one as it is not
2. Set labels of the last host that we set ready

tsorya: The idea is to split logic and show that first time we will patch 2 nodes and on the next loop we will patch 1 node

src/assisted_installer_controller/assisted_installer_controller_test.go
tsorya: not right, we have only 1 not ready node and i pretty ok with calling it k8sNodes

src/assisted_installer_controller/assisted_installer_controller_test.go
eranco74: Hmmm, I don't understand.
you have 3 nodes, 2 of them should be skipped due the the status in AI ([here](https://github.com/openshift/assisted-installer/pull/620#discussion_r1089900370))
That leaves one node, what am I missing?

src/assisted_installer_controller/assisted_installer_controller_test.go
tsorya: this test doesn't check anything with host statuses in assisted though we call GetHost with those statuses in function

src/assisted_installer_controller/assisted_installer_controller_test.go
tsorya: See the function and all other tests, this test doesn't validate any status in assisted and calls GetHosts with statuses exactly the same as function do

src/assisted_installer_controller/assisted_installer_controller_test.go
eranco74: If hosts with these status should be skipped why this test is creating nodes with these statuses and check that they get updated???
I understand that what this test is covering isn't that nodes with these statuses get skipped, but why create misleading nodes entity, why not create a test that checks a valid case? 
/hold

tsorya: Please show me where it creates nodes in this statuses?

eranco74: IT's not creating it's calling `GetHosts`

tsorya: https://github.com/openshift/assisted-installer/blob/master/src/assisted_installer_controller/assisted_installer_controller.go#L1033

In order to run test we need to mock all the calls to all the functions inside tested function please see attached link that explains why this mock is expected here

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -1098,7 +1098,34 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
+					}
+					notReadyNodeName = k8sNodesWith1NotReady.Items[0].Name
+				}
+
+				hosts := create3Hosts(models.HostStatusInstalled, models.HostStageDone, nodeLabels)
+				mockbmclient.EXPECT().GetHosts(gomock.Any(), gomock.Any(), []string{models.HostStatusDisabled, models.HostStatusError}).
+					Return(hosts, nil).Times(2)
+				// first run with 2 ready nodes
+				mockk8sclient.EXPECT().ListNodes().Return(k8sNodesWith1NotReady, nil).Times(1)
+				// Set 3rd node as ready
```
eranco74, line 1120:
```suggestion
				// Return all nodes as ready
```

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -1098,7 +1098,34 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
+				mockk8sclient.EXPECT().ListNodes().Return(GetKubeNodes(kubeNamesIds), nil).Times(1)
+				mockk8sclient.EXPECT().PatchNodeLabels(gomock.Any(), nodeLabels).Return(nil).Times(2)
+				mockk8sclient.EXPECT().PatchNodeLabels(notReadyNodeName, nodeLabels).Return(nil).Times(1)
```
eranco74, line 1123: verify the order with `gomock.InOrder`

tsorya: no need for it, there is not order here. it is string and not list

eranco74 (commented):
/lgtm

PR #623, opened 2023-02-05
MGMT-13568: Add pause mcp while setting new role labels

tsorya:
Adding pause machine config pool while adding role labels to hosts


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1046,17 +1049,47 @@ func (c *controller) updateNodesLabels() bool {
...
+		}
+	}
+
+	if len(hostsWithLabels) > c.patchNodesLabels(c.log, hostsWithLabels) {
 		return KeepWaiting
-	} else {
-		return ExitWaiting
 	}
+
+	if pauseMCO {
```
ori-amizur, line 1084: If someone clears the node labels while the MCP is already paused, then it will remain paused. Right ?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1046,17 +1049,47 @@ func (c *controller) updateNodesLabels() bool {
...
+	}
+	return hostsWithLabels, strings.Contains(allNodeLabelsAsOneString, roleLabel), nil
+}
+
+// Set labels to hosts
+// If one of the labels was role pause mcp and set this role
+// after setting all labels unpause in case it was paused
+func (c *controller) setNodesLabels() bool {
+	c.log.Infof("Setting node labels if require")
+	hostsWithLabels, pauseMCO, err := c.getAssistedHostsWithLabels()
```
ori-amizur, line 1062: We should differentiate between labels on master nodes or worker nodes.

ori-amizur, line 1062: The decision if to pause/unpause should mostly rely on the machine_config_pool_name.  If not then there is no reason to do it.

tsorya: It is difficult decision as pool can have different names from role as far as i understand and just have relevant labels. I actually don't think we should go and list all the pools in order to check if one of them have relevant labels inside. 
Not sure it worth such complication

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: Clears from assisted service? I believe we don't allow such api calls in the middle of installation

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: Agree no need to pause mcp if master have new role, i believe it is not healthy to set custom roles on masters though

tsorya: Currently will check only workers in order to decide

ori-amizur: If exists(machine_config_pool_name) != "" ?

ori-amizur: Checking machine_config_pool_name is an indication that the user actually wanted to use different MCP.  Otherwise it can be just a label.

tsorya: But do it requires same pool name as role? if not it is not going to be that easy. Though we can check that there is more than 2 mcps. To say the truth i don't see any problem in pausing mco as it does nothing if we unpause it at the end

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: actually we can always unpause though i think this case can't really happened.

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: I think we should not complicate it now and see how it is going.

romfreiman:
jira please

tsorya:
> jira please

It is draft, don't push it now. The idea was to show it

romfreiman:
is it for day1 only? or day2 as well?

tsorya:
> is it for day1 only? or day2 as well?

Day1, day2 will have familiar logic in one of the controllers

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1057,17 +1060,53 @@ func (c *controller) updateNodesLabels() bool {
 			continue
 		}
 		hostsWithLabels[hostname] = hostData
+		// if host is worker and new role was set we need to pause MCP
+		if hostData.Host.Role == models.HostRoleWorker && strings.Contains(hostData.Host.NodeLabels, roleLabel) {
```
ori-amizur, line 1064: MCP has field match labels.  It contains not just roles.
We should find out if any of the requested labels by the MCP are defined on the node.

tsorya: We can do it another way, pause worker pool if host has mcp pool and not role label
WDYT?

ori-amizur (approved):

PR #626, opened 2023-02-06
MULTIARCH-3387: Set boot device to target disk for powervm

cs-zhang:
For powervm, using the bootlist to set target disk as boot device.

Signed-off-by: CS Zhang <zhangcho@us.ibm.com>

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -182,7 +182,8 @@ func (o *ops) SetBootOrder(device string) error {
 	}
 
 	o.log.Infof("SetBootOrder, runtime.GOARCH: %s, device: %s", runtime.GOARCH, device)
-	if runtime.GOARCH == "ppc64le" {
+	_, err := o.ExecPrivilegeCommand(nil, "test", "-f", "/usr/sbin/bootlist")
+	if err == nil {
 		_, err := o.ExecPrivilegeCommand(o.logWriter, "bootlist", "-m", "normal", "-o", device)
 		if err != nil {
 			o.log.WithError(err).Errorf("Failed to set boot disk with bootlist. Skipping...")
@@ -190,7 +191,7 @@ func (o *ops) SetBootOrder(device string) error {
 		return nil
 	}
 
-	_, err := o.ExecPrivilegeCommand(nil, "test", "-d", "/sys/firmware/efi")
+	_, err = o.ExecPrivilegeCommand(nil, "test", "-d", "/sys/firmware/efi")
 	if err != nil {
 		o.log.Info("setting the boot order on BIOS systems is not supported. Skipping...")
 		return nil
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -184,7 +184,7 @@ func (o *ops) SetBootOrder(device string) error {
 	o.log.Infof("SetBootOrder, runtime.GOARCH: %s, device: %s", runtime.GOARCH, device)
 	_, err := o.ExecPrivilegeCommand(nil, "test", "-f", "/usr/sbin/bootlist")
 	if err == nil {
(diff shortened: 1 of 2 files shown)


src/ops/ops.go
```diff
@@ -181,6 +181,15 @@ func (o *ops) SetBootOrder(device string) error {
 		return nil
 	}
 
+	o.log.Infof("SetBootOrder, runtime.GOARCH: %s, device: %s", runtime.GOARCH, device)
+	if runtime.GOARCH == "ppc64le" {
+		_, err := o.ExecPrivilegeCommand(o.logWriter, "bootlist", "-m", "normal", "-o", device)
+		if err != nil {
+			o.log.WithError(err).Errorf("Failed to set boot disk with bootlist. Skipping...")
+		}
+		return nil
```
omertuc, line 190: Do all ppc64le architecture servers have the bootlist command, or just particular IBM machines? If it's the latter, maybe we should only `return nil` within an `else` statement

cs-zhang: All RHCOS for ppc64le has the bootlist command.

src/ops/ops.go
```diff
@@ -181,6 +181,15 @@ func (o *ops) SetBootOrder(device string) error {
 		return nil
 	}
 
+	o.log.Infof("SetBootOrder, runtime.GOARCH: %s, device: %s", runtime.GOARCH, device)
+	if runtime.GOARCH == "ppc64le" {
```
gamli75, line 185: What about s390x? Do we need to handle it also? Do we have another task for it?

src/ops/ops.go
```diff
@@ -181,6 +181,15 @@ func (o *ops) SetBootOrder(device string) error {
 		return nil
 	}
 
+	o.log.Infof("SetBootOrder, runtime.GOARCH: %s, device: %s", runtime.GOARCH, device)
```
osherdp, line 184: I'm pretty sure ``runtime.GOARCH`` applies to complie-time (despite its name containing ``runtime``)
I guess it's not too critical, but if it's easily possible to use real runtime information it's better

osherdp, line 184: This change should come with a matching unit test. Where is it? :smiley:

osherdp (changes requested):
Changes look good, raising just two points

src/ops/ops.go
cs-zhang: Let me check with Z  team about it.

cs-zhang: Z system does not has such cmd to set boot order inside RHCOS.

src/ops/ops.go
cs-zhang: Build time and run time CPU arch have to be the same, no code can build on one platform and run on other platform.

osherdp: Right, that's exactly the problem. If someone in the future removes those lines you're adding here (or does any other change) this PR is going to get merged without anything to stop it
Can you please add tests?

src/ops/ops_test.go
```diff
@@ -313,3 +314,25 @@ var _ = Describe("RemoveAllDMDevicesOnDisk", func() {
...
+		conf     *config.Config
+	)
+
+	BeforeEach(func() {
+		ctrl = gomock.NewController(GinkgoT())
+		execMock = execute.NewMockExecute(ctrl)
+		conf = &config.Config{}
+	})
+
+	It("Set boot order", func() {
```
osherdp, line 331: from what I'm seeing tests are passing even without your changes in ``ops.go``
what am I missing?

cs-zhang: I have tested on Power if remove my changes, the test will fail, it is platform dependent. Also it passed on amd system.

osherdp (approved):

osherdp:
/hold
need to fix lint

osherdp (approved):
/unhold

PR #628, opened 2023-02-07
MGMT-13595: Allow moving node from rebooting to joined

tsorya:
[MGMT-13595](https://issues.redhat.com//browse/MGMT-13595): Allow moving node from rebooting to joined
Sometimes we are missing configuring state due to different issue and installaiton will be blocked, we want to allow controller to move node from rebooting to joined in order to continue


filanov:
/hold 
until 2.7 GA

PR #629, opened 2023-02-08
MGMT-13605: Remove uninitialized taint from vSphere nodes

vrutkovs:
Since 4.13 cloud controller is external and requires valid
vSphere credentials  to remove uninitialized taint, which blocks
bootstrap. This change will make assisted-installer watch nodes and remove
the taint once it appears to complete installation - both on bootstrap and post-control-plane.

Verified that installed no longer gets stuck on latest nightly, needs https://github.com/openshift/release/pull/36133 to make sure it works in CI

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -191,7 +191,9 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
 // it will update joined/done status
 // approve csr will run as routine and cancelled whenever all nodes are ready and joined
 // this will allow to run it and end only when it is needed
-func (c *controller) WaitAndUpdateNodesStatus(ctx context.Context, wg *sync.WaitGroup) {
+// removeUninitializedTaint parameter is removing unitialized taint on platforms with
+// external cloud provider, as assisted-installer is using fake credentials
+func (c *controller) WaitAndUpdateNodesStatus(ctx context.Context, wg *sync.WaitGroup, removeUninitializedTaint bool) {
 	approveCtx, approveCancel := context.WithCancel(ctx)
 	defer func() {
 		approveCancel()
@@ -202,12 +204,13 @@ func (c *controller) WaitAndUpdateNodesStatus(ctx context.Context, wg *sync.Wait
 	go c.ApproveCsrs(approveCtx)
 
 	c.log.Infof("Waiting till all nodes will join and update status to assisted installer")
-	_ = utils.WaitForPredicateWithContext(ctx, LongWaitTimeout, GeneralWaitInterval, c.waitAndUpdateNodesStatus)
+	_ = utils.WaitForPredicateParamsWithContext(ctx, LongWaitTimeout, GeneralWaitInterval, c.waitAndUpdateNodesStatus, removeUninitializedTaint)
 }
 
-func (c *controller) waitAndUpdateNodesStatus() bool {
+func (c *controller) waitAndUpdateNodesStatus(arg interface{}) bool {
 	ignoreStatuses := []string{models.HostStatusDisabled}
 	var hostsInError int
+	removeUninitializedTaint := arg.(bool)
 	ctxReq := utils.GenerateRequestContext()
 	log := utils.RequestIDLogger(ctxReq, c.log)
 
@@ -258,7 +261,15 @@ func (c *controller) waitAndUpdateNodesStatus() bool {
 			}
 		}
 
-		if common.IsK8sNodeIsReady(node) && host.Host.Progress.CurrentStage != models.HostStageDone {
+		ready := common.IsK8sNodeIsReady(node)
+		if !ready && removeUninitializedTaint {
+			if err := c.kc.UntaintNode(node.Name); err != nil {
+				log.WithError(err).Errorf("Failed to remove uninitialized taint from node %s", node.Name)
+				continue
+			}
+		}
+
+		if ready && host.Host.Progress.CurrentStage != models.HostStageDone {
 			log.Infof("Found new ready node %s with inventory id %s, kubernetes id %s, updating its status to %s",
 				node.Name, host.Host.ID.String(), node.Status.NodeInfo.SystemUUID, models.HostStageDone)
 
@@ -1143,7 +1154,7 @@ func (c controller) logHostResolvConf() {
 }
 
 // This is temporary function, till https://bugzilla.redhat.com/show_bug.cgi?id=2097041 will not be resolved,
-//that should validate router state only in case of failure
+// that should validate router state only in case of failure
 // It will patch router to add access logs
 // It will run http router health check to see if router is healthy on host network
 func (c controller) logRouterStatus() {
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -210,7 +210,7 @@ func (i *installer) finalize() error {
 	return nil
 }
 
-//updateSingleNodeIgnition will download the host ignition config and add the files under storage
+// updateSingleNodeIgnition will download the host ignition config and add the files under storage
 func (i *installer) updateSingleNodeIgnition(singleNodeIgnitionPath string) error {
 	if i.DryRunEnabled {
 		return nil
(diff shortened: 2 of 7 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -191,7 +191,7 @@ func logHostsStatus(log logrus.FieldLogger, hosts map[string]inventory_client.Ho
 // it will update joined/done status
 // approve csr will run as routine and cancelled whenever all nodes are ready and joined
 // this will allow to run it and end only when it is needed
-func (c *controller) WaitAndUpdateNodesStatus(ctx context.Context, wg *sync.WaitGroup) {
+func (c *controller) WaitAndUpdateNodesStatus(ctx context.Context, wg *sync.WaitGroup, removeUninitializedTaint bool) {
```
gamli75, line 194: please add an explanation for this parameter in the comment above the method

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -202,12 +202,13 @@ func (c *controller) WaitAndUpdateNodesStatus(ctx context.Context, wg *sync.Wait
 	go c.ApproveCsrs(approveCtx)
 
 	c.log.Infof("Waiting till all nodes will join and update status to assisted installer")
-	_ = utils.WaitForPredicateWithContext(ctx, LongWaitTimeout, GeneralWaitInterval, c.waitAndUpdateNodesStatus)
+	_ = utils.WaitForPredicateParamsWithContext(ctx, LongWaitTimeout, GeneralWaitInterval, c.waitAndUpdateNodesStatus, removeUninitializedTaint)
 }
 
-func (c *controller) waitAndUpdateNodesStatus() bool {
+func (c *controller) waitAndUpdateNodesStatus(arg interface{}) bool {
```
eranco74, line 208:
```suggestion
func (c *controller) waitAndUpdateNodesStatus(removeUninitializedTaint bool)) bool {
```

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -258,6 +259,13 @@ func (c *controller) waitAndUpdateNodesStatus() bool {
...
 		}
 
+		if !common.IsK8sNodeIsReady(node) && removeUninitializedTaint {
+			if err := c.kc.UntaintNode(node.Name); err != nil {
+				log.WithError(err).Errorf("Failed to remove uninitialized taint from node %s", node.Name)
+				continue
+			}
+		}
+
 		if common.IsK8sNodeIsReady(node) && host.Host.Progress.CurrentStage != models.HostStageDone {
```
eranco74, line 269:
```suggestion
                ready := common.IsK8sNodeIsReady(node)
		if !ready && removeUninitializedTaint {
			if err := c.kc.UntaintNode(node.Name); err != nil {
				log.WithError(err).Errorf("Failed to remove uninitialized taint from node %s", node.Name)
				continue
			}
		}

		if ready && host.Host.Progress.CurrentStage != models.HostStageDone {
```

src/installer/installer.go
```diff
@@ -586,6 +586,11 @@ func (i *installer) waitForMasterNodes(ctx context.Context, minMasterNodes int,
 			i.log.Warnf("Still waiting for master nodes: %v", err)
 			return false
 		}
+		for _, node := range nodes.Items {
+			if err = kc.UntaintNode(node.Name); err != nil {
+				i.log.Warnf("Failed to untaint node %s: %v", node.Name, err)
+			}
+		}
```
eranco74, line 593: I think we want to do this only for not Ready master nodes.
* We are checking if they are ready in the next function call

src/k8s_client/k8s_client.go
```diff
@@ -627,3 +629,42 @@ func (c *k8sClient) IsClusterCapabilityEnabled(capability configv1.ClusterVersio
...
+	node, err := c.client.CoreV1().Nodes().Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	foundTaintIndex := -1
+	for i, taint := range node.Spec.Taints {
+		if taint.Key == UNINITIALIZED_TAINT_KEY {
+			foundTaintIndex = i
+		}
+	}
```
eranco74, line 648:
```suggestion
	uninitialized := slices.Contains(node.Spec.Taints, UNINITIALIZED_TAINT_KEY)
```
https://pkg.go.dev/k8s.io/utils/strings/slices

eranco74 (commented):
Overall looks good.
Just wondering won't the cloud controller just keep adding back the uninitialized taint?
Have we considered disabling the cloud controller in case of vsphere platform? (we can instruct the users to enable it once they want to add the creds)

src/assisted_installer_controller/assisted_installer_controller.go
vrutkovs: it has to be `arg interface{}` since we're calling it via `utils.WaitForPredicateParamsWithContext` (see similar `waitForCSVBeCreated`)

src/k8s_client/k8s_client.go
vrutkovs: 1. This needs to be an exact Taint match (including value and operator)
2. Later on we need the found taint index to remove it via JSON patch

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: Not better to have as different go routine? or we want to block node from being Done in assisted till we don't remove this taint?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -258,7 +261,15 @@ func (c *controller) waitAndUpdateNodesStatus() bool {
 			}
 		}
 
-		if common.IsK8sNodeIsReady(node) && host.Host.Progress.CurrentStage != models.HostStageDone {
+		ready := common.IsK8sNodeIsReady(node)
+		if !ready && removeUninitializedTaint {
```
tsorya, line 265: It will not become ready till we will not remove it?

src/k8s_client/k8s_client.go
```diff
@@ -627,3 +629,42 @@ func (c *k8sClient) IsClusterCapabilityEnabled(capability configv1.ClusterVersio
...
+	if foundTaintIndex == -1 {
+		return nil
+	}
+	c.log.Infof("Removing uninitialized taint on node %s", name)
+	patch := []taintRemovePatch{{
+		Op:   "remove",
+		Path: fmt.Sprintf("/spec/taints/%d", foundTaintIndex),
+	}}
+	c.log.Debug(patch)
```
tsorya, line 657: Lets make message a little more readable even if it is debug

src/assisted_installer_controller/assisted_installer_controller.go
vrutkovs: It won't become ready with the taint (CNI won't schedule there)

src/k8s_client/k8s_client.go
vrutkovs: Good idea

src/assisted_installer_controller/assisted_installer_controller.go
vrutkovs: That'd increase a number of Nodes.Get() calls, is it really necessary?

The node won't reach Ready status until the taint is not removed (usually cloud-controller removes it but we have fake creds and have to do it ourselves)

osherdp:
@eranco74 @tsorya @gamli75 reviewers, anything else here?

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -147,7 +147,11 @@ func main() {
 		logger.Infof("Finished all")
 	}()
 
-	go assistedController.WaitAndUpdateNodesStatus(mainContext, &wg)
+	removeUninitializedTaint := false
+	if cluster.Platform != nil && *cluster.Platform.Type == models.PlatformTypeVsphere {
```
gamli75, line 151: I think we should add a condition about OCP version >= 4.13

gamli75:
> @eranco74 @tsorya @gamli75 reviewers, anything else here?

added 1 comment

src/main/assisted-installer-controller/assisted_installer_main.go
vrutkovs: This taint is used by external cloud controller only, so it will not appear in 4.12 or less

src/k8s_client/k8s_client.go
eliorerz, line 648:
```suggestion
	for i, taint := range node.Spec.Taints {
		if taint.Key == UNINITIALIZED_TAINT_KEY {
			foundTaintIndex = i
			break
		}
	}
```

src/k8s_client/k8s_client.go
```diff
@@ -627,3 +629,42 @@ func (c *k8sClient) IsClusterCapabilityEnabled(capability configv1.ClusterVersio
...
+	if err != nil {
+		return err
+	}
+	foundTaintIndex := -1
+	for i, taint := range node.Spec.Taints {
+		if taint.Key == UNINITIALIZED_TAINT_KEY {
+			foundTaintIndex = i
+		}
+	}
+	if foundTaintIndex == -1 {
```
eliorerz, line 649: Maybe log here that we couldn't find the Taint

vrutkovs: Done

PR #646, opened 2023-03-02
NO-ISSUE: controller logs optimizations

tsorya:
NO-ISSUE: In case upload of must-gather fails we need to retry to upload
    and not to gather it one more time as it has less chance to succeed
    In case pod name doesn't exist add one in order to upload file
NO-ISSUE: Adding log operator status conditions


eranco74 (commented):
/lgtm

omertuc:
[Note that this caused a regression in signature behavior](https://github.com/openshift-assisted/assisted-installer-deployment/pull/387)

PR #660, opened 2023-04-18
MGMT-14299: Limit untaint nodes to vsphere/nutanix

adriengentil:
In the context of installing OCP on Oracle Cloud, we want to wait for
the Cloud Controller Manager to untaint the nodes, so they get labelled
properly at installation time.

Unlike vsphere and nutanix, the Oracle CCM gets the authentication token
through the instance metadata, it should just work as soon as it is
deployed on the nodes.

This change limits the untaint of the boostrap nodes to vsphere and
nutanix platforms.

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -417,7 +417,14 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 	}
 	i.UpdateHostInstallProgress(models.HostStageWaitingForControlPlane, waitingForMastersStatusInfo)
 
-	if err = i.waitForMinMasterNodes(ctx, kc); err != nil {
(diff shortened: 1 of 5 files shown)


src/installer/installer.go
```diff
@@ -417,7 +417,14 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 	}
 	i.UpdateHostInstallProgress(models.HostStageWaitingForControlPlane, waitingForMastersStatusInfo)
 
-	if err = i.waitForMinMasterNodes(ctx, kc); err != nil {
+	cluster, callErr := i.inventoryClient.GetCluster(ctx, false)
```
adriengentil, line 420: Not sure if calling back the service to get cluster details is the right way, let me know

vrutkovs (approved):
/approve

PR #670, opened 2023-05-03
MGMT-13586: Wait for ETCD Bootstrap to complete

CrystalChun:
https://issues.redhat.com/browse/MGMT-13586
There have previously been issues where the bootstrap node will reboot before ETCD is ready. This change waits for the new status provided by the ETCD operator before rebooting the bootstrap node.

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -429,6 +429,10 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 	}
 
 	i.waitForBootkube(ctx)
+	if err = i.waitForETCDBootstrap(ctx); err != nil {
+		i.log.Error(err)
+		return err
+	}
 
 	// waiting for controller pod to be running
 	if err = i.waitForController(kc); err != nil {
@@ -437,6 +441,23 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+func (i *installer) waitForETCDBootstrap(ctx context.Context) error {
+	i.UpdateHostInstallProgress(models.HostStageWaitingForBootkube, "waiting for ETCD bootstrap to be complete")
+	i.log.Infof("Started waiting for ETCD bootstrap to complete")
+	return utils.WaitForPredicate(waitForeverTimeout, generalWaitInterval, func() bool {
+		// check if ETCD bootstrap has completed every 5 seconds
+		if result, err := i.ops.ExecPrivilegeCommand(nil, "systemctl", "is-active", "progress.service"); result == "inactive" {
+			i.log.Infof("ETCD bootstrap progress service status: %s", result)
+			out, _ := i.ops.ExecPrivilegeCommand(nil, "systemctl", "status", "progress.service")
+			i.log.Info(out)
+			return true
+		} else if err != nil {
+			i.log.WithError(err).Warnf("error occurred checking ETCD bootstrap progress: %s", result)
+		}
+		return false
+	})
 }
 
 func numDone(hosts models.HostList) int {
--- a/src/installer/installer_test.go
+++ b/src/installer/installer_test.go
@@ -272,6 +272,15 @@ var _ = Describe("installer HostRoleMaster role", func() {
 			mockops.EXPECT().ExecPrivilegeCommand(gomock.Any(), "systemctl", "status", "bootkube.service").Return("1", nil).Times(1)
 		}
 
+		waitForETCDBootstrapSuccess := func() {
+			mockbmclient.EXPECT().UpdateHostInstallProgress(gomock.Any(), infraEnvId, hostId, models.HostStageWaitingForBootkube, "waiting for ETCD bootstrap to be complete").Return(nil).Times(1)
+			mockops.EXPECT().ExecPrivilegeCommand(gomock.Any(), "systemctl", "is-active", "progress.service").Return("inactive", nil).Times(1)
+		}
+
+		bootstrapETCDStatusSuccess := func() {
+			mockops.EXPECT().ExecPrivilegeCommand(gomock.Any(), "systemctl", "status", "progress.service").Return("1", nil).Times(1)
+		}
+
 		extractSecretFromIgnitionSuccess := func() {
 			mockops.EXPECT().ExtractFromIgnition(filepath.Join(InstallDir, bootstrapIgn), dockerConfigFile).Return(nil).Times(1)
 		}
@@ -317,6 +326,8 @@ var _ = Describe("installer HostRoleMaster role", func() {
 					WaitMasterNodesSucccess()
 					waitForBootkubeSuccess()
 					bootkubeStatusSuccess()
+					waitForETCDBootstrapSuccess()
+					bootstrapETCDStatusSuccess()
 					resolvConfSuccess()
 					waitForControllerSuccessfully(conf.ClusterID)
 					//HostRoleMaster flow:
@@ -346,6 +357,8 @@ var _ = Describe("installer HostRoleMaster role", func() {
 					WaitMasterNodesSucccess()
 					waitForBootkubeSuccess()
 					bootkubeStatusSuccess()
+					waitForETCDBootstrapSuccess()
+					bootstrapETCDStatusSuccess()
 					resolvConfSuccess()
 					waitForControllerSuccessfully(conf.ClusterID)
 					//HostRoleMaster flow:
@@ -400,6 +413,8 @@ var _ = Describe("installer HostRoleMaster role", func() {
 			WaitMasterNodesSucccess()
 			waitForBootkubeSuccess()
 			bootkubeStatusSuccess()
+			waitForETCDBootstrapSuccess()
+			bootstrapETCDStatusSuccess()
 			resolvConfSuccess()
 			waitForControllerSuccessfully(conf.ClusterID)
 			//HostRoleMaster flow:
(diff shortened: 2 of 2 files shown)


src/installer/installer.go
```diff
@@ -436,9 +437,46 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
...
+		i.log.WithError(err).Warnf("failed to create new version constraint")
+		return
+	}
+	if !constraints.Check(v1) {
+		i.log.Infof("Detected OCP version is %s, skipping wait for ETCD bootstrap, constraint: %s", i.OpenshiftVersion, constraints.String())
+		return
+	}
+	i.log.Infof("Detected OCP version is %s. Waiting for ETCD bootstrap to be complete", i.OpenshiftVersion)
+	//TODO: Should it use a pre-existing status or should a new one be created?
+	i.UpdateHostInstallProgress(models.HostStageWaitingForBootkube, "waiting for ETCD bootstrap to be complete")
```
CrystalChun, line 462: Could use feedback on this!

omertuc:
Please make sure `PrepareInstallerDryK8sMock`  is also ready to receive those new calls to the k8s API (which has to be mocked in dry-run mode)

src/common/common.go
```diff
@@ -92,6 +92,19 @@ func GetPodInStatus(k8Client k8s_client.K8SClient, podNamePrefix string, namespa
...
 }
 
+func IsETCDBootstrapComplete(k8Client k8s_client.K8SClient) error {
+	configMapName := "bootstrap"
+	configMapNamespace := "kube-system"
+	bootstrapConfigMap, err := k8Client.GetConfigMap(configMapNamespace, configMapName)
+	if err != nil {
+		return err
+	}
+	if bootstrapConfigMap.Data["status"] != "complete" {
```
filanov, line 102: Is that relevant for specific ocp verison?

eranco74:
Why are we "fixing" this issue here and not in `bootkube` or `wait_for_ceo`?

src/installer/installer.go
```diff
@@ -436,9 +437,46 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
...
+	if err != nil {
+		i.log.WithError(err).Warnf("failed to create new version with %s", i.OpenshiftVersion)
+		return
+	}
+	constraints, err := version.NewConstraint(">= 4.12")
+	if err != nil {
+		i.log.WithError(err).Warnf("failed to create new version constraint")
+		return
+	}
+	if !constraints.Check(v1) {
```
eranco74, line 456: What happens prior to 4.12?
how will we know etcd is ready?

src/installer/installer.go
eranco74: The stage seems wrong, this can stay unchanged if you move the `i.waitForETCD(ctx, kc)` to be called right after waitForBootkube  (line 433)

src/installer/installer.go
```diff
@@ -436,9 +437,46 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
...
+	//TODO: Should it use a pre-existing status or should a new one be created?
+	i.UpdateHostInstallProgress(models.HostStageWaitingForBootkube, "waiting for ETCD bootstrap to be complete")
+
+	for {
+		select {
+		case <-ctx.Done():
+			i.log.Infof("Wait for bootstrap ETCD has been cancelled")
+			return
+		case <-time.After(generalWaitInterval):
+			// check if bootstrap etcd configmap has been created every minute
```
eranco74, line 470: The comment seems misleading, the `generalWaitInterval` interval is 5 seconds not a minute

src/installer/installer.go
CrystalChun: Agreed about the wrong stage, but would waiting for etcd to be ready still be under waiting for bootkube?

src/installer/installer.go
CrystalChun: Thanks for catching that! Sorry meant to make it actually one minute

tsorya:
@eranco74 @CrystalChun in order to make it simpler and remove dependency on ocp version we can do it in 2 ways:
1. We can wait for [report-progress.service] (https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/report-progress.sh) to finish
2. Wait for configmap that this service creates.
I think waiting for service is enough though as it already verifies if configmap was created

src/installer/installer.go
```diff
@@ -436,9 +437,46 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 		return err
 	}
 
+	i.waitForETCD(ctx, kc)
```
filanov, line 440: where is the one minute wait that we have right now? 
This PR only added functionality, if i remember correctly @eranco74 you wanted to replace the one minute wait with a new logic no? 
I understand that we will need both because we have older version as well, just was thinking that it should be in the same place.

eranco74: Here is the one minute delay we currently have
https://github.com/openshift/assisted-installer/blob/adf62c69759137c6d351bcacc74464c867299adc/src/installer/installer.go#L202

CrystalChun: Do you all think we still need this delay?

src/common/common.go
```diff
@@ -92,6 +92,19 @@ func GetPodInStatus(k8Client k8s_client.K8SClient, podNamePrefix string, namespa
 	return nil
 }
 
+func IsETCDBootstrapComplete(k8Client k8s_client.K8SClient) error {
```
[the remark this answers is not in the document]
CrystalChun: Modified as suggested to wait for the service (report-progress.service)! How does it look?

CrystalChun:
> Why are we "fixing" this issue here and not in `bootkube` or `wait_for_ceo`?

My understanding is that it's a chicken and egg scenario where one is waiting for the other to finish and vice-versa
Comment here: https://issues.redhat.com/browse/OCPBUGS-5988?focusedId=21650341&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21650341

src/installer/installer.go
CrystalChun: Moved it to under waitForBootkube as suggested! How does it look? Does the stage make sense?

src/common/common.go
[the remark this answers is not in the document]
omertuc: I'm against the idea of involving the reporter service in critical flows and timing. The reporter service should be an independent, separate logging and telemetry service, free to do its own thing and change in the future without having to worry about being a load-bearing wall in our design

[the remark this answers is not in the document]
CrystalChun: I can see that, would waiting for the configmap be a better option, then? 
Alternatively, the service reports its progress to a json file, would it better to watch this or is it too similar to watching the service?
```
jq '.[] | select((.stage == "report-bootstrap-complete") and (.phase == "stage end"))' /var/log/openshift/progress.json
```

src/installer/installer.go
filanov: if we have this code i think that we can skip this minute, @eranco74 what do you think?

src/common/common.go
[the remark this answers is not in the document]
CrystalChun: @eranco74 what do you think?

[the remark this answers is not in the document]
omertuc: Forget all I said, after a conversation with @tsorya I realized this is a different reporter service than the one I was thinking about

CrystalChun:
@eranco74 may I have your review on my PR again?

src/installer/installer.go
```diff
@@ -429,16 +429,38 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
...
 		return err
 	}
-
 	return nil
 }
 
+func (i *installer) waitForETCDBootstrap(ctx context.Context) {
+	i.UpdateHostInstallProgress(models.HostStageWaitingForBootkube, "waiting for ETCD bootstrap to be complete")
+	i.log.Infof("Started waiting for ETCD bootstrap to complete")
+	for {
```
tsorya, line 445: lets use https://github.com/openshift/assisted-installer/blob/master/src/utils/utils.go#L221

CrystalChun: Thanks Igal! Modified it to use this function, how does it look?

src/installer/installer.go
```diff
@@ -439,6 +443,23 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 	return nil
 }
 
+func (i *installer) waitForETCDBootstrap(ctx context.Context) error {
+	i.UpdateHostInstallProgress(models.HostStageWaitingForBootkube, "waiting for ETCD bootstrap to be complete")
+	i.log.Infof("Started waiting for ETCD bootstrap to complete")
+	return utils.WaitForPredicate(5*time.Minute, generalWaitInterval, func() bool {
```
eranco74, line 449: Let's change this to:
```suggestion
	return utils.WaitForPredicate(waitForeverTimeout, generalWaitInterval, func() bool {
```
I know it shouldn't take that long but if there's a chance that etcd will complete the bootstrap after 10 minutes we don't want this timeout to be the reason the installation failed.
In general the installer isn't setting timeouts for the installation the assusted-service does.

src/installer/installer.go
```diff
@@ -439,6 +443,23 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 	return nil
 }
 
+func (i *installer) waitForETCDBootstrap(ctx context.Context) error {
+	i.UpdateHostInstallProgress(models.HostStageWaitingForBootkube, "waiting for ETCD bootstrap to be complete")
```
eranco74, line 447: :+1: 
nice

eranco74 (commented):
Looks good

src/installer/installer.go
CrystalChun: Makes sense! Accepted suggested changes. Thank you!

src/installer/installer.go
```diff
@@ -439,6 +443,23 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
...
+	i.log.Infof("Started waiting for ETCD bootstrap to complete")
+	return utils.WaitForPredicate(waitForeverTimeout, generalWaitInterval, func() bool {
+		// check if ETCD bootstrap has completed every 5 seconds
+		if result, err := i.ops.ExecPrivilegeCommand(nil, "systemctl", "is-active", "progress.service"); result == "inactive" {
```
CrystalChun, line 451: For history, we're waiting for progress.service to complete https://github.com/openshift/installer/blob/191e0707762555796dceb4ecd80f8032d4a664e7/data/data/bootstrap/files/usr/local/bin/report-progress.sh

Which is OCP version independent

PR #688, opened 2023-07-23
MGMT-15344: Assisted-controller should not timeout on waiting cvo by itself

tsorya:
[MGMT-15344](https://issues.redhat.com//browse/MGMT-15344): Assisted-controller should not timeout on waiting cvo by itself


eranco74 (commented):
/lgtm

PR #692, opened 2023-07-26
MGMT-15343: dependabot group updates

eifrach:


.github/dependabot.yml
```diff
@@ -13,13 +13,27 @@ updates:
...
+      go-dependencies:
+        patterns:
+          - "*"
 
   - package-ecosystem: "docker"
     directory: "/"
     schedule:
       interval: "weekly"
+    labels:
+      - "approved"
```
osherdp, line 29: here I'm not so sure we want to automatically approve. D/S will usually need to update as well
@adriengentil @gamli75 what are your thoughts?

adriengentil: I agree, the change of docker images must be coordinated with a CI config change

osherdp: actually I have a conspiracy theory that CI config will sometimes update based on the repo changes
not sure what's it based on :sweat_smile: 

but either way yes, the image changes are too infrastructural to my liking

osherdp: yep, auto-config-brancher does that: https://github.com/openshift/release/pull/41641

adriengentil: how do it work? https://github.com/openshift/release/pull/41641/files#diff-1bfd62ea788d34f33afa71e426ba5e0f21239201b17965eb5bb7e74cc760e19e

isn't it actually reflecting the image defined in the source repo?

osherdp: yes I think it actually looks at the source code 
though I'm not sure if it does it for all image-substitutions or not

osherdp (approved):

PR #697, opened 2023-08-08
MGMT-15398: Use ubi golang base image to fix multi-arch builds

CrystalChun:
https://issues.redhat.com/browse/MGMT-15398
The base image previously used did not actually contain gcc compilers for different architectures (other than amd64). Compiling with the flag CGO_ENABLED=1 (for fips) caused issues for arm64 and ppc64le since the C libs and gcc compiler for these architectures didn't exist.

This new base image actually has the correct gcc and glibc for each architecture.

### DIFF
--- a/Dockerfile.assisted-installer
+++ b/Dockerfile.assisted-installer
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.19 AS builder
 ARG TARGETPLATFORM
 ENV GOFLAGS=-mod=mod
 WORKDIR /go/src/github.com/openshift/assisted-installer


CrystalChun:
UBI libc version doesn't match centos stream8 version so we get this error running assisted-installer: (see journal in logs.tar of failing job [link](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-installer/697/pull-ci-openshift-assisted-installer-master-edge-e2e-ai-operator-ztp/1688962296581197824/artifacts/e2e-ai-operator-ztp/assisted-baremetal-operator-gather/artifacts/assisted-spoke-cluster_assisted-agent-cluster-install/))
```
Aug 08 18:55:11 extraworker-0 next_step_runne[2412]: time="08-08-2023 18:55:11" level=warning msg="Sending step <install-74f89e30> reply output <> error </usr/bin/installer: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /usr/bin/installer)\n/usr/bin/installer: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /usr/bin/installer)\n> exit-code <1>" file="step_processor.go:76" request_id=e03ee1d9-0cef-4b2e-993c-63de25d719b3

``` 

This an be fixed if we use centos stream9 BUT that won't be supported until ACM 2.10, see [slack thread](https://redhat-internal.slack.com/archives/CUPJTHQ5P/p1691600698454009?thread_ts=1662404854.877509&cid=CUPJTHQ5P)


There's also an issue with using UBI images 
https://issues.redhat.com/browse/MGMT-6635

PR #706, opened 2023-09-04
NO-ISSUE: dependabot exclude k8s

eifrach:
* fix for failing test on PR #705

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,3 +22,4 @@ updates:
           - "*"
         exclude-patterns:
           - "k8s.io/*"
+          - "sigs.k8s.io/*"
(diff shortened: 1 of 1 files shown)


eifrach:
/cc @adriengentil 
/cc @oshercc 

we should put a tracker to remove this fix in the future

.github/dependabot.yml
```diff
@@ -20,3 +20,5 @@ updates:
       go-dependencies:
         patterns:
           - "*"
+        exclude-patterns:
+          - "k8s.io/*"
```
adriengentil, line 24:
```suggestion
          - "k8s.io/*"
          - "sigs.k8s.io/*"
```

I think think one also needs to be added to avaoid the controller runtime to be upgraded

adriengentil (approved):

PR #713, opened 2023-09-06
MGMT-15710: Assisted installer changes for avoiding MCO reboot

ori-amizur:
iIf skip mco reboot is enabled (passed as argument), skip mco reboot is applied.  This logic consists from:
- Extracting the encapsulated machine config from ignition, that contains both the os-image and the kernel arguments needed for reboot avoidance.
- Calling ostree command to write os image with the expected kernel arguments.

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/openshift/assisted-installer/src/ops/execute"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
 	"github.com/go-openapi/swag"
 	"github.com/google/uuid"
@@ -148,6 +149,17 @@ func (i *installer) InstallNode() error {
 		//return err
 	}
 
+	if i.Config.Role == string(models.HostRoleWorker) {
+		// Wait for 2 masters to be ready before rebooting
+		if err = i.workerWaitFor2ReadyMasters(ctx); err != nil {
+			return err
+		}
+	}
+
+	if i.EnableSkipMcoReboot {
+		i.skipMcoReboot(ignitionPath)
+	}
+
 	if isBootstrap {
 		i.UpdateHostInstallProgress(models.HostStageWaitingForControlPlane, waitingForBootstrapToPrepare)
 		if err = bootstrapErrGroup.Wait(); err != nil {
@@ -158,13 +170,8 @@ func (i *installer) InstallNode() error {
 			return err
 		}
 		i.log.Info("Setting bootstrap node new role to master")
-
-	} else if i.Config.Role == string(models.HostRoleWorker) {
-		// Wait for 2 masters to be ready before rebooting
-		if err = i.workerWaitFor2ReadyMasters(ctx); err != nil {
-			return err
-		}
 	}
+
 	//upload host logs and report log status before reboot
 	i.log.Infof("Uploading logs and reporting status before rebooting the node %s for cluster %s", i.Config.HostID, i.Config.ClusterID)
 	i.inventoryClient.HostLogProgressReport(ctx, i.Config.InfraEnvID, i.Config.HostID, models.LogsStateRequested)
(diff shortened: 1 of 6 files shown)


src/ops/ops.go
```diff
@@ -917,3 +929,166 @@ func installerArgs(ignitionPath string, device string, extra []string) []string
...
+	b, err := os.ReadFile(ignitionPath)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to read ignition file %s", ignitionPath)
+	}
+	conf, _, err := config_latest.ParseCompatibleVersion(b)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to parse ignition file %s", ignitionPath)
+	}
+	source := ""
+	for i := range conf.Ignition.Config.Merge {
```
omertuc, line 943: Wouldn't this will fail in case of https://github.com/openshift/assisted-installer/blob/2e590cfa9d7052cbc3cdf601b167dcda4a348502/src/installer/installer.go#L130?

ori-amizur: This is not used for SNO.  Please see `getEmbeddedIgnition`

ori-amizur: Tested it with SNO. 
SNO has a merge section, but the source is not a pointer to another ignition.  The ignition is embedded as base64 encoding.

ori-amizur: Here it is merged fro SNO
https://github.com/openshift/assisted-installer/blob/2e590cfa9d7052cbc3cdf601b167dcda4a348502/src/installer/installer.go#L236

src/ops/ops.go
```diff
@@ -917,3 +929,166 @@ func installerArgs(ignitionPath string, device string, extra []string) []string
...
+		}
+	}
+	if source == "" {
+		return "", errors.Errorf("source not found for ignition %s", ignitionPath)
+	}
+	return source, nil
+}
+
+func (o *ops) getEmbeddedIgnition(source string) ([]byte, error) {
+	b, err := base64.StdEncoding.DecodeString(strings.Replace(source, "data:;base64,", "", 1))
```
omertuc, line 959: `data:blablabla` is actually a URL, with scheme `data`, you can use a dataurl library to parse it:

https://pkg.go.dev/github.com/vincent-petithory/dataurl

We use it in assisted-service so you can see how it's being used there

eranco74:
What's the use case for this?
Assisted-installer is executed from a live-iso and typically reboots the host once it's done so that the host will start as an OCP node. why would we want to avoid reboot?

omertuc:
> What's the use case for this? Assisted-installer is executed from a live-iso and typically reboots the host once it's done so that the host will start as an OCP node. why would we want to avoid reboot?

It avoids the extra reboot MCO does after downloading machine-os-content

omertuc:
We still do the "big" reboot from live to disk

omertuc:
See epic https://issues.redhat.com/browse/MGMT-13964

ori-amizur:
> why would we want to avoid reboot?

As an optimization.  Sometimes these reboots take long time.

src/installer/installer.go
```diff
@@ -243,8 +244,72 @@ func (i *installer) updateSingleNodeIgnition(singleNodeIgnitionPath string) erro
...
+		}
+	}
+	return ret
+}
+
+func (i *installer) getEncapsulatedMC(ctx context.Context, ignitionPath string, mc **mcfgv1.MachineConfig) func() error {
+	return func() error {
+		interval := 5 * time.Second
+		var err error
+		_ = utils.WaitForPredicateWithContext(ctx, 2*time.Hour, interval, func() bool {
```
eranco74, line 273: Why do we need to spend 2 hours here?
this is an optimization if it happens to delay the installation we should drop it and carry on the installation without it.
2 hours seems way too long for fetching the ignition config.

src/installer/installer.go
```diff
@@ -243,8 +244,72 @@ func (i *installer) updateSingleNodeIgnition(singleNodeIgnitionPath string) erro
...
+			*mc, err = i.ops.GetEncapsulatedMC(ignitionPath)
+			if err != nil {
+				i.log.WithError(err).Error("failed to get encapsulated MC")
+			}
+			return err == nil
+		})
+		if err != nil {
+			return fmt.Errorf("failed getting required kargs: %w", err)
+		}
```
eranco74, line 282: Unclear how this is related to the rest of the function

src/installer/installer.go
```diff
@@ -253,6 +318,40 @@ func (i *installer) writeImageToDisk(ignitionPath string) error {
...
+	}
+	return nil
+}
+
+func (i *installer) writeImageToDisk(ignitionPath string) error {
+	i.UpdateHostInstallProgress(models.HostStageWritingImageToDisk, "")
+
+	var err error
+	if i.EnableSkipMcoReboot {
+		err = i.writeReleaseOsImageToDisk(ignitionPath)
```
eranco74, line 347: In case this fail let's log it and publish an event but let's not fail the installation, instead we can just fall back to  `writeRunningOsImageToDisk`

src/ops/ops.go
```diff
@@ -917,3 +928,159 @@ func installerArgs(ignitionPath string, device string, extra []string) []string
...
+		Storage struct {
+			Files []struct {
+				Path     string
+				Contents struct {
+					Source string
+				}
+			}
+		}
+	}
+	err = json.Unmarshal(ignBytes, &ignition)
```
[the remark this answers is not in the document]
ori-amizur: Didn't work.
This is ignition 2.x so parsing fails.

src/installer/installer.go
ori-amizur: Actually for workers we might need to wait until bootstrapping complete.  The worker ignition needs to be retrieved from the control plane.

src/installer/installer.go
ori-amizur: The error is what received from the function GetEncapsulatedMC. The function WaitForPredicateWithContext does not return it.  This is the error that actually interests us.

src/ops/ops.go
```diff
@@ -917,3 +928,159 @@ func installerArgs(ignitionPath string, device string, extra []string) []string
...
+	var out bytes.Buffer
+	if _, err = io.CopyN(&out, r, math.MaxInt64); err != nil && err != io.EOF {
+		return nil, errors.Wrap(err, "getEmbeddedIgnition: failed to copy")
+	}
+	return out.Bytes(), nil
+}
+
+func (o *ops) getIgnitionFromBoostrap(source string) ([]byte, error) {
+	// #nosec G402
+	tr := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}
```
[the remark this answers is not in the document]
ori-amizur: OK

src/ops/ops.go
```diff
@@ -917,3 +930,178 @@ func installerArgs(ignitionPath string, device string, extra []string) []string
...
+
+func (o *ops) getMCSIgnition(source, ca string) ([]byte, error) {
+	if strings.HasPrefix(source, "data:") {
+		return o.getEmbeddedIgnition(source)
+	} else {
+		return o.getIgnitionFromBoostrap(source, ca)
+	}
+}
+
+func (o *ops) GetEncapsulatedMC(ignitionPath string) (*mcfgv1.MachineConfig, error) {
```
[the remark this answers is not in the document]
ori-amizur: Is it needed elsewhere?

src/installer/installer.go
tsorya: First line in writeReleaseOsImageToDisk is writeRunningOsImageToDisk , why we can't always run writeRunningOsImageToDisk first and only then run all other logic from writeRunningOsImageToDisk ?

PR #715, opened 2023-09-11
MGMT-15507: Set restricted list of approvers for protected release-ocm-2.9 branch

eliorerz:
/cc @filanov @gamli75 @romfreiman


filanov (changes requested):
the target branch should be release-ocm-2.9

PR #724, opened 2023-09-19
NO-ISSUE: fix dependable ignored packages

eifrach:
Move excluded to ignore list.

- excluded are excluded from the group only.
- ignored are removed from the dependency tree ( i.e. golang )


eifrach:
This will remove and close PR:
- #722
- #721
- #720

/cc @osherdp

osherdp (approved):

PR #725, opened 2023-09-19
MGMT-15810: change base image to stream9

osherdp:
Now that D/S building system fully supports RHEL-9 base images (and allows easy installation of nmstate and other libs that were previously unsupported) we can move U/S to stream9 and D/S to ubi9-minimal.

It doesn't seem possible to easily use ubi9-minimal U/S for the assisted-service, because installation of nmstate packages either requires entitlements running on a RHEL node, or alternatively an access to RedHat's internal network (which can be a major problem for U/S development). For a better consistency we're sticking with the same for the agent, installer, etc.

### DIFF
--- a/Dockerfile.assisted-installer-controller
+++ b/Dockerfile.assisted-installer-controller
@@ -17,9 +17,8 @@ RUN go mod download
 COPY . .
 RUN TARGETPLATFORM=$TARGETPLATFORM make controller
 
-FROM quay.io/centos/centos:stream8
+FROM quay.io/centos/centos:stream9
 
-RUN yum -y install make gcc unzip wget curl rsync && yum clean all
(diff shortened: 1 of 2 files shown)


Dockerfile.assisted-installer-controller
```diff
@@ -17,9 +17,8 @@ RUN go mod download
 COPY . .
 RUN TARGETPLATFORM=$TARGETPLATFORM make controller
 
-FROM quay.io/centos/centos:stream8
+FROM quay.io/centos/centos:stream9
 
-RUN yum -y install make gcc unzip wget curl rsync && yum clean all
```
osherdp, line 22: no idea why the controller needs all those packages. it doesn't seem like we're calling any of those executables

adriengentil (approved):

PR #727, opened 2023-09-21
MGMT-15833: Coerce OLM to reconcile subscriptions by re-annotating them

omertuc:
OLM often gives up on reconciling subscriptions. Today we have code in
the assisted-controller to coerce OLM on the assisted installed cluster
into reconciling subscriptions. We do it by deleting Jobs and deleting
InstallPlans.

Starting 4.14rc1, this doesn't seem to be enough. We noticed that we
also have to edit subscriptions to force OLM to reconcile them.

This commit adds code that would periodically add and remove a dummy
annotation from subscriptions to force OLM to reconcile them.

### DIFF
--- a/src/assisted_installer_controller/operator_handler.go
+++ b/src/assisted_installer_controller/operator_handler.go
@@ -2,6 +2,7 @@ package assisted_installer_controller
 
 import (
 	"context"
+	"fmt"
 	"time"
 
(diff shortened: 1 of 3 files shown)


omertuc:
Closing because this is just an example, I'm not intending to merge this myself

PR #739, opened 2023-10-20
OCPBUGS-20049: Remove uninitialized taint for agent-based installs

rwsu:
assisted-service runs on the bootstrap node in agent-based installations. assisted-service becomes unavailable after the bootstrap completes and when the bootstrap node reboots to join the control plane. When assisted-service becomes unavailable, assisted-installer-controller is then unable to retrieve the list of hosts. The call to retrieve the host list fails and the waitAndUpdateNodesStatus function breaks and retries again. This is a problem because the code that removes the taint happens further down in the function and is now unable to be exercised.

The consequence is the taint is not removed for the bootstrap node and any worker nodes.

The fix adds an alternative mechanism to remove the uninitialized taint for agent-based installs. This mechanism works without the presence of assisted-service.

### DIFF
--- a/src/main/assisted-installer-controller/assisted_installer_main.go
+++ b/src/main/assisted-installer-controller/assisted_installer_main.go
@@ -195,7 +195,7 @@ func main() {
 	// monitoring installation by cluster status
 	switch invoker {
 	case assistedinstallercontroller.InvokerAgent:
-		waitForInstallationAgentBasedInstaller(kc, logger)
+		waitForInstallationAgentBasedInstaller(kc, logger, removeUninitializedTaint)
 	default:
 		waitForInstallation(client, logger, assistedController.Status)
 	}
@@ -250,8 +250,27 @@ func waitForInstallation(client inventory_client.InventoryClient, log logrus.Fie
 // the inventory client because assisted-service is deployed on the bootstrap node
 // and when the bootstrap node reboots to join the cluster, assisted-service becomes
 // unavailable.
-func waitForInstallationAgentBasedInstaller(kubeClient k8s_client.K8SClient, log logrus.FieldLogger) {
+func waitForInstallationAgentBasedInstaller(kubeClient k8s_client.K8SClient, log logrus.FieldLogger, removeUninitializedTaint bool) {
 	for {
+		if removeUninitializedTaint {
+			// After the boostrap node has rebooted, assisted-service becomes unavailable
+			// and nodes can no longer be untainted through
+			// assistedController.WaitAndUpdateNodesStatus(mainContext, &wg, removeUninitializedTaint).
+			// because the call to retrieve hosts (c.ic.GetHosts(ctxReq, log, ignoreStatuses))
+			// will fail immediately.
+			// Here we have an alternative mechanism to untaint nodes when
+			// assisted-service is no longer available.
+			nodes, err := kubeClient.ListNodes()
+			if err != nil {
+				log.WithError(err).Error("Failed to get list of nodes from k8s client")
+			}
+			for _, node := range nodes.Items {
+				if err := kubeClient.UntaintNode(node.Name); err != nil {
+					log.WithError(err).Errorf("Failed to remove uninitialized taint from node %s", node.Name)
+					continue
+				}
+			}
+		}
 		clusterVersion, err := kubeClient.GetClusterVersion()
 		if err != nil {
 			log.WithError(err).Error("Failed to get cluster version from k8s client")
--- a/src/main/assisted-installer-controller/assisted_installer_main_test.go
+++ b/src/main/assisted-installer-controller/assisted_installer_main_test.go
@@ -1,19 +1,24 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"testing"
 	"time"
 
 	"github.com/go-openapi/swag"
 	"github.com/openshift/assisted-service/client/installer"
+	v1 "k8s.io/api/core/v1"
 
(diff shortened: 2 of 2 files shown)


src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -250,8 +250,32 @@ func waitForInstallation(client inventory_client.InventoryClient, log logrus.Fie
...
+			// After the boostrap node has rebooted, assisted-service becomes unavailable
+			// and nodes can no longer be untainted through
+			// assistedController.WaitAndUpdateNodesStatus(mainContext, &wg, removeUninitializedTaint).
+			// because the call to retrieve hosts (c.ic.GetHosts(ctxReq, log, ignoreStatuses))
+			// will fail immediately.
+			// Here we have an alternative mechanism to untaint nodes when
+			// assisted-service is no longer available.
+			nodes, err := kubeClient.ListNodes()
+			if err != nil {
+				log.WithError(err).Error("Failed to get list of nodes from k8s client")
```
bfournie, line 265: Should this return here on error instead of trying to use the nodes value below?

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -250,8 +250,32 @@ func waitForInstallation(client inventory_client.InventoryClient, log logrus.Fie
...
+			}
+			for _, node := range nodes.Items {
+				if taintIndex, err := kubeClient.HasUninitializedTaint(node.Name); taintIndex != -1 {
+					if err != nil {
+						log.WithError(err).Errorf("Checking if node %s contained uninitialized taint failed", node.Name)
```
bfournie, line 270: Should continue on error here instead of calling UntaintNode.

src/k8s_client/k8s_client.go
```diff
@@ -642,12 +643,12 @@ type taintRemovePatch struct {
 	Path string `json:"path"`
 }
 
-func (c *k8sClient) UntaintNode(name string) error {
+func (c *k8sClient) HasUninitializedTaint(name string) (int, error) {
```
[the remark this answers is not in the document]
rwsu: The index is used in UntaintNode. I'll rename the function to UninitializedTaintIndex. The original version I had was returning a bool but later realized I needed to pass the index to UntaintNode so I switched to in the index, but should have also renamed the function.

src/main/assisted-installer-controller/assisted_installer_main.go
rwsu: Thanks. A continue should be added here on error.

src/main/assisted-installer-controller/assisted_installer_main.go
rwsu: I think return should only happen to signal that the cluster has finished installing. On error, nodes.Items will be empty so should be ok. We still want the clusterVersion check to happen below.

src/k8s_client/k8s_client.go
```diff
@@ -642,12 +643,12 @@ type taintRemovePatch struct {
 	Path string `json:"path"`
 }
 
-func (c *k8sClient) UntaintNode(name string) error {
+func (c *k8sClient) UninitializedTaintIndex(name string) (int, error) {
```
CrystalChun, line 646: Sorry, I'm not understanding why we need this additional function to find the index, can't the full logic exist as it is in one function?

gamli75:
@vrutkovs you did something similar in the past - https://github.com/openshift/assisted-installer/pull/629/files 
can you review this PR?

src/k8s_client/k8s_client.go
zaneb: It's now called from both `UntaintNode()` and `waitForInstallationAgentBasedInstaller()`

rwsu: You're right, we don't need to have an additional function to find the index. I'll remove the new function and just use UntaintNode.

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -250,8 +250,27 @@ func waitForInstallation(client inventory_client.InventoryClient, log logrus.Fie
 // the inventory client because assisted-service is deployed on the bootstrap node
 // and when the bootstrap node reboots to join the cluster, assisted-service becomes
 // unavailable.
-func waitForInstallationAgentBasedInstaller(kubeClient k8s_client.K8SClient, log logrus.FieldLogger) {
+func waitForInstallationAgentBasedInstaller(kubeClient k8s_client.K8SClient, log logrus.FieldLogger, removeUninitialiedTaint bool) {
```
vrutkovs, line 253:
```suggestion
func waitForInstallationAgentBasedInstaller(kubeClient k8s_client.K8SClient, log logrus.FieldLogger, removeUninitializedTaint bool) {
```

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -250,8 +250,27 @@ func waitForInstallation(client inventory_client.InventoryClient, log logrus.Fie
...
 	for {
+		if removeUninitialiedTaint {
```
vrutkovs, line 255:
```suggestion
		if removeUninitializedTaint {
```

vrutkovs (changes requested):

src/main/assisted-installer-controller/assisted_installer_main.go
rwsu: Thanks. Fixed the spelling.

CrystalChun (approved):
/lgtm

PR #744, opened 2023-10-29
MGMT-15902: Trigger event that notifies the number of reboots once the node installation is completed

ori-amizur:
When installation stage moves to done, notify the assisted service with a specific event, the number of reboots that were required by a node in order to complete installation.

/cc @filanov


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1354,19 +1357,23 @@ func (c controller) parseMustGatherImages() []string {
...
+		log.Errorf("Failed to download noingress kubeconfig %v\n", err)
 		return "", err
 	}
-	c.log.Infof("Downloaded %s to %s.", kubeconfigFileName, kubeconfigPath)
+	log.Infof("Downloaded %s to %s.", kubeconfigFileName, kubeconfigPath)
 
 	return kubeconfigPath, nil
 }
 
+func (c controller) downloadKubeconfigNoingress(ctx context.Context, dir string) (string, error) {
```
tsorya, line 1373: no need to have it here at all, move it to common or utils

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1354,19 +1357,23 @@ func (c controller) parseMustGatherImages() []string {
 	return images
 }
 
-func (c controller) downloadKubeconfigNoingress(ctx context.Context, dir string) (string, error) {
+func downloadKubeconfigNoingress(ctx context.Context, dir string, ic inventory_client.InventoryClient, log logrus.FieldLogger) (string, error) {
```
tsorya, line 1360: no need to have it here at all, move it to common or utils

src/assisted_installer_controller/reboots_notifier.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (r *rebootsNotifier) getKubeconfigPath(ctx context.Context) (string, error) {
+	if r.kubeconfigPath != "" {
+		return r.kubeconfigPath, nil
+	}
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if r.kubeconfigPath == "" {
+		var err error
+		if r.kubeconfigPath, err = common.DownloadKubeconfigNoingress(ctx, os.TempDir(), r.ic, r.log); err != nil {
```
tsorya, line 57: Asking, not urgent. We do it at couple of places, maybe we can download it once to known path and use it from there? as you do it here?

ori-amizur: I yes, I don't think it should be part of this PR

PR #754, opened 2023-11-24
NO-ISSUE: make dockerfile multi-arch compatible

rccrdpccl:
cli image is not multiarch
security scans need rpm to validate oc binaries

### DIFF
--- a/Dockerfile.assisted-installer-controller
+++ b/Dockerfile.assisted-installer-controller
@@ -1,11 +1,9 @@
-FROM quay.io/openshift/origin-cli-artifacts:latest as cli-artifacts
-ARG TARGETPLATFORM
-RUN mkdir -p /usr/share/openshift/assisted
-RUN case $TARGETPLATFORM in "") dir=linux_amd64;; *) dir=`echo $TARGETPLATFORM | sed 's@/@_@'` ;; esac ;  ln /usr/share/openshift/${dir}/oc /usr/share/openshift/assisted
-
-FROM registry.ci.openshift.org/openshift/release:golang-1.20 AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20.10-2 AS builder
 ARG TARGETPLATFORM
 ENV GOFLAGS=-mod=mod
+
+USER root
+
 WORKDIR /go/src/github.com/openshift/assisted-installer
 
 # Bring in the go dependencies before anything else so we can take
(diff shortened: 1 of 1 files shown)


adriengentil:
let's use `registry.ci.openshift.org/openshift/release:golang-1.20` for the moment, and open a ticket to move to `registry.access.redhat.com/ubi9/go-toolset` here and in the other repos?

osherdp:
@rccrdpccl what did we decide to do here?

rccrdpccl:
sorry, I've totally lost track of this, I've created https://issues.redhat.com/browse/MGMT-16720 to keep track of it, will have a look soon because I totally forgot why I was doing this :sweat:

rccrdpccl:
in the meantime...
/retest

Dockerfile.assisted-installer-controller
```diff
@@ -19,7 +17,14 @@ RUN TARGETPLATFORM=$TARGETPLATFORM make controller
...
 FROM quay.io/centos/centos:stream9
 
-COPY --from=cli-artifacts /usr/share/openshift/assisted/oc /usr/bin/oc
+ENV OCP_VERSION=4.15-el8-beta
+RUN echo -e "\
+[mirror.openshift.com_pub_openshift-v4_multiarch_dependencies_rpms_${OCP_VERSION}_]\n\
+name=created by dnf config-manager from https://mirror.openshift.com/pub/openshift-v4/\$basearch/dependencies/rpms/${OCP_VERSION}/\n\
+baseurl=https://mirror.openshift.com/pub/openshift-v4/\$basearch/dependencies/rpms/${OCP_VERSION}/\n\
+enabled=1" >  /etc/yum.repos.d/mirror.openshift.com_pub_openshift-v4_${TARGETPLATFORM}_dependencies_rpms_${OCP_VERSION}_.repo &&\
+    dnf install --nogpgcheck -y openshift-clients && dnf clean all && rm -rf /var/cache
```
osherdp, line 26: @rccrdpccl I see that currently it cannot find `https://mirror.openshift.com/pub/openshift-v4/ppc64le/dependencies/rpms/4.15-el8-beta/repodata/repomd.xml` and that's the reason it fails

I guess we have two other options:
* Copy the static binaries instead of rpm installation, e.g. https://mirror.openshift.com/pub/openshift-v4/ppc64le/clients/ocp/4.15.0-rc.0/
* Use an older version. I see that this one is available for ppc: https://mirror.openshift.com/pub/openshift-v4/ppc64le/dependencies/rpms/4.12-el8-beta/

rccrdpccl: Thanks for the hint. I would avoid using binary if possible as it cannot be scanned by security scans, but I'll it take into account

danmanor:
@rccrdpccl How is this one going ? we have a perma-failing CI job that tries to build multi-arch images.

PR #757, opened 2023-12-07
MGMT-16266: Indication event showing how often host has been rebooted missing on some nodes

ori-amizur:
There are many possible failures.  Many of these failures are temporary such as API is not reachable, or failure to create debug pod. Added retry for the retrieval of number of reboots

/cc @filanov

### DIFF
--- a/src/assisted_installer_controller/reboots_notifier.go
+++ b/src/assisted_installer_controller/reboots_notifier.go
@@ -12,14 +12,18 @@ import (
 	"github.com/openshift/assisted-installer/src/common"
 	"github.com/openshift/assisted-installer/src/inventory_client"
 	"github.com/openshift/assisted-installer/src/ops"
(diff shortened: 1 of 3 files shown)


src/assisted_installer_controller/reboots_notifier.go
```diff
@@ -20,8 +21,12 @@ import (
 const (
 	eventName            = "reboots_for_node"
 	eventMessageTemplate = "Node %s has been rebooted %d times before completing installation"
+	getNumRebootsRetries = 10
+	getNumRebootsTimeout = 6 * time.Minute
 )
 
+var getNumRebootsRetrySleepDuration = 15 * time.Second
```
filanov, line 28: why keep it ourside of the const?

ori-amizur: I set it to short duration in unit tests

filanov (approved):

PR #758, opened 2023-12-08
MGMT-16370: fix golangci-lint failing to be pulled

adriengentil:
app-sre made their repo mirroring golangci-lint private for legal reasons, and
now our jobs are failing to install it. Installing it now from the cli.

We could also have pulled golangci-lint from dockerhub, but rate
limitation there is quite aggressive.


filanov (approved):

adriengentil:
/retitle MGMT-16370: fix golangci-lint failing to be pulled

carbonin:
I think this needs to be backported. Older version periodics are still failing on this.

PR #778, opened 2024-01-30
OCPBUGS-25718: Do not remove uninitialized taints if vSphere credentials are provided.

rwsu:
Beginning with OpenShift version 4.15, the agent-based installer allows vSphere credentials to be set. The credentials are passed through in the install-config overrides.
    
Because valid credentials can be used with vSphere, we do not want the assisted-installer or the assisted-installer-controller to remove the uninitialized taints when credentials are set in install-config.yaml because that prevents the vSphere CCM from initializing the nodes and set a proper providerID/UUID. If the providerID/UUID is not set in the node.spec, the vSphere CSI operator fails to install.

The RemoveUninitializedTaint function has been updated so that it returns false if the invoker is the agent-installer, the platform is vSphere, the OpenShift version is equal to or greater than 4.15, and valid credentials are set in the install-config.yaml.


src/installer/installer.go
```diff
@@ -483,14 +483,18 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 	}
 	i.UpdateHostInstallProgress(models.HostStageWaitingForControlPlane, waitingForMastersStatusInfo)
 
+	hasValidvSphereCredentials := common.HasValidvSphereCredentials(ctx, i.inventoryClient, i.log)
+	if hasValidvSphereCredentials {
+		i.log.Infof("hasValidvSphereCredentials: %v", hasValidvSphereCredentials)
```
patrickdillon, line 488: nit: with an info-level log should the msg be more user friendly? There are a few of these scattered around and thought they were perhaps leftover from debugging

patrickdillon (commented):
I'm not that familiar with this codebase, but the fix looks good to me

src/common/common_test.go
```diff
@@ -182,51 +183,148 @@ var _ = Describe("verify common", func() {
...
-			platformType := test.PlatformType
-			expectedRemoveUninitializedTaint := test.ExpectedRemoveUninitializedTaint
-
 			Context("by platform", func() {
-				var platform *models.Platform
-				BeforeEach(func() {
-					platform = &models.Platform{
-						Type: &platformType,
+				It(fmt.Sprintf("platform %v, invoker %v, hasValidCredentials %v, version %v, expected remove uninitialized taint = %v", test.PlatformType, test.Invoker, test.HasValidvSphereCredentials, test.VersionOpenshift, test.ExpectedRemoveUninitializedTaint), func() {
+					platform := &models.Platform{
```
adriengentil, line 253:
```suggestion
					platform := &models.Platform{
```

does this test loop works as expected? usually we need to assign test to some variable before using it because of the way ginkgo works. I suggest to move this test to use `DescribeTable` to avoid such issue.

rwsu: It didn't work as expected. Thanks for pointing out this issue. It has been fixed.

adriengentil (approved):

PR #779, opened 2024-02-04
MGMT-16211: Remove all timeouts from controller

ori-amizur:
Remove all timeouts from controller.  These timeouts are managed by assisted service. Instead, the controller waits forever for each activity that previously may have been timed out by the controller.

Add support to finalizing stages.  The controller reports the finalizing stages to the assisted service.

The controller may stop waiting if the assisted service moved to non-installing status, or in case of non-failing stages (mainly OLM stages), it stops waiting for stage to complete and moves to the next stage in case a timeouts on this stage occurs.


src/utils/utils.go
```diff
@@ -228,20 +230,18 @@ func WaitForPredicateWithTimer(ctx context.Context, timeout time.Duration, inter
...
+func WaitForever(ctx context.Context, interval time.Duration, predicate func() bool) error {
+	return WaitForPredicateWithContext(ctx, maxDuration, interval, predicate)
 }
 
-func WaitForPredicateParamsWithContext(ctx context.Context, timeout time.Duration, interval time.Duration, predicate func(arg interface{}) bool, arg interface{}) error {
-	return WaitForPredicateWithTimer(ctx, timeout, interval, func(timer *time.Timer) bool {
-		return predicate(arg)
-	})
+func WaitForPredicate(timeout time.Duration, interval time.Duration, predicate func() bool) error {
+	return WaitForPredicateWithContext(context.TODO(), timeout, interval, predicate)
```
tsorya, line 244: Lets provide context from outside

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -496,72 +521,95 @@ func (c controller) postInstallConfigs(ctx context.Context) (error, map[string]i
 	return nil, nil
 }
 
-func (c controller) waitForOLMOperators(ctx context.Context) error {
+func (c *controller) CancelFinalizingStage(stage models.FinalizingStage) {
+	c.stagesCancellers.cancelForStage(stage)
+}
+
+func (c *controller) finalizingStageCtx(parentCtx context.Context, stage models.FinalizingStage) context.Context {
```
tsorya, line 528: Can you provide some explanation for this one?Not sure i understand

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -200,15 +200,22 @@ func main() {
 	case assistedinstallercontroller.InvokerAgent:
 		waitForInstallationAgentBasedInstaller(kc, logger, removeUninitializedTaint)
 	default:
-		waitForInstallation(client, logger, assistedController.Status)
+		waitForInstallation(client, logger, assistedController)
+	}
+}
+
+func checkForTimedOutFinalizingStage(cluster *models.Cluster, controller assistedinstallercontroller.Controller, log logrus.FieldLogger) {
+	if cluster != nil && cluster.Progress != nil && cluster.Progress.FinalizingStageTimedOut {
```
tsorya, line 208: Something i am missing here, cluster is not in error state? Why we verify progress and not final cluster stage?

src/utils/utils.go
```diff
@@ -228,20 +230,18 @@ func WaitForPredicateWithTimer(ctx context.Context, timeout time.Duration, inter
...
+	return WaitForever(ctx, interval, func() bool {
+		return predicate(e)
+	})
 }
 
-func WaitForPredicateWithContext(ctx context.Context, timeout time.Duration, interval time.Duration, predicate func() bool) error {
-	return WaitForPredicateWithTimer(ctx, timeout, interval, func(timer *time.Timer) bool {
-		return predicate()
-	})
+func WaitForever(ctx context.Context, interval time.Duration, predicate func() bool) error {
```
tsorya, line 239: WaitForever is a little bit misleading as for me, we still wait for predicate no?

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -911,16 +925,23 @@ var _ = Describe("installer HostRoleMaster role", func() {
 				hosts := create3Hosts(models.HostStatusInstalled, models.HostStageDone, "")
 				mockbmclient.EXPECT().GetHosts(gomock.Any(), gomock.Any(), []string{models.HostStatusDisabled, models.HostStatusError}).
 					Return(hosts, nil).Times(0)
+				mockbmclient.EXPECT().UpdateFinalizingProgress(gomock.Any(), gomock.Any(), models.FinalizingStageWaitingForFinalizing).Times(1)
```
tsorya, line 928: maybe create function like successUpdateFinalizingProgress ?

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -938,11 +959,18 @@ var _ = Describe("installer HostRoleMaster role", func() {
 						validateClusterOperaotrReport(data)
 						return nil
 					}).Times(1)
+				mockbmclient.EXPECT().UpdateFinalizingProgress(gomock.Any(), gomock.Any(), models.FinalizingStageWaitingForFinalizing).Times(1)
+				mockbmclient.EXPECT().UpdateFinalizingProgress(gomock.Any(), gomock.Any(), models.FinalizingStageWaitingForClusterOperators).Times(1)
 
 				wg.Add(1)
-				go assistedController.PostInstallConfigs(context.TODO(), &wg)
+				go func() {
+					defer GinkgoRecover()
```
tsorya, line 967: nice

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -1065,12 +1116,65 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
-				assistedController.PostInstallConfigs(context.TODO(), &wg)
+				assistedController.PostInstallConfigs(ctx, &wg)
 				wg.Wait()
-				Expect(assistedController.Status.HasError()).Should(Equal(false))
-				Expect(assistedController.Status.GetOperatorsInError()).To(ContainElement("lso"))
+				Expect(assistedController.status.HasError()).Should(Equal(false))
+				Expect(assistedController.status.GetOperatorsInError()).To(ContainElement("lso"))
+			})
+
+			It("waiting for single OLM operator which soft timeouts", func() {
```
tsorya, line 1137: which -> with?

tsorya: Failure in OLM operators bring cluster to degraded state and not error, so why need soft timeouts on it?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -454,37 +478,38 @@ func (c controller) PostInstallConfigs(ctx context.Context, wg *sync.WaitGroup)
 
 	errMessage := ""
 	err, data := c.postInstallConfigs(ctx)
-	// context was cancelled, requires usage of WaitForPredicateWithContext
-	// no reason to set error
-	if ctx.Err() != nil {
-		return
-	}
+	// Even if the context was cancelled, the error is sent to the assisted service.
```
tsorya, line 481: Something i not fully understand, if context was cancelled it means that service ended installation process so why we need to send anything to it?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -454,37 +478,38 @@ func (c controller) PostInstallConfigs(ctx context.Context, wg *sync.WaitGroup)
...
+	// The timeouts will be handled by the service
 	if err != nil {
 		c.log.Error(err)
 		errMessage = err.Error()
-		c.Status.Error()
+		c.status.Error()
 	}
 	success := err == nil
+	if success {
+		c.updateFinalizingProgress(ctx, models.FinalizingStageDone)
```
tsorya, line 490: I think i not fully understand here too, if service is managing the process, why don't we have transition for this state in it? I think service should decide Done or not and not controller

src/assisted_installer_controller/assisted_installer_controller.go
ori-amizur: I think we still want to send the result of what was done until now.
Before context was cancelled only if timeout occurs.  Now even in case of timeout, context may be cancelled which will stop the controller.

src/assisted_installer_controller/assisted_installer_controller.go
ori-amizur: Finalizing stages are set by controller.  It is an indication what is the controller doing.

src/assisted_installer_controller/assisted_installer_controller.go
ori-amizur: There are some stages (today OLM stages), that timeout does not cause installation failure. Instead, the installation proceeds.
To accomplish it, when the controller monitors the cluster, it finds out if the current stage was timed out.
If it was timed out, the context us cancelled.
Since the context is also used by many others in the controller, a context for this stage was created separately and also can be cancelled separately.  This cancellation will not effect any other using the main context.

src/assisted_installer_controller/assisted_installer_controller_test.go
ori-amizur: OK

src/assisted_installer_controller/assisted_installer_controller_test.go
ori-amizur: Maybe the title was misleading

src/assisted_installer_controller/finalizing_stages_cancellers.go
```diff
@@ -0,0 +1,34 @@
+package assisted_installer_controller
+
+import (
+	"context"
+	"sync"
+
+	"github.com/openshift/assisted-service/models"
+)
+
+type finalizingStagesCancellers struct {
```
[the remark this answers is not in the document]
ori-amizur: For OLM stages we want to cancel the current stage but not the proceeding stage.

src/utils/utils.go
ori-amizur: OK.  Will change

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: Still not sure i understand. Context will be cancelled only in case service finished the installation and there is no point to actually send anything to service as it is done installation process. 
And actually nothing changed in logic here, this verification was added in order not to send error in case service decided to finish the installation and not to confuse with error from controller. I don't think we need to change this behavior

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -496,72 +517,92 @@ func (c controller) postInstallConfigs(ctx context.Context) (error, map[string]i
...
 	}()
 	// Get the monitored operators:
-	err = utils.Retry(maxFetchAttempts, FetchRetryInterval, c.log, func() error {
-		operators, err = c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID, c.OpenshiftVersion)
-		if err != nil {
-			return errors.Wrapf(err, "Error while fetch the monitored operators from assisted-service.")
+	c.updateFinalizingProgress(ctx, models.FinalizingStageWaitingForOlmOperatorsCsvInitialization)
+	retryErr = utils.WaitForeverForPredicateWithCancel(ctx, FetchRetryInterval, func() bool {
+		var err2 error
+		operators, err2 = c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID, c.OpenshiftVersion)
```
tsorya, line 554: why not to provide same context here?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -496,72 +517,92 @@ func (c controller) postInstallConfigs(ctx context.Context) (error, map[string]i
...
+	}
+}
+
+func (c *controller) waitForOLMOperators(ctx context.Context) error {
 	var operators []models.MonitoredOperator
-	var err error
+	var err, retryErr error
+	var errs []error
 
 	// In case the timeout occur, we have to update the pending OLM operators to failed state,
```
tsorya, line 536: still relevant? I don't understand a little bit here too, if all the timeouts are part of service so why controller should set failed on operator for timeout? I think that if we move all the timeout decisions to service we should handle operators timeout failure in service too.

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -496,72 +517,92 @@ func (c controller) postInstallConfigs(ctx context.Context) (error, map[string]i
...
+		if errs == nil {
 			return
 		}
-		updateFunc := func() bool {
-			if err = c.updatePendingOLMOperators(ctx); err != nil {
-				return KeepWaiting
-			}
-			return ExitWaiting
+		updateFunc := func() error {
+			return c.updatePendingOLMOperators(context.TODO())
```
tsorya, line 543: Not sure why not to use same context as in lline 545

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: I still think done can be set by service itself as it has all the needed data for it , don't see a need for it as next step we send sendCompleteInstallation, so maybe we don't need one of them?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -434,12 +443,23 @@ func isCsrApproved(csr *certificatesv1.CertificateSigningRequest) bool {
...
 }
 
-func (c controller) PostInstallConfigs(ctx context.Context, wg *sync.WaitGroup) {
+func (c *controller) updateFinalizingProgress(ctx context.Context, stage models.FinalizingStage) {
+	_ = utils.RetryWithContext(ctx, math.MaxInt, GeneralWaitInterval, c.log, func() error {
+		return c.ic.UpdateFinalizingProgress(ctx, c.ClusterID, stage)
+	})
+}
+
+func (c *controller) waitOnStage(ctx context.Context, stage models.FinalizingStage, interval time.Duration, predicate func() bool) error {
```
tsorya, line 452: Found name not very obvious as we set stage here and wait on predicate. Not sure which name is better though

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -434,12 +443,23 @@ func isCsrApproved(csr *certificatesv1.CertificateSigningRequest) bool {
...
+	return utils.WaitForeverForPredicate(ctx, interval, predicate)
+}
+
+func (c *controller) PostInstallConfigs(ctx context.Context, wg *sync.WaitGroup) {
 	defer func() {
 		c.log.Infof("Finished PostInstallConfigs")
 		wg.Done()
 	}()
-	err := utils.WaitForPredicateWithContext(ctx, LongWaitTimeout, GeneralWaitInterval, func() bool {
+	err := c.waitOnStage(ctx, models.FinalizingStageWaitingForFinalizing, GeneralWaitInterval, func() bool {
```
tsorya, line 462: This code runs in parallel with WaitAndUpdateNodesStatus as far as i remember.
From my point we don't need stage that says that we wait for stage. 
Can you explain why we actually need it?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -496,72 +517,92 @@ func (c controller) postInstallConfigs(ctx context.Context) (error, map[string]i
...
 	if err != nil {
-		return errors.Wrapf(err, "Timeout while waiting for OLM operators be installed")
-	}
+		err = errors.Wrap(err, "Stopped waiting for OLM operators be installed")
+		c.log.WithError(err).Warning("Installing OLM operators")
 
-	return nil
+		errs = append(errs, err)
+	}
+	return stderrors.Join(errs...)
```
tsorya, line 599: in case of no errors it returns nil?

src/assisted_installer_controller/assisted_installer_controller.go
ori-amizur: It might a bit technical, but timeouts are set on stages.  The assisted service needs to know that the controller is waiting on this stage.  This stage might be timed out if finalizing stage doesn't start after the timeout duration.
BTW, WaitAndUpdateNodesStatus might continue after finalizing started if there are some workers that are not done.

src/assisted_installer_controller/assisted_installer_controller.go
ori-amizur: With this change, context might also be cancelled in case of timeout expiration.

src/assisted_installer_controller/assisted_installer_controller.go
ori-amizur: OK

src/assisted_installer_controller/assisted_installer_controller.go
ori-amizur: OK. Will use

src/assisted_installer_controller/assisted_installer_controller.go
ori-amizur: For the same reason the controller notifies success.
And the service cannot decide on success unless the controller notifies (with same API call).

src/assisted_installer_controller/assisted_installer_controller.go
ori-amizur: Yes

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: But service on other hand decides that it got timeout so it should set timeout per operator and not to wait for controller to do it. It sort of service decides that timeout occurred and only then controller sends failed to operator cause service decided that operator failed. I am not sure i understand why in this case we need controller to send failed state. It can be next ticket but i really think if we already moving failure decision to service we should manage full failure flow from service and not to wait for controller to set it cause controller doesn't decide anything here, i actually would expect that service should decide success too by looking on operator status

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: Ok, so we get timeout that means that cluster stopped installation. I am not sure i understand why we cancel main context on timeout and not on cluster state as timeout should mean "failed or degraded" cluster state, no?

src/assisted_installer_controller/assisted_installer_controller.go
tsorya: I mean that assisted service can timeout on Finalizing stage itself, why it requires FinalizingStageWaitingForFinalizing ? Still not sure i understand, we don't have timeout on finalizing stage anymore? 
Still believe this stage is not needed and we don't need to timeout on it, we should timeout on regular finalizing stage and here to wait for cluster to go to finalizing stage. If we will get timeout cause 3 masters didn't join the cluster, cluster will go to error and context will be cancelled from main in all other cases we just should wait here.

PR #787, opened 2024-02-20
MGMT-16966: Problem creating extra partition on main disk in 4.15+

ori-amizur:
When skipping MCO reboot, the last partition is extended to fill the remaining free space in the disk. This may be undesirable since there are scenarios that a user adds a partition in the installation disk.

This change doubles the space occupied by the last disk partition instead of occupying all the remining space.  The assumption is that doubling the space should be enough for image overwrite.

/cc @javipolo

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -1100,17 +1100,67 @@ func (o *ops) ignitionPlatformId() string {
 	return id
 }
 
-func getPartition(device, partitionNumber string) string {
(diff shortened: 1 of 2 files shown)


romfreiman:
what's the behavior in coreos-installer?

ori-amizur:
> what's the behavior in coreos-installer?

No change for coreos-installer regarding this change

PR #788, opened 2024-02-21
MGMT-16843: Ensure valid hostname during install

CrystalChun:
https://issues.redhat.com/browse/MGMT-16843
Some processes during install time depend on the hostname and an invalid hostname (e.g. 'localhost', having capital letters, etc.) that could cause the install to fail.

Extra validation is added when checking the hostname to ensure it is valid. A random hostname will be assigned if the current hostname is invalid.

See above issue for an example of the scenario with an invalid hostname fails to install.

/cc @omertuc @gamli75

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -15,6 +15,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/openshift/assisted-installer/src/main/drymock"
 	"github.com/openshift/assisted-service/pkg/secretdump"
+	"github.com/openshift/assisted-service/pkg/validations"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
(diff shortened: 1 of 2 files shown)


gamli75:
@CrystalChun please add a test for hostname with Capital letters

gamli75:
@tsorya please review

src/installer/installer.go
```diff
@@ -935,24 +936,25 @@ func (i *installer) createSingleNodeMasterIgnition() (string, error) {
...
+	i.log.Infof("Start checking hostname")
 	hostname, err := i.ops.GetHostname()
 	if err != nil {
 		i.log.Errorf("Failed to get hostname from kernel, err %s157", err)
 		return err
 	}
-	if hostname != "localhost" {
-		i.log.Infof("hostname is not localhost, no need to do anything")
+
+	if err := validations.ValidateHostname(hostname); err == nil && hostname != "localhost" {
```
tsorya, line 951: maybe we can always write requested hostname?
Though this seems fine

PR #794, opened 2024-02-23
[release-4.14] MGMT-16843: Ensure valid hostname during install

CrystalChun:
Cherry pick of PR https://github.com/openshift/assisted-installer/pull/788 
https://issues.redhat.com/browse/MGMT-16843
Some processes during install time depend on the hostname and an invalid hostname (e.g. 'localhost', having capital letters, etc.) that could cause the install to fail.

Extra validation is added when checking the hostname to ensure it is valid. A random hostname will be assigned if the current hostname is invalid.

/cc @gamli75


CrystalChun:
/remove-label downstream-change-needed

Since the dockerfile change is for pulling the correct linter

PR #798, opened 2024-03-10
MGMT-16993: [STG] avoid reboot not working correctly when there is a partition on installation disk

ori-amizur:
When adding a partition on installation disk, the partition is part of ignition that contains machine config.  The ignition has a version and may contain some fields that are present only in later versions of coreos-ignition.
Since the basic ignition served by machine-config-server has version 2.2, when there is parition in a ignition 3.*, the MCS refuses to serve
  the content with internal server error.  In order to overcome this, a
header "Accept: application/vnd.coreos.ignition+json; version=3.4.0" is added which instructs MCS to convert the returned ignition to 3.4.0

/cc @tsorya 
/cc @javipolo

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -43,6 +43,7 @@ const (
 	dryRunCoreosInstallerExecutable = "dry-installer"
 	encapsulatedMachineConfigFile   = "/etc/ignition-machine-config-encapsulated.json"
 	defaultIgnitionPlatformId       = "ignition.platform.id=metal"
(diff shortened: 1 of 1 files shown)


javipolo:
Why 3.2.0 and not 3.4.0 which is newer?

ori-amizur:
> Why 3.2.0 and not 3.4.0 which is newer?

OK. Will change

javipolo (approved):

PR #808, opened 2024-04-02
MGMT-17305:  Extract disk cleanup from the installer to ops package to allow using it in IBI flow

eranco74:


eranco74:
packet-setup failed
/retest

eranco74:
packet-setup failed
/retest

src/installer/installer.go
```diff
@@ -95,14 +98,17 @@ func (i *installer) InstallNode() error {
 
 	i.UpdateHostInstallProgress(models.HostStageStartingInstallation, i.Config.Role)
 	i.Config.Device = i.ops.EvaluateDiskSymlink(i.Config.Device)
-	err := i.cleanupInstallDevice()
-	if err != nil {
-		i.UpdateHostInstallProgress(models.HostStageStartingInstallation, fmt.Sprintf("Could not clean install device %s. The installation will continue. If the installation fails, clean the disk and try again", i.Device))
-		// Do not change the phrasing of this error message, as we rely on it in a triage signature
-		i.log.Errorf("failed to prepare install device %s, err %s", i.Device, err)
+	if i.DryRunEnabled || i.Config.SkipInstallationDiskCleanup {
```
tsorya, line 101: Maybe better to wrap those if-esle in function ? and to leave it same way it was?

src/ops/ops.go
```diff
@@ -1260,3 +1260,94 @@ func (o *ops) OverwriteOsImage(osImage, device string, extraArgs []string) error
 	}
 	return nil
 }
+
+type DiskOps struct {
```
tsorya, line 1264: why not to create new file for diskops?

eranco74: Because in order to use diskOps you  need ops anyway so if you want to use it outside of assisted-installer you will need to import both packages

eranco74:
packet-setup failed
/retest

PR #823, opened 2024-04-15
OCPBUGS-32105: Fix race to mark node Joined

zaneb:
In the race between assisted-installer on the bootstrap node and assisted-installer-controller on the cluster control plane to mark nodes as Joined, a win for the assisted-installer-controller would cause the bootstrapping process to lock up for 30+ minutes.

Prevent this by not retrying HTTP requests that receive a 409 response at the HTTP transport level. Instead, retry at the logic level and avoid making requests that cannot succeed.

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -776,15 +776,17 @@ func (i *installer) updateReadyMasters(nodes *v1.NodeList, readyMasters *[]strin
 			ctx := utils.GenerateRequestContext()
 			log := utils.RequestIDLogger(ctx, i.log)
 			log.Infof("Found a new ready master node %s with id %s", node.Name, node.Status.NodeInfo.SystemUUID)
-			*readyMasters = append(*readyMasters, node.Name)
 
 			host, ok := common.HostMatchByNameOrIPAddress(node, inventoryHostsMap, knownIpAddresses)
+			if ok && (host.Host.Status == nil || *host.Host.Status != models.HostStatusInstalled) {
+				if err := i.inventoryClient.UpdateHostInstallProgress(ctx, host.Host.InfraEnvID.String(), host.Host.ID.String(), models.HostStageJoined, ""); err != nil {
+					log.Errorf("Failed to update node installation status, %s", err)
+					continue
+				}
+			}
+			*readyMasters = append(*readyMasters, node.Name)
 			if !ok {
 				return fmt.Errorf("node %s is not in inventory hosts", node.Name)
-			}
-			ctx = utils.GenerateRequestContext()
-			if err := i.inventoryClient.UpdateHostInstallProgress(ctx, host.Host.InfraEnvID.String(), host.Host.ID.String(), models.HostStageJoined, ""); err != nil {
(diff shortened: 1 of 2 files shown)


src/installer/installer.go
```diff
@@ -776,16 +776,18 @@ func (i *installer) updateReadyMasters(nodes *v1.NodeList, readyMasters *[]strin
 			ctx := utils.GenerateRequestContext()
 			log := utils.RequestIDLogger(ctx, i.log)
 			log.Infof("Found a new ready master node %s with id %s", node.Name, node.Status.NodeInfo.SystemUUID)
-			*readyMasters = append(*readyMasters, node.Name)
 
 			host, ok := common.HostMatchByNameOrIPAddress(node, inventoryHostsMap, knownIpAddresses)
+			if ok && (host.Host.Status == nil || *host.Host.Status != models.HostStatusInstalled) {
```
tsorya, line 781: I think better to leave it as is, in case we failed to update status but node is ready we better to exit and continue installation, no? Nothing critical in not setting Joined state it will be handle in controller afterwards

src/installer/installer.go
```diff
@@ -776,16 +776,18 @@ func (i *installer) updateReadyMasters(nodes *v1.NodeList, readyMasters *[]strin
 			ctx := utils.GenerateRequestContext()
 			log := utils.RequestIDLogger(ctx, i.log)
 			log.Infof("Found a new ready master node %s with id %s", node.Name, node.Status.NodeInfo.SystemUUID)
-			*readyMasters = append(*readyMasters, node.Name)
```
tsorya, line 779: I think better to leave it as is, in case we failed to update status but node is ready we better to exit and continue installation, no? Nothing critical in not setting Joined state it will be handle in controller afterwards

src/installer/installer.go
```diff
@@ -776,16 +776,18 @@ func (i *installer) updateReadyMasters(nodes *v1.NodeList, readyMasters *[]strin
...
+				if err := i.inventoryClient.UpdateHostInstallProgress(ctx, host.Host.InfraEnvID.String(), host.Host.ID.String(), models.HostStageJoined, ""); err != nil {
+					log.Errorf("Failed to update node installation status, %s", err)
+					continue
+				}
+			}
+			*readyMasters = append(*readyMasters, node.Name)
 			if !ok {
 				return fmt.Errorf("node %s is not in inventory hosts", node.Name)
 			}
-			ctx = utils.GenerateRequestContext()
```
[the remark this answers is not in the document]
zaneb: It's already set on line 776. We don't need to generate another one, we haven't even made a request with the first one yet. And this made it really hard to debug, since the request ID that showed up in the assisted-service log never appeared in the logs here.

src/installer/installer.go
zaneb: Yeah, it's not as big a deal as I originally thought, because the controller will handle it anyway. But this does guarantee that everything is in the state we expect before we carry on to other work in here. It's still robust against unexpected nodes joining, and if the controller does win the race then this will still work on the next attempt.

PR #838, opened 2024-04-24
MGMT-17305: Creating sub module for disk operations in order to import lifecycle-agent-operator

tsorya:
[MGMT-17305](https://issues.redhat.com//browse/MGMT-17305): Creating sub module for disk operations in order to import it in lifecycle-agent-operator.
 This submodule will be used for disk cleanup operations.

When performing IBI, we're not previously wiping the installation disk
As our experience with assisted-installer shows, this can create problems
We should see how to reuse the disk wiping that we already do in assisted: https://github.com/openshift/assisted-installer/blob/f4b8cfd85dfe8194aac489bacfb93ef8501fd290/src/installer/installer.go#L780

Changing structure of assisted-installer and adding new submodel the same way we did it in assisted-service.
It is required cause we have very hard to import dependencies and splitting it to submodule will alow to import only needed one.
Adding logic to Makefile to handle submodel logic


omertuc:
What are these operations? Why does LCA need them from assisted-service?

tsorya:
> I don't think code sharing is a great idea in this case - people making changes to assisted-installer code have no idea there are external projects, that are completely different from assisted, depending on this code. So they might make breaking changes that only make sense for assisted and not those other projects (e.g. LCA).
> 
> There's no versioning of this new module, no guarantees being made. Either we make this disk cleanup thing become its own separate project, where it's clear it has multiple external users (LCA + Assisted), or we should just copy the code to LCA and let the two copies evolve independently to fit the needs of either project, without need for compatibility guarantees across them

We didn't wanted to create new repo and didn't wanted to copy paste code. In this case it was sort of requirement coming from @eranco74. I have nothing against all 3 approaches to say the truth but in this case it was the requirement.
Person who adds changes to assisted-installer submodule should understand that submodule is something that can be used externally, more of it,  the only thing why it exists is cause it is used externally.
Versioning will looks like in assisted-service submodules (though it is not the best one i agree).
Another thing that can be taken into consideration is that IBI and assisted are very close to each other and are part of the same group so there is a logical explanation why we can use assisted code in LCA

omertuc:
>should understand that submodule is something that can be used externally

How would they understand that? They're in their IDE editing assisted code, jumping between functions, they won't even notice they're in a different module. How would they even know what other projects are using this module, and what those projects' expectations are? I don't honestly believe anyone would have LCA in mind editing this code, unless some reviewer is sharp enough to notice it

The CI is also different - you could edit this code and break LCA without knowing, and you'll only find out when LCA bumps the dependency and the CI hopefully catches it (or not and a bug is reported much later). Then you have to bisect which commit is the culprit.

For these reasons I'm against having code being "lightly" shared between projects. Either it's clearly shared, in which case it's separate repo, has a clear interface, clear expectations of behavior, and tests/CI to ensure those expectations are met, or it should be copied

tsorya:
We should talk about, i really understand the arguments though this part is totally the same between both. When we will find any bug in one of them we will need to fix the other. This makes logical to use same code for both cases without copy pasting it. Creating new repo seems like overkill too. Importing this code as is is very problematic cause of pretty hard dependencies. That's why i did submodule here.

eranco74:
OK, so we would like to share the the code (note that the disk cleanup keeps evolving to handles new special cases) and not to duplicate it.
Overall both AI and IBI use this code to perform the same task (running from live-iso and writing rhcos to the installation disk).
I agree that one might change the disk cleanup code in a way that can break IBI (once LCA updates the assisted-installer dependency) but that's true to assisted-installer as well as we don't have real coverage for all the edge cases the disk cleanup tries to handle (RAID, disks preconfigured with LVS or DM devices).
In any case, we all agree that another repo is the preferred solution that should allow us to easily reuse the code with a separate dedicated library.
The downside is that it will add complexity every change will require changes to the lib and then updating the dependency.
Also, the maintenance overhead of a separate project (within Openshift) isn't cheap.
What about adding a label on PRs that changes files in the disk cleanup submodules? we use a similar approach when changing the API, I think it should help here as well.

omertuc:
https://github.com/rh-ecosystem-edge/preinstall-utils

paul-maidment:
I actually support the point from @omertuc about this. I think that a completely isolated library is the way to go.

This ensures that there is no risk of divergence between consumers of the library (though it is accepted that a consumer may choose to remain on an older version for whatever reason...this is reasonably clear to see in the `go.mod` / `go.sum` though...) 

I would argue that as the intent is to enable code reuse that it's worth the inconvenience of splitting this into another library.
This approach actively enforces the maintenance of this code in a single location which I think is a massive upside.

PR #839, opened 2024-04-24
MGMT-16843: Use hostnamectl to replace illegal hostname

CrystalChun:
Hosts may not restart before beginning installation
so if we generate a random hostname, it won't take
effect unless we use hostnamectl to actually replace
the hostname live.

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -383,10 +383,10 @@ func (i *installer) startBootstrap() error {
 		return err
 	}
 
-	/* in case hostname is localhost, we need to set hostname to some random value, in order to
-	   disable network manager activating hostname service, which will in turn may reset /etc/resolv.conf and
-	   remove the work done by 30-local-dns-prepender. This will cause DNS issue in bootkube and it will fail to complete
-	   successfully
+	/* in case hostname is illegal (localhost, contains capital letters, etc.), we need to set hostname to
+	   some random value, in order to disable network manager activating hostname service, which will in
+	   turn may reset /etc/resolv.conf and remove the work done by 30-local-dns-prepender. This will cause
+	   DNS issue in bootkube and it will fail to complete successfully
 	*/
 	err = i.checkHostname()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


src/installer/installer.go
```diff
@@ -883,8 +883,17 @@ func (i *installer) checkHostname() error {
...
-	return i.ops.CreateRandomHostname(data)
+	i.log.Infof("Hostname [%s] is invalid, generated random hostname [%s] and writing data into /etc/hostname", data)
+	if err := i.ops.CreateRandomHostname(data); err != nil {
+		i.log.Errorf("Failed to generate random hostname", err)
+		return err
+	}
+
+	if highAvailabilityMode == models.ClusterHighAvailabilityModeNone {
+		i.log.Infof("Restarting systemd-hostnamed service for SNO after writing new hostname [%s]", data)
+		i.ops.SystemctlAction("restart", "systemd-hostnamed")
```
tsorya, line 894: Wondering, maybe it is just easier to run hostnamectl command for all cases? it should set everything as expected without need to restart service?

CrystalChun: Good idea to use hostnamectl, will update. Do you also mean to do it on multi-node clusters too?

CrystalChun: I tested both ways and I get this when trying to access the installed spoke cluster 
```
Unable to connect to the server: tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "kube-apiserver-lb-signer")
```

CrystalChun: Should I send the requested hostname to the installer?

tsorya: very strange, hostname is not part of kubeconfig

tsorya: very strange, hostname is not part of kubeconfig

CrystalChun: Ah sorry this error was a complete fluke. Something weird with my env 😅 Made the change to use `hostnamectl` to set the hostname for any cluster (not just SNO)!

CrystalChun:
/retitle MGMT-16843: Use hostnamectl to replace illegal hostname

CrystalChun:
/test edge-e2e-ai-operator-ztp

Forgot `ci/prow/edge-e2e-metal-assisted` is actually broken https://issues.redhat.com/browse/OCPBUGS-33493

CrystalChun:
/cherry-pick release-ocm-2.10 release-ocm-2.9 release-4.16 release-4.15

PR #840, opened 2024-05-01
MGMT-17305: Moving disk operations to new repo

tsorya:
[MGMT-17305](https://issues.redhat.com//browse/MGMT-17305): Moving disk operations to new repo
Start using new repo for disk cleanup

Created new repo for disk cleanup operations, moving to use it.
As github.com/golang/mock is not maintained anymore change it to go.uber.org/mock


src/installer/installer_test.go
```diff
@@ -95,11 +98,15 @@ var _ = Describe("installer HostRoleMaster role", func() {
...
-				mockops.EXPECT().GetVolumeGroupsByDisk(device).Return([]string{}, nil).Times(1)
-				mockops.EXPECT().RemoveAllPVsOnDevice(device).Return(nil).Times(1)
-				mockops.EXPECT().RemoveAllDMDevicesOnDisk(device).Return(nil).Times(1)
-				mockops.EXPECT().IsRaidMember(device).Return(false).Times(1)
-				mockops.EXPECT().Wipefs(device).Return(nil).Times(1)
+				cleanupDevice.EXPECT().CleanupInstallDevice(device).Return(nil).Times(1)
+			}
+
+			cleanInstallDeviceWithDiskOps := func() {
+				diskOps.EXPECT().GetVolumeGroupsByDisk(device).Return([]string{}, nil).Times(1)
```
paul-maidment, line 105: I'm not completely against testing this here but is there not an argument to be made that testing of the "internal" functionality of the preinstall util library should now be performed in the rh-ecosystem-edge/preinstall-utils library?

As I see it, the responsibility that now exists in assisted-installer is greatly reduced. 

The only responsibilities being around the following call and the handling of any errors from the library?

`cleanupDevice.EXPECT().CleanupInstallDevice(device).Return(nil).Times(1)`

In theory, the manner in which the preinstall library works could change over time, thus rendering this test more fragile?

eranco74 (commented):
/lgtm

PR #858, opened 2024-06-20
NO-ISSUE: We should set disk symlink only for coreos installer command

tsorya:
As part of installation flow we set boot order in efibootmanager.
Till now we used disk name as device for the command and it creates many issues after booting as in many machines disk name changes and we start with wrong disk. This happened due to old bug in coreos installer that required disk name rather than disk id.
This change leaves workaround for a bug due to support of old ocp versions but uses disk name only for coreos-isntall command and all other commands will get disk id, this should allow us to set right disk in eftbootmanager and node should boot form the right disk


eranco74:
@tsorya why do we need this change?

tsorya:
I will open a ticket, opened PR for test (still waiting for qe). Will update the description

PR #862, opened 2024-07-02
NO-ISSUE: Replace base golang image in the build Dockerfile

danmanor:
Replace base golang image in the build Dockerfile as it is based on centos7 and it is EOL, which results in yum can't reach its repositories

PR #865, opened 2024-07-02
[release-ocm-2.10] NO-ISSUE: Replace base golang image in the build Dockerfile as it is based on centos7 and it is EOL, which results in yum can't reach its repositories (#862)

adriengentil:


danmanor:
not affected

PR #866, opened 2024-07-02
[release-ocm-2.9] NO-ISSUE: Replace base golang image in the build Dockerfile as it is based on centos7 and it is EOL, which results in yum can't reach its repositories (#862)

adriengentil:


adriengentil:
@danmanor this one need to be force merged because the dockerfile is used as build root

adriengentil:
/override ci/prow/e2e-ai-operator-ztp ci/prow/format-check ci/prow/images ci/prow/lint ci/prow/mce-images ci/prow/unit-test

PR #872, opened 2024-07-07
[release-ocm-2.8] MGMT-18332: Use centos stream 9 as the base image for test container

danmanor:


Dockerfile.assisted-installer-build
```diff
@@ -1,19 +1,34 @@
...
+
+RUN dnf install -y 'dnf-command(config-manager)' && \
+    dnf config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo && \
+    dnf install -y \
+        docker-ce \
+        docker-ce-cli \
+        containerd.io \
+        make \
+        git \
+        openssl-devel \
```
pastequo, line 23: Did you check if openssl-devel was needed as well in those repo ?

Dockerfile.assisted-installer-build
```diff
@@ -1,19 +1,34 @@
...
+        gcc \
```
pastequo, line 24: I would also install util-linux explicitly. I think it comes with one of the docker stuff, but without it I had trouble to runner skipper locally

danmanor: why would util-linux inside the container would relate to skipper on the host ? I don't need for skipper to work

pastequo: https://github.com/Stratoscale/skipper/blob/upstream/skipper/data/skipper-entrypoint.sh#L34

danmanor: Oh this is because you are using later version of skipper, I use 2.0.2

danmanor: OK lets add it for any case

danmanor: I think this problem would occur for you with a lot of other Dockerfiles we use. Maybe add it in different PRs ?

danmanor: Although I would recommend using skipper==2.0.2

PR #882, opened 2024-07-18
NO-ISSUE: fix for ARM and multi images

eifrach:


eifrach:
/retitle NO-ISSUE: fix for ARM and multi images

PR #889, opened 2024-07-28
[release-ocm-2.9] MGMT-18313: Replace golang base image as it is based on Centos Linux 7

danmanor:
This PR moves centos base image from stream8 to stream 9 as well as stream8 is EOL

### DIFF
--- a/Dockerfile.assisted-installer
+++ b/Dockerfile.assisted-installer
@@ -1,4 +1,8 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18 AS builder
+
+# Workaround for creating build folder
+USER root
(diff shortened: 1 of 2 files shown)


Dockerfile.assisted-installer
```diff
@@ -1,4 +1,8 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18 AS builder
```
eifrach, line 1: why are you downgrading the Go version here?
is it by mistake ?

eifrach (changes requested):

Dockerfile.assisted-installer
danmanor: This is the actual go version, look at `go.mod`

PR #895, opened 2024-08-29
MGMT-18684: Ensure that bootstrap CSR is authenticated against cluster hosts.

paul-maidment:
Once the minimal control plane has been performed and consists of two nodes, we start the assisted-installer-controller pod and then reboot the bootstrap when this is ready. The bootstrap will attempt to join the control plane when it boots up. As part of this process, CSR's are submitted. Presently, we just assume that these are meant to be signed and blindly approve them.

This is problematic from a security perspective and could lead to an attack where, with careful timing, someone might be able to add a rogue node to the cluster. To prevent this, we should check that the hostname being presented in the CSR actually matches one of our known hosts.

To do this, requires two steps:

1: Prior to start of the assisted-installer-controller pod, we will save a list of known hosts in a config map called `known-hosts`, this will contain a single entry `content` which will be a newline separated list of hostnames
2: During the CSR validation, this will be read and used to verify that the CSR originates from a known host

### DIFF
--- a/src/common/common.go
+++ b/src/common/common.go
@@ -3,6 +3,10 @@ package common
 import (
 	"bytes"
 	"context"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"io"
 	"os"
@@ -15,6 +19,8 @@ import (
 	"github.com/openshift/assisted-installer/src/k8s_client"
 	"github.com/openshift/assisted-installer/src/utils"
 	"github.com/openshift/assisted-service/models"
+
+	cryptorand "crypto/rand"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -35,6 +41,7 @@ const (
 	installConfigMapAttribute      = "invoker"
 	InvokerAssisted                = "assisted-service"
 	InvokerAgent                   = "agent-installer"
+	ECPrivateKeyPEMLabel           = "EC PRIVATE KEY"
 )
 
 func GetHostsInStatus(hosts map[string]inventory_client.HostData, status []string, isMatch bool) map[string]inventory_client.HostData {
@@ -305,4 +312,23 @@ func DownloadKubeconfigNoingress(ctx context.Context, dir string, ic inventory_c
 	log.Infof("Downloaded %s to %s.", KubeconfigFileName, kubeconfigPath)
 
 	return kubeconfigPath, nil
+}
+
+func MakeEllipticPrivatePublicKeyPems() ([]byte, []byte, error) {
+	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), cryptorand.Reader)
+	if err != nil {
+		return nil, nil, err
+	}
+	derBytes, err := x509.MarshalECPrivateKey(privateKey)
+	if err != nil {
+		return nil, nil, err
+	}
+	publicKeyBytes, err := x509.MarshalPKIXPublicKey(&privateKey.PublicKey)
+	if err != nil {
+		return nil, nil, err
+	}
+	return pem.EncodeToMemory(&pem.Block{
+		Type:  ECPrivateKeyPEMLabel,
+		Bytes: derBytes,
+	}), publicKeyBytes, nil
 }
--- a/src/ignition/ignition.go
+++ b/src/ignition/ignition.go
@@ -2,13 +2,17 @@ package ignition
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
+
+	"encoding/base64"
 
 	ignitionConfigPrevVersion "github.com/coreos/ignition/v2/config/v3_1"
 	ignitionConfig "github.com/coreos/ignition/v2/config/v3_2"
 	"github.com/coreos/ignition/v2/config/v3_2/translate"
 	"github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/coreos/ignition/v2/config/validate"
+	"github.com/openshift/assisted-installer/src/convert"
 	"github.com/pkg/errors"
 )
 
@@ -21,6 +25,7 @@ type Ignition interface {
 	ParseIgnitionFile(path string) (*types.Config, error)
 	WriteIgnitionFile(path string, config *types.Config) error
 	MergeIgnitionConfig(base *types.Config, overrides *types.Config) (*types.Config, error)
+	InjectKubeletTempPrivateKey(pathToSourceIgnition string, privateKeyBytes []byte, certPathToInject string) error
 }
 
 type ignition struct{}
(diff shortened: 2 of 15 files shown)


paul-maidment:
> /hold Since it's trivial to create CSRs with any hostname, this doesn't improve security

I accept the point about this doing little to nothing to advance security as I can see that you can't be sure that the CSR was not created with an identical host name and different credentials.

However, thinking carefully about the original request in [MGMT-18684](https://issues.redhat.com/browse/MGMT-18684)
I am not actually sure we even have a security issue in the first place...

"This means that there is a window during installation when a rogue node can join the cluster with no verification. "

is partially true but we need to consider _how_ a CSR can be registered with the cluster; by an authenticated user of the control plane who has been granted permission to create CSRs ( https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#authorization )

This means that they 

1) Need a Kubeconfig
2) Need to have authority to create CSRs

Doesn't this mean that the joining node needs to satisfy these requirements when registering the CSR?
If so then we only really need to check in a similar fashion to how the cluster-machine-operator does already...

https://github.com/openshift/cluster-machine-approver/blob/master/pkg/controller/csr_check.go#L364

Granted we would probably be best to forgo the "pre reboot" information gathering phase and instead rely on DNS records in the cluster post reboot.

WDYT?

omertuc:
> This means that they
> 
> 1. Need a Kubeconfig
> 2. Need to have authority to create CSRs
> 
> Doesn't this mean that the joining node needs to satisfy these requirements when registering the CSR?

1. "Need a kubconfig" - all you need is the network connectivity and the IP address of the API server. `kubeconfig` is "needed" only to make `oc` work, but users can just as well `curl` the API server or use whatever they want, it's just HTTP requests

2. Pretty sure in OCP anyone is authorized to create CSRs, this is how completely novel nodes without any credentials or information about the cluster are able to create CSRs that allow them to join. CSR *approvals* are the only line of defense we have against rogue nodes joining as far as my understanding goes

paul-maidment:
> > This means that they
> > 
> > 1. Need a Kubeconfig
> > 2. Need to have authority to create CSRs
> > 
> > Doesn't this mean that the joining node needs to satisfy these requirements when registering the CSR?
> 
>     1. "Need a kubconfig" - all you need is the network connectivity and the IP address of the API server. `kubeconfig` is "needed" only to make `oc` work, but users can just as well `curl` the API server or use whatever automation they want, it's just HTTP requests. Also MCS will give you this kubeconfig for free, unauthenticated when you query it for ignition inside the ignition
> 
>     2. Pretty sure in OCP anyone is authorized to create CSRs, this is how completely novel nodes without any credentials or information about the cluster are able to create CSRs that allow them to join. CSR _approvals_ are the only line of defense we have against rogue nodes fully joining the cluster as far as my understanding goes

What are your thoughts on this implementation here?

https://github.com/openshift/cluster-machine-approver/blob/master/pkg/controller/csr_check.go#L364

The checks that we perform to ensure that the node with a specific name

1) Does not already exist
2) Is resolvable to an IP within the cluster DNS
3) Resolves to a known machine.

Are these sufficient to prevent a rogue node from joining or do they still fall short? Is `cluster-machine-approver` secure enough to be trusted or is it prone  to the same issues you highlight?

paul-maidment:
"Pretty sure in OCP anyone is authorized to create CSRs, this is how completely novel nodes without any credentials or information about the cluster are able to create CSRs that allow them to join. CSR approvals are the only line of defense we have against rogue nodes fully joining the cluster as far as my understanding goes"

I have tried various ways of using curl requests to create and list CSR's as an anonymous user, pretty sure that I am using the correct syntax (which oc will quite happily dump for you if you add a parameter like `-v 1000000`)

```
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"certificatesigningrequests.certificates.k8s.io is forbidden: User \"system:anonymous\" cannot list resource \"certificatesigningrequests\" in API group \"certificates.k8s.io\" at the cluster scope","reason":"Forbidden","details":{"group":"certificates.k8s.io","kind":"certificatesigningrequests"},"code":403}
```

I receive similar messages for creation attempts also. It seems that there is not an "anonymous" way to create a CSR over HTTPS, I could be mistaken but I think we should check this in detail. Do you know who might be able to verify this? One of the Node team perhaps?

These requests are being made against the "minimal" control plane that is created just before the bootstrap joins.

omertuc:
>Are these sufficient to prevent a rogue node from joining or do they still fall short? 

>if an attack on cluster-machine-approver is possible, does this mean we have a broader issue?

>What are your thoughts on this implementation here?

Not sure, seems rather insufficient. But it depends on the requirements for cluster-machine-approver and their threat model, I can't comment on the security of that component without having this context.

I can however comment on assisted-installer, which is supposed to be a user-friendly installer, not meant for power users, so the security requirements should be tight and fool-proof or come with a ton of UI warnings to help users protect themselves from our non-ideal behavior.

>Is resolvable to an IP within the cluster DNS on the cluster side

> Is it still viable that an attacker with the right kind of static IP setup and timing could add a rogue node? I think maybe, but certainly a lot harder to pull off...

What does "Resolveable to an IP" mean here? If you're referring to the IP from which the request to create the CSR came from, then that definitely makes life harder for an attacker. But I'm not sure that information is even available outside of audit logs. If you're referring to the IP addresses specified in the CSR itself, anyone can set those to anything arbitrarily as far as I'm aware, just like they can set the hostname.

>This sounds like a sizeable security hole, much worse than the timing issue as if I have the KubeConfig I have full control, I don't need to play games to add nodes, I can just do it...

It gives you _a kubeconfig_, not an _admin kubeconfig_. It's basically a YAML with a URL in it

>I receive similar messages for creation attempts also. It seems that there is not an "anonymous" way to create a CSR over HTTPS, I could be mistaken but I think we should check this in detail. 

There must be some way to do it, as this is what the kubelet on fresh nodes with 0 credentials does (all they have is the ignition they got from MCS), so you're probably missing something.

src/common/common.go
```diff
@@ -306,3 +314,55 @@ func DownloadKubeconfigNoingress(ctx context.Context, dir string, ic inventory_c
...
+		return nil, nil, err
+	}
+	derBytes, err := x509.MarshalECPrivateKey(privateKey)
+	if err != nil {
+		return nil, nil, err
+	}
+	ecdhPubKey, err := privateKey.PublicKey.ECDH()
+	if err != nil {
+		return nil, nil, err
+	}
```
omertuc, line 363: ECDH is unrelated, we should do something along the lines of:

```suggestion
	derBytes, err = x509.MarshalPKIXPublicKey(&privateKey.PublicKey)
	if err != nil {
		return nil, nil, err
	}
```

src/ignition/ignition.go
```diff
@@ -78,3 +83,32 @@ func (i *ignition) MergeIgnitionConfig(base *types.Config, overrides *types.Conf
...
+		FileEmbedded1: types.FileEmbedded1{
+			Contents: types.Resource{
+				Source: swag.String(
+					fmt.Sprintf(
+						"data:text/plain;charset=utf-8;base64,%s",
+						base64.StdEncoding.EncodeToString([]byte(privateKeyBytes)),
+					),
+				),
+			},
+			Mode: swag.Int(420),
```
omertuc, line 109: I don't like bringing in swag as a dependency into this code, just to abuse its shortcut for making int pointers. Also use Octal literals:
```suggestion
			Mode: swag.Int(0644),
```

Also `644` is inappropriate for a private key, it should be `600`, we want the to restrict access to it as much as possible

image

image

src/installer/installer.go
```diff
@@ -96,6 +98,19 @@ func (i *installer) FormatDisks() {
 	}
 }
 
+func (i *installer) StorePublicKeyInControlPlane(publicKey []byte, client k8s_client.K8SClient) error {
+	err := client.CreateNamespace("node-public-keys")
```
omertuc, line 102: Prefix with `assisted-` so its obvious this is an assisted mechanism and not an OCP thing

src/installer/installer.go
```diff
@@ -522,6 +562,21 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
...
+	// 	i.log.Warnf("Failed to get hosts info from inventory, err %s", err)
+	// }
+	// err = kc.CreateNamespace("bootstrap-host-info")
+	// if err != nil {
+	// 	i.log.Error("unable to create bootstrap-host-info namespace on control plane", err)
+	// }
+	// err = kc.CreateConfigMap("known-hosts", "bootstrap-host-info", map[string]string{"content": strings.Join(hostNames, "\n")})
+	// if err != nil {
+	// 	i.log.Error("unable to create known-hosts cm on control plane", err)
+	// }
```
omertuc, line 578:
```suggestion
```

src/installer/installer.go
```diff
@@ -531,6 +586,19 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
...
+// 	var inventoryHostsMap map[string]inventory_client.HostData
+// 	inventoryHostsMap, err := i.getInventoryHostsMap(inventoryHostsMap)
+// 	if err != nil {
+// 		return nil, errors.Wrap(err, "failed to get hosts info from inventory")
+// 	}
+// 	for k := range inventoryHostsMap {
+// 		keys = append(keys, k)
+// 	}
+// 	return keys, nil
+// }
```
omertuc, line 600:
```suggestion
```

omertuc (changes requested):
Very cool

PR #896, opened 2024-09-03
OCPBUGS-38466: Allow controller to continue when assisted-service

rwsu:
is unavailable in agent-based installs.

assisted-service runs on the bootstrap node in agent-based installs. The bootstrap node reboots after the control plane is available.

If the assisted-installer-controller restarts after the bootstrap node reboots, or for some reason the controller is never able to contact assisted-service, the controller loops waiting or assisted-service to become available, times out, and exits.

With compact clusters, because the controller exited and is unable to approve CSRs, the third control plane node is unable to join the cluster causing the cluster installation to fail.

If the invoker is agent-installer, instead of exiting, this patch allows the controller to continue to run when assisted-service is offline.

Because assisted-service may be unavailable,
HasValidvSphereCredentials has been updated to also look at the install-config to determine if credentials were set. Because username and password are redacted, only the server name is used to determine if valid credentials were provided.

### DIFF
--- a/src/common/common.go
+++ b/src/common/common.go
@@ -21,6 +21,7 @@ import (
 	"github.com/thoas/go-funk"
 	"github.com/tidwall/gjson"
 	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/yaml"
 )
 
 const (
@@ -33,6 +34,9 @@ const (
 	installConfigMapName           = "openshift-install-manifests"
 	installConfigMapNS             = "openshift-config"
 	installConfigMapAttribute      = "invoker"
+	clusterConfigCMName            = "cluster-config-v1"
+	clusterConfigCMNamespace       = "kube-system"
+	clusterConfigCMAttribute       = "install-config"
 	InvokerAssisted                = "assisted-service"
 	InvokerAgent                   = "agent-installer"
 )
@@ -215,24 +219,23 @@ func HostMatchByNameOrIPAddress(node v1.Node, namesMap, IPAddressMap map[string]
 // Returns True when uninitialized taint must be removed.
 // This is required for some external platforms (e.g. VSphere, Nutanix) to proceed
 // with the installation using fake credentials.
-func RemoveUninitializedTaint(platform *models.Platform, invoker string, hasValidCredentials bool, openshiftVersion string) bool {
+func RemoveUninitializedTaint(ctx context.Context, ic inventory_client.InventoryClient, kc k8s_client.K8SClient,
+	log logrus.FieldLogger, platformType models.PlatformType, openshiftVersion, invoker string) bool {
 	removeUninitializedTaintForPlatforms := [...]models.PlatformType{models.PlatformTypeNutanix, models.PlatformTypeVsphere}
-	version := semver.New(parseOpenshiftVersionIntoMajorMinorZOnly(openshiftVersion))
-	if invoker == InvokerAgent &&
-		*platform.Type == models.PlatformTypeVsphere &&
-		version.Compare(*semver.New("4.15.0")) >= 0 &&
-		hasValidCredentials {
-		// Starting with OpenShift 4.15, the agent-installer can pass valid credentials
-		// to vSphere. Do not remove the taint if:
-		// 1) invoker == agent-installer
-		// 2) platform == vSphere
-		// 3) OpenShift version >= 4.15
-		// 4) and valid vSphere credentials were provided in install-config.yaml
-		// With valid credentials, the vSphere CCM will remove the uninitialized taints
-		// when it initializes the nodes.
-		return false
+	if invoker == InvokerAgent && platformType == models.PlatformTypeVsphere {
+		hasValidvSphereCredentials := HasValidvSphereCredentials(ctx, ic, kc, log)
+		if hasValidvSphereCredentials {
+			log.Infof("Has valid vSphere credentials: %v", hasValidvSphereCredentials)
+		}
+		version := semver.New(parseOpenshiftVersionIntoMajorMinorZOnly(openshiftVersion))
+		if hasValidvSphereCredentials && version.Compare(*semver.New("4.15.0")) >= 0 {
+			// Starting with OpenShift 4.15, vSphere credentials can be used with ABI.
+			// In such cases, we should not remove the taints because the vSphere CCM will
+			// remove them after it initializes the nodes.
+			return false
+		}
 	}
-	return platform != nil && funk.Contains(removeUninitializedTaintForPlatforms, *platform.Type)
+	return funk.Contains(removeUninitializedTaintForPlatforms, platformType)
 }
 
 // parseOpenshiftVersionIntoMajorMinorZOnly adds a .0 to verions that only specify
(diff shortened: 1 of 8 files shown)


src/common/common.go
```diff
@@ -215,11 +216,10 @@ func HostMatchByNameOrIPAddress(node v1.Node, namesMap, IPAddressMap map[string]
 // Returns True when uninitialized taint must be removed.
 // This is required for some external platforms (e.g. VSphere, Nutanix) to proceed
 // with the installation using fake credentials.
-func RemoveUninitializedTaint(platform *models.Platform, invoker string, hasValidCredentials bool, openshiftVersion string) bool {
-	removeUninitializedTaintForPlatforms := [...]models.PlatformType{models.PlatformTypeNutanix, models.PlatformTypeVsphere}
+func RemoveUninitializedTaint(platformType string, invoker string, hasValidCredentials bool, openshiftVersion string) bool {
```
tsorya, line 219: why to change platformType type?

src/common/common.go
```diff
@@ -248,35 +248,85 @@ func parseOpenshiftVersionIntoMajorMinorZOnly(version string) string {
...
 
 // HasValidvSphereCredentials returns true if the the platform is
-// vSphere and the install config overrides contains real
+// vSphere and the install config or equivalent overrides contains real
 // credential values and not placeholder values.
 // Deprecated credential fields are not considered valid.
-func HasValidvSphereCredentials(ctx context.Context, ic inventory_client.InventoryClient, log logrus.FieldLogger) bool {
+func HasValidvSphereCredentials(ctx context.Context, ic inventory_client.InventoryClient, kc k8s_client.K8SClient, log logrus.FieldLogger) bool {
 	cluster, callErr := ic.GetCluster(ctx, false)
+	if cluster != nil {
```
tsorya, line 256: Should not we check for invoker here? I believe service can be not reachable for period of time and i believe if invoker is agent there is not need for assisted service at all, right?

tsorya, line 256: I see that invoker check is part of another function so i believe it should be fine, right?

paul-maidment:
I know that this use case is to cover the scenario where the user is using agent based install and there is a restart of assisted-service-controller after the reboot of the bootstrap node.

Surely this scenario can affect other installation modes also, are there any mitigations or is this out of scope here?

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -138,20 +138,42 @@ func main() {
...
+			// dereference" and having the controller restart because cluster is nil,
+			// log warning and exit 0. Otherwise the controller will keep restarting
+			// leaving the controller in a running state, even through the cluster
+			// has finished install.
+			logger.Warnf("SetReadyState timed out fetching cluster from assisted-service, invoker = %v", invoker)
+			return
+		}
+
+		if cluster == nil {
+			platformType, err = common.GetPlatformTypeFromInstallConfig(kc, logger)
```
tsorya, line 165: why we need platformType ?

tsorya: Got it, sorry

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -165,20 +187,28 @@ func main() {
...
 
-	hasValidvSphereCredentials := common.HasValidvSphereCredentials(mainContext, client, logger)
-	if hasValidvSphereCredentials {
-		logger.Infof("Has valid vSphere credentials: %v", hasValidvSphereCredentials)
+	if cluster != nil && cluster.Platform != nil {
+		platformType = string(*cluster.Platform.Type)
+	}
+
+	hasValidvSphereCredentials := false
+	if platformType == string(models.PlatformTypeVsphere) {
```
tsorya, line 195: We have this validation https://github.com/openshift/assisted-installer/pull/896/files#diff-a5bd908493da50da6a2905424525321bb4b7b81b90da798952ae8aac3ec53546R258 ? 
Need both places?

src/common/common.go
```diff
@@ -248,35 +248,85 @@ func parseOpenshiftVersionIntoMajorMinorZOnly(version string) string {
...
+}
+
+func getInstallConfigYAML(kc k8s_client.K8SClient, log logrus.FieldLogger) (string, error) {
+	clusterConfigCMName := "cluster-config-v1"
+	clusterConfigCM, err := kc.GetConfigMap("kube-system", clusterConfigCMName)
+	if err != nil {
+		log.Warnf("error retrieving %v ConfigMap", clusterConfigCMName)
+		return "", err
+	}
+	return clusterConfigCM.Data["install-config"], nil
```
tsorya, line 329: const?

src/common/common.go
```diff
@@ -248,35 +248,85 @@ func parseOpenshiftVersionIntoMajorMinorZOnly(version string) string {
...
+	for key := range platformMap {
+		platformType = key
+		break
 	}
-	return false
+	return platformType, nil
+}
+
+func getInstallConfigYAML(kc k8s_client.K8SClient, log logrus.FieldLogger) (string, error) {
+	clusterConfigCMName := "cluster-config-v1"
```
tsorya, line 323: consts?

src/common/common.go
rwsu: The check to see whether there are valid vsphere credentials is only relevant when the invoker is agent. Perhaps I should have RemoveUnitializedTaint call HasValidvSphereCredentials when invoker == agent. Is this what you were suggesting?

In regards to your second question, when assisted-service cannot be reached, this function should then look at the install-config (in else clause) to see if valid credentials were specified there. If both assisted-service and kube-api cannot be reached to check credentials then the function should return false.

src/main/assisted-installer-controller/assisted_installer_main.go
rwsu: It is not need in both places. Thanks for spotting it.

rwsu:
> I know that this use case is to cover the scenario where the user is using agent based install and there is a restart of assisted-service-controller after the reboot of the bootstrap node.
> 
> Surely this scenario can affect other installation modes also, are there any mitigations or is this out of scope here?

@paul-maidment ,I believe it is out of scope for non-ABI installs. The patch is addressing a problem that is unique to ABI installs because assisted-service is hosted on the bootstrap node and it disappears after bootstrap reboots. This problem appears to occur rarely and thus far has only been reported by a single customer. If assisted-installer-controller restarts, because the bootstrap has rebooted, it can never connect to the assisted-service api and the controller fails hard without approving CSRs.

For saas, if assisted-installer-controller restarts, the controller should be able to reconnect to assisted-service. I presume assisted-service is accessible within the connection retry window if it some how becomes unreachable for a short span of time. So it is unaffected when the bootstrap node reboots.

src/common/common.go
rwsu: In the case when assisted-service is unable, the platform type is determined by inspecting the install-config. That comes in as a string. So that is why RemoveUnitializedTaint looks at the string value.

src/common/common.go
rwsu: In additional to having valid credentials, RemoveUninitializedTaint has additional checks on cluster version, invoker, and platformType to determine if the taints should be removed. That seemed to be a specific ABI case, but the function felt that the appropriate place for it. If we don't run RemoveUninitializedTaint, then the additional checks would need to be elsewhere.

src/common/common.go
tsorya: Not easier was to cast string to models.Platform at one place rather than changing previous code?

src/common/common.go
tsorya: Yep, i don't see why not agent code should actually run HasValidCredentials, it makes it not clear in non-agent flow

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -138,20 +138,42 @@ func main() {
...
-		return
+	var cluster *models.Cluster
+	var platformType string
+	if invoker == common.InvokerAgent {
+		// With the agent-based installer, assisted-service runs on the bootstrap node.
+		// We do not need to wait for an extended amount of time to reach assisted-service
+		// because it should be running on the local network.
+		cluster = assistedController.SetReadyState(5 * time.Minute)
+
+		if cluster == nil && Options.ControllerConfig.HighAvailabilityMode == models.ClusterHighAvailabilityModeNone {
```
tsorya, line 149: Will not you always be enter this part in sno mode? Why we actually should even try to get cluster in this case?

src/common/common.go
rwsu: Ok, I'll separate it out of non-agent flow.

src/main/assisted-installer-controller/assisted_installer_main.go
rwsu: You're right, the SNO case doesn't need cluster.

src/common/common.go
rwsu: I'll move it back to type. Not sure how to cast to models.Platform. Could you show me how?

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -304,3 +284,63 @@ func didInstallationFinish(clusterStatus string, log logrus.FieldLogger, status
...
+		if platformType == models.PlatformTypeVsphere {
+			hasValidvSphereCredentials := common.HasValidvSphereCredentials(ctx, ic, kc, logger)
+			if hasValidvSphereCredentials {
+				logger.Infof("Has valid vSphere credentials: %v", hasValidvSphereCredentials)
+			}
+		}
+		removeUninitializedTaint = common.RemoveUninitializedTaintAgent(platformType,
+			hasValidvSphereCredentials, Options.ControllerConfig.OpenshiftVersion)
+	} else {
+		cluster = assistedController.SetReadyState(assistedinstallercontroller.WaitTimeout)
```
tsorya, line 338: Should we just log in case cluster not found? I mean in not agent flow it is error that we should not continue no?

src/common/common.go
```diff
@@ -215,24 +219,24 @@ func HostMatchByNameOrIPAddress(node v1.Node, namesMap, IPAddressMap map[string]
 // Returns True when uninitialized taint must be removed.
 // This is required for some external platforms (e.g. VSphere, Nutanix) to proceed
 // with the installation using fake credentials.
-func RemoveUninitializedTaint(platform *models.Platform, invoker string, hasValidCredentials bool, openshiftVersion string) bool {
+func RemoveUninitializedTaint(platformType models.PlatformType) bool {
```
tsorya, line 222: Sorry, i think i was pretty not clear :( 
i actually meant to run hasValidCredentials function under https://github.com/openshift/assisted-installer/pull/896/files#diff-a5bd908493da50da6a2905424525321bb4b7b81b90da798952ae8aac3ec53546L221
and not to create 2 functions, WDYT?

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -304,3 +284,63 @@ func didInstallationFinish(clusterStatus string, log logrus.FieldLogger, status
...
+			// node reboots to become the SNO cluster. SetReadyState will never be
+			// able to get a connection to assisted-service in this circumstance.
+			//
+			// Instead of exiting with panic with "invalid memory address or nil pointer
+			// dereference" and having the controller restart because cluster is nil,
+			// log warning and exit 0. Otherwise the controller will keep restarting
+			// leaving the controller in a running state, even through the cluster
+			// has finished install.
+			logger.Warnf("cluster is SNO and invoker = %v, skipping assisted-installer-controller", invoker)
+			os.Exit(0)
```
tsorya, line 305: exit? not sure this is the right place

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -304,3 +284,63 @@ func didInstallationFinish(clusterStatus string, log logrus.FieldLogger, status
...
+		}
+		// With the agent-based installer, assisted-service runs on the bootstrap node.
+		// We do not need to wait for an extended amount of time to reach assisted-service
+		// because it should be running on the local network.
+		cluster = assistedController.SetReadyState(5 * time.Minute)
```
tsorya, line 310: I don't think this is the right function to use to say the truth, seems not relevant to way for ready state here, you need cluster from the service and not to set ready even on k8s

src/common/common.go
rwsu: Sorry for my misunderstanding. Let me convert it back to a single function and have hasValidvSphereCredentials be called within RemoveUniniitializedTaint.

src/main/assisted-installer-controller/assisted_installer_main.go
rwsu: What would you suggest to be a better function to use here?

In the majority of the time, ABI should be able to retrieve the cluster from assisted-service. For ABI, if assisted-service isn't unavailable, we just want to fail more quickly (in 5 minutes instead of 70 minutes).

In the customer case for the bug, assisted-service is unavailable, but k8s is available. In this case, we use the data in k8s to determine whether the taint should be removed and try to move forward and allow CSR approval to run so that the remaining nodes can join the cluster and have the cluster installation succeed.

src/main/assisted-installer-controller/assisted_installer_main.go
rwsu: I think today, in non agent flow, the controller fails and restarts again. But yes, let's log that cluster wasn't able to be retrieved to aid in debugging.

src/main/assisted-installer-controller/assisted_installer_main.go
rwsu: I'm going to revert this function back into main and just return instead of calling os.Exit(0).

src/main/assisted-installer-controller/assisted_installer_main.go
tsorya: maybe lets create new one that just waits? i mean you can fail here cause of 2 different things. 
polling service and polling k8s, maybe just try to get cluster from assisted?

src/common/common.go
```diff
@@ -248,35 +251,84 @@ func parseOpenshiftVersionIntoMajorMinorZOnly(version string) string {
...
 
 // HasValidvSphereCredentials returns true if the the platform is
-// vSphere and the install config overrides contains real
+// vSphere and the install config or equivalent overrides contains real
 // credential values and not placeholder values.
 // Deprecated credential fields are not considered valid.
-func HasValidvSphereCredentials(ctx context.Context, ic inventory_client.InventoryClient, log logrus.FieldLogger) bool {
+func HasValidvSphereCredentials(ctx context.Context, ic inventory_client.InventoryClient, kc k8s_client.K8SClient, log logrus.FieldLogger) bool {
 	cluster, callErr := ic.GetCluster(ctx, false)
+	if cluster != nil {
```
tsorya, line 259: maybe lets add comment here that describes why it is ok not to have cluster?

tsorya: Not relevant to this pr but i wonder why we could create some secret with those credentials as manifest and reference it to controller?

src/main/assisted-installer-controller/assisted_installer_main.go
rwsu: I think it is fine they way it is actually. For the common case, it should check that both assisted-service and kube-apiserver are ready.

If assisted-service is ready but the k8s poll fails, then it should wait until it succeeds, otherwise the next step where it tries the determine the platform type through the ConfigMap would fail.

If assisted-service is not ready, then the check on kube-apiserver would be skipped. I'm not sure if it is worth adding a check or wait here for kube-apiserver. If it is not ready, then the controller will fail and restart. 

Does that sound ok to you?

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -137,22 +137,55 @@ func main() {
...
+				logger.Warnf("error determining platform type from install-config: %v", err)
+			} else {
+				platformType = pt
+				logger.Infof("platform type from install-config: %v", platformType)
+			}
+		}
+	} else {
+		cluster = assistedController.SetReadyState(assistedinstallercontroller.WaitTimeout)
+		if cluster == nil {
+			logger.Errorf("Failed to retrieve cluster from assisted service")
```
tsorya, line 181: Was sure i asked it before but in general we should fail here i think , previously we would fail with nil pointer exception but now we will continue without setting ready.
Let change it to Fatal

rwsu: Sure, changed to Fatal.

src/common/common.go
andfasano: (but I also understand that it may require additional work, so don't want to block this PR)

PR #899, opened 2024-09-06
WIP: overlay node image before bootstrapping if necessary

jlebon:
As per https://github.com/openshift/enhancements/pull/1637, we're trying
to get rid of all OpenShift-versioned components from the bootimages.

This means that there will no longer be oc, kubelet, or crio
binaries for example, which bootstrapping obviously relies on.

To adapt to this, the OpenShift installer now ships a new
`node-image-overlay.service` in its bootstrap Ignition config. This
service takes care of pulling down the node image and overlaying it,
effectively updating the system to the node image version.

Here, we accordingly also adapt assisted-installer so that we run
`node-image-overlay.service` before starting e.g. `kubelet.service` and
`bootkube.service`.

See also: https://github.com/openshift/installer/pull/8742


src/installer/installer.go
```diff
@@ -410,6 +410,18 @@ func (i *installer) startBootstrap() error {
...
+	// If we're in a pure RHEL/CentOS environment, we need to overlay the node image
+	// first to have access to e.g. oc, kubelet, cri-o, etc...
+	// https://github.com/openshift/enhancements/pull/1637
+	if !i.ops.FileExists("/usr/bin/oc") {
+		err = i.ops.SystemctlAction("start", "node-image-overlay.service")
+		if err != nil {
+			return err
+		}
+
+		i.waitForNodeImageOverlay(context.Background())
```
paul-maidment, line 422: Should there be a check performed after this wait to ensure that `/usr/bin/oc` is available as expected?

jlebon:
/remove-lifecycle stale

Now that https://github.com/openshift/enhancements/pull/1637 has merged, we will need this. I'll leave it in draft for now until I can rebase and retest it, but feel free to start reviewing.

jlebon:
Rebased and verified this still works! Requires https://github.com/openshift/installer/pull/8742.

src/installer/installer.go
```diff
@@ -669,6 +685,25 @@ func (i *installer) waitForBootkube(ctx context.Context) {
 	}
 }
 
+func (i *installer) waitForNodeImageOverlay(ctx context.Context) {
```
tsorya, line 688: Probably it can be change to just call utils.WaitForPredicate with 
i.ops.SystemctlAction("is-active", "node-image-overlay.service")
WDYT?

src/ops/ops.go
```diff
@@ -648,11 +648,12 @@ func (o *ops) CreateManifests(kubeconfig string, content []byte) error {
 	return nil
 }
 
-// DryRebootHappened checks if a reboot happened according to a particular reboot marker path
-// The dry run installer creates this file on "Reboot" (instead of actually rebooting)
-// We use this function to check whether the given node in the cluster have already rebooted
-func (o *ops) DryRebootHappened(markerPath string) bool {
```
tsorya, line 654: IMO lets leave this one as it helps to understand that we are looking for dry reboot logic, it can though call for FileExists

src/installer/installer.go
```diff
@@ -453,6 +453,22 @@ func (i *installer) startBootstrap() error {
...
+	// https://github.com/openshift/enhancements/pull/1637
+	if !i.ops.FileExists("/usr/bin/oc") {
+		err = i.ops.SystemctlAction("start", "node-image-overlay.service")
+		if err != nil {
+			return err
+		}
+
+		i.waitForNodeImageOverlay(context.Background())
+
+		if !i.ops.FileExists("/usr/bin/oc") {
```
tsorya, line 467: lets make /usr/bin/oc to be a const?

src/installer/installer.go
```diff
@@ -453,6 +453,22 @@ func (i *installer) startBootstrap() error {
 		return err
 	}
 
+	// If we're in a pure RHEL/CentOS environment, we need to overlay the node image
+	// first to have access to e.g. oc, kubelet, cri-o, etc...
+	// https://github.com/openshift/enhancements/pull/1637
+	if !i.ops.FileExists("/usr/bin/oc") {
```
tsorya, line 459: Shouldn't we pull overlay the node image as a first thing? I mean we already applied bootstrap ignition at this point and this can be a little bit late for us to pull another image layer, we probably should do it before

src/installer/installer.go
carbonin: What happens if the overlay service fails? Should this be able to detect that and return an error?

src/installer/installer.go
```diff
@@ -453,6 +453,22 @@ func (i *installer) startBootstrap() error {
...
+		err = i.ops.SystemctlAction("start", "node-image-overlay.service")
+		if err != nil {
+			return err
+		}
+
+		i.waitForNodeImageOverlay(context.Background())
+
+		if !i.ops.FileExists("/usr/bin/oc") {
+			return stderrors.New("/usr/bin/oc still doesn't exist after node image overlay")
```
carbonin, line 468: Is this the only failure mode? Is there any way this does exist, but something else didn't get extracted properly?

I'd be more comfortable if we could check the actual status of the services to ensure nothing failed.

src/installer/installer.go
jlebon: The code to pull the node image is part of the bootstrap Ignition so we can't use it before that.

jlebon:
Updated this now!
- Restored `DryRebootHappened()`
- Switched to `utils.WaitForPredicate()`
- Made various strings into constants
- Added stronger error-handling. I've tested various error cases. At the `openshift-install` level, it looks like:

```
INFO Host: master-0, reached installation stage Installing: bootstrap
INFO Host: master-0, reached installation stage Failed: service node-image-overlay.service failed
```

and then the host logs have more information about the failure.

src/installer/installer.go
```diff
@@ -453,6 +456,30 @@ func (i *installer) startBootstrap() error {
 		return err
 	}
 
+	// If we're in a pure RHEL/CentOS environment, we need to overlay the node image
```
paul-maidment, line 459: What happens if we are not in a pure RHEL/Centos environment?
Is this a possible scenario. What should we do in those cases?

jlebon:
OK, fixed a lint, a unit test, and added two unit tests (one positive and one negative) that checks the new path!

carbonin (approved):

carbonin:
Looks good to me, thanks for taking this on @jlebon!

jlebon:
Hmm, `edge-e2e-ai-operator-ztp` failing with

```
 [2025-01-20 21:43:45] ++ echo '2025-01-20 21:43:45+00:00 ERROR: failed Waiting for crd/agentserviceconfigs.agent-install.openshift.io on namespace '
[2025-01-20 21:43:45] 2025-01-20 21:43:45+00:00 ERROR: failed Waiting for crd/agentserviceconfigs.agent-install.openshift.io on namespace 
[2025-01-20 21:43:45] ++ oc get crd/agentserviceconfigs.agent-install.openshift.io --namespace= -o json
[2025-01-20 21:43:45] Error from server (NotFound): customresourcedefinitions.apiextensions.k8s.io "agentserviceconfigs.agent-install.openshift.io" not found 
```

Which I don't _think_ is related to this change. OK yeah, I see a similar looking failure in e.g. https://github.com/openshift/assisted-installer/pull/1003.

PR #904, opened 2024-09-18
OCPBUGS-42156: Switch to github.com/docker/distribution/reference to Mitigate CVE-2024-3727

paul-maidment:
The library github.com/containers/image/v5 has a vulnerability that has as of yet been unresolved.

Thankfully, it is possible to change the part of the library that we use

We can change github.com/containers/image/v5/docker/reference for github.com/docker/distribution/reference

In the case of assisted-installer, we achieve this by changing the dependency to the latest assisted-installer


adriengentil (approved):

gamli75:
@zaneb @andfasano can you help with the required label? thanks

PR #905, opened 2024-09-19
MGMT-18965: Update docker files to use universal base image

paul-maidment:
This updates the docker files of this project to use the universal base image

### DIFF
--- a/Dockerfile.assisted-installer
+++ b/Dockerfile.assisted-installer
@@ -1,7 +1,10 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.20 AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20 AS builder
 ARG TARGETPLATFORM
 ENV GOFLAGS=-mod=mod
 WORKDIR /go/src/github.com/openshift/assisted-installer
+
+# Workaround for creating build folder
+USER root
 
 # Bring in the go dependencies before anything else so we can take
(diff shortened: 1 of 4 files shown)


Dockerfile.assisted-installer
```diff
@@ -1,16 +1,20 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.20 AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20 AS builder
 ARG TARGETPLATFORM
 ENV GOFLAGS=-mod=mod
 WORKDIR /go/src/github.com/openshift/assisted-installer
 
+# Workaround for creating build folder
```
rccrdpccl, line 6: Why using a workaround when we can use this image the way it's meant to?

danmanor:
@paul-maidment  Can you explain the reasoning for this branch ?

paul-maidment:
> @paul-maidment Can you explain the reasoning for this branch ?

The 4.16 branch is broken unless we do something, try running

`skipper make build` on it

Presumably moving to UBI would be the correct course of action?

Dockerfile.assisted-installer
paul-maidment: It's not being used correctly on master?

https://github.com/openshift/assisted-installer/blob/master/Dockerfile.assisted-installer#L7

paul-maidment:
No longer needed, it's possible to work around this for now

PR #906, opened 2024-09-23
OCPBUGS-42294: Switch to github.com/docker/distribution/reference to Mitigate CVE-2024-3727

paul-maidment:
The library github.com/containers/image/v5 has a vulnerability that has as of yet been unresolved.

Thankfully, it is possible to change the part of the library that we use

We can change github.com/containers/image/v5/docker/reference for github.com/docker/distribution/reference

This PR bumps to the latest 4.16 version of the service to pick up the changes already made there

Additionally, minimal dockerfile changes have been made to ensure that we are able to build and test this change locally.


adriengentil (approved):

Dockerfile.assisted-installer-build
```diff
@@ -1,14 +1,34 @@
...
 
-RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/bin v1.53.2
-RUN yum install -y docker && \
-    yum clean all
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.53.2 && \ 
+  go install golang.org/x/tools/cmd/goimports@v0.1.0 && \
+  go install github.com/onsi/ginkgo/ginkgo@v1.16.1 && \
+  go install github.com/golang/mock/mockgen@v1.6.0 && \
+  go install gotest.tools/gotestsum@v1.6.3 && \
+  go install github.com/axw/gocov/gocov@latest && \
```
rccrdpccl, line 10: I understand this is not critical, but why latest? can't we pin it?

PR #910, opened 2024-09-25
OCPBUGS-41811: (agent-based installer) let the bootstrap wait for workers before the reboot

andfasano:
As described by the analysis in https://issues.redhat.com/browse/OCPBUGS-41811, in some cases the bootstrap node may reboot before the workers started the joining process, thus removing the assisted-service that it's still required by the workers. This prevents the worker to successfully join the cluster, causing the failure of the cluster deployment.

This patch introduces an explicit synchronization between the bootstrap node and the workers (only in case the installation was performed via the agent-based installer), delaying the bootstrap reboot until all the workers passed the `waiting for control plane` stage.

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -172,6 +172,13 @@ func (i *installer) InstallNode() error {
 			return err
 		}
 		i.log.Info("Setting bootstrap node new role to master")
+
+		// In the agent-based installer the assisted-service runs on the bootstrap node, so let's
+		// wait for the workers (if any) to kick-off the joining process before rebooting the
+		// bootstrap node.
+		if err = i.waitForWorkers(ctx); err != nil {
+			return err
+		}
 	}
 
 	//upload host logs and report log status before reboot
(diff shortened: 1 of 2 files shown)


src/installer/installer.go
```diff
@@ -184,6 +191,42 @@ func (i *installer) InstallNode() error {
...
+		workers, err := i.inventoryClient.ListsHostsForRole(ctx, string(models.HostRoleWorker))
+		if err != nil {
+			i.log.WithError(err).Error("failed to get workers")
+			return false
+		}
+		if len(workers) == 0 {
+			return true
+		}
+
+		i.log.Infof("Found %d workers, verifying their current installation stage", len(workers))
```
rwsu, line 215: From just reading the log output, it is unclear what the wait is for. Perhaps change this log message to indicate that we are waiting for the worker to reach reboot.
```suggestion
		i.log.Infof("Found %d workers, waiting for workers to reach reboot stage", len(workers))
```

rwsu (changes requested):

src/installer/installer.go
```diff
@@ -184,6 +191,42 @@ func (i *installer) InstallNode() error {
...
+		}
+		if len(workers) == 0 {
+			return true
+		}
+
+		i.log.Infof("Found %d workers, waiting for workers to reach reboot stage", len(workers))
+		for _, worker := range workers {
+			// If any worker is still in the waiting for control plane stage, then
+			// it means it didn't reboot yet and still requires assisted-service
+			if worker.Progress.CurrentStage == models.HostStageWaitingForControlPlane {
```
eranco74, line 219: The worker might have an earlier stage (e.g. HostStageInstalling in case of a very slow disk)
So you might want to change this: 
funk.Contains([]models.HostStage{models.HostStageWaitingForControlPlane, models.HostStageInstalling, models.HostStageStartingInstallation) c.host.Progress.CurrentStage)

eranco74 (commented):
/lgtm

src/installer/installer.go
```diff
@@ -172,6 +172,13 @@ func (i *installer) InstallNode() error {
 			return err
 		}
 		i.log.Info("Setting bootstrap node new role to master")
+
+		// In the agent-based installer the assisted-service runs on the bootstrap node, so let's
+		// wait for the workers (if any) to kick-off the joining process before rebooting the
+		// bootstrap node.
+		if err = i.waitForWorkers(ctx); err != nil {
```
tsorya, line 179: this is relevant only for ABI no ? There is no point to wait for worker in regular flow

tsorya: found https://github.com/openshift/assisted-installer/pull/910/files#diff-1046cc2d18cf5f82336bbad36a2d28540606e1c6aaa0b5073c545301ef60ffd4R201

PR #932, opened 2024-10-15
NO-ISSUE: Bump PGX version

paul-maidment:
The PGX version has been bumped in all branches, somehow this one was missed


rccrdpccl:
We should not need pg stuff in assisted-installer, right? It seems that this is necessary because we import assisted-service's common package... If we could get rid of this nasty dependencies it would make our life easier https://issues.redhat.com/browse/MGMT-19139

Unfortunately it's not straightforward so we have to merge this
/lgtm
/approve

PR #936, opened 2024-10-21
MGMT-19148: Update assisted-service dependencies.

jhernand:
This patch updates the project to the latest version of the assisted-service dependencies. The reason to do that is to be able to use the new _Waiting for OLM operator setup jobs_ finalizing stage.

In addition to updating the assisted-service dependencies it is also necessary to update the `docker/cli` and `docker/docker` packages, because the versions currently used cause build errors.

Also the `baremetal-operator` dependencies need to be replaced by the versions that the updates assisted service needs.

Related: https://issues.redhat.com/browse/MGMT-19148
Related: https://issues.redhat.com/browse/MGMT-19056


CrystalChun (approved):
/approve
/lgtm

PR #938, opened 2024-10-25
MGMT-19148: Add options parameter to `ListJobs`

jhernand:
This patch adds a new `options metav1.ListOptions` parameter to the `ListJobs` method of Kubernetes client interface. This new parameter will be needed in order to add support for waiting for operator setup jobs, because that will need to list jobs in all namespaces using a label selector.

In addition to adding the options the patch also changes the mock configuration so that they are more precise. Instead of this:

```go
mockk8sclient.EXPECT().ListJobs(gomock.Any(), gomock.Any())
```

We use this now:

```go
mockk8sclient.EXPECT().ListJobs(olmNamespace, metav1.ListOptions{})
```

That will help when introducing other tests where the parameters will be
different.

Related: https://issues.redhat.com/browse/MGMT-19148
Related: https://issues.redhat.com/browse/MGMT-19056


jhernand:
This is no longer needed because #940 was closed.

PR #940, opened 2024-11-11
MGMT-19148: Wait for OLM operator setup jobs

jhernand:
When operators are automatically enabled from assisted installer the custom manifests may contain Kubernetes jobs to complete the setup. For example, a hypothetical storage operator can include in the custom manifests a job that waits for the storage class and marks it as the default. Currently the installer doesn't wait for jobs to finish, and it will declare the cluster to be ready even if those jobs haven't been completed. This patch changes the controller so that it assumes that jobs that have the `agent-install.openshift.io/setup-job` label are such setup jobs, and waits for them to finish. The value of the label must be the name of the corresponding operator.

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -8,11 +8,13 @@ import (
 	stderrors "errors"
 	"fmt"
 	"io"
+	"maps"
 	"math"
 	"net"
 	"net/http"
 	"os"
 	"path"
+	"sort"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -25,6 +27,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	batchV1 "k8s.io/api/batch/v1"
 	certificatesv1 "k8s.io/api/certificates/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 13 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -1506,3 +1543,110 @@ func (c controller) SetReadyState(waitTimeout time.Duration) *models.Cluster {
 func (c *controller) GetStatus() *ControllerStatus {
 	return c.status
 }
+
+// waitForSetupJobs waits till the jobs are completed successfully.
+func (c *controller) waitForSetupJobs(ctx context.Context) error {
```
tsorya, line 1548: Can't we make it part of waiting for specific operator in operator handler?

src/assisted_installer_controller/operator_handler.go
```diff
@@ -48,7 +49,22 @@ func (c *controller) isOperatorAvailable(handler OperatorHandler) bool {
 			return false
 		}
 
-		err = c.ic.UpdateClusterOperator(context.TODO(), c.ClusterID, operatorName, operatorVersion, operatorStatus, operatorMessage)
+		// If the operator is available but has setup jobs then we don't want to mark it as available here
+		// because if we do that for all operators then the service will consider that the cluster is ready
+		// and that will in turn stop the controller without first checking if the setup jobs have finished.
+		// So instead we mark the operator as progressing, and will mark it as available later, when we check
+		// the setup jobs.
+		var reportedStatus models.OperatorStatus
```
tsorya, line 57: I believe all setup jobs logic should be here, if operator has setup job don't set it as available but i don't think we should split the logic

jhernand:
I am closing this because it doesn't bring a real benefit. Operators can still include jobs to complete their setup, we will just not explicitly wait for them.

PR #947, opened 2024-12-01
OCPBUGS-45116: Add retry to mount efi device

giladravid16:
There is a race condition between the mount and setting the boot order, this pr adds a retry to the mount operation.


giladravid16:
/override ci/prow/okd-scos-e2e-aws-ovn

This test is broken, the build for agent-installer-csr-approver doesn't work

PR #979, opened 2024-12-16
[release-ocm-2.10] ACM-16418: Bump golang.org/x/crypto from 0.25.0 to 0.31.0

adriengentil:
Downgraded hive as it is not compatible with the assisted-service on release-ocm-2.10. For some reason the version of hive in use in this branch is higher than the one in used release-ocm-2.11/2.12.


pastequo:
Should we pin the version of hive to be sure it's not upgraded again ? (with a replace directive)

adriengentil:
> Should we pin the version of hive to be sure it's not upgraded again ? (with a replace directive)

not sure it would help, it was updated by humans targeting assisted-service in master instead of ocm branches.

gamli75 (approved):

PR #1003, opened 2025-01-09
MGMT-19100: Install to the current boot device when CoreosImage is set

carbonin:
This indicates the container image that should be installed and booted for the installed host. It also indicates that we should install to the currently booted disk, in a new stateroot, rather than expecting a device path.

https://issues.redhat.com/browse/MGMT-19100

cc @tsorya @rccrdpccl @eranco74

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -155,14 +155,17 @@ func (i *installer) InstallNode() error {
 		}
 	}
 
-	if i.EnableSkipMcoReboot {
-		i.skipMcoReboot(ignitionPath)
-	}
+	// we can't do any of this if we're installing to the existing root
+	if i.Config.CoreosImage == "" {
+		if i.EnableSkipMcoReboot {
+			i.skipMcoReboot(ignitionPath)
+		}
 
-	if err = i.ops.SetBootOrder(i.Device); err != nil {
-		i.log.WithError(err).Warnf("Failed to set boot order")
-		// Ignore the error for now so it doesn't fail the installation in case it fails
-		//return err
+		if err = i.ops.SetBootOrder(i.Device); err != nil {
+			i.log.WithError(err).Warnf("Failed to set boot order")
+			// Ignore the error for now so it doesn't fail the installation in case it fails
+			//return err
+		}
 	}
 
 	if isBootstrap {
@@ -321,7 +324,7 @@ func convertToOverwriteKargs(args []string) []string {
 }
 
 func (i *installer) cleanupInstallDevice() {
-	if i.DryRunEnabled || i.Config.SkipInstallationDiskCleanup {
+	if i.DryRunEnabled || i.Config.SkipInstallationDiskCleanup || i.Config.CoreosImage != "" {
 		i.log.Infof("skipping installation disk cleanup")
 	} else {
 		err := i.cleanupDevice.CleanupInstallDevice(i.Config.Device)
@@ -365,7 +368,11 @@ func (i *installer) writeImageToDisk(ignitionPath string) error {
 	interval := time.Second
 	liveLogger := coreos_logger.NewCoreosInstallerLogWriter(i.log, i.inventoryClient, i.Config.InfraEnvID, i.Config.HostID)
 	err := utils.Retry(3, interval, i.log, func() error {
-		return i.ops.WriteImageToDisk(liveLogger, ignitionPath, i.Device, i.Config.InstallerArgs)
+		if i.Config.CoreosImage == "" {
+			return i.ops.WriteImageToDisk(liveLogger, ignitionPath, i.Device, i.Config.InstallerArgs)
+		} else {
+			return i.ops.WriteImageToExistingRoot(liveLogger, ignitionPath)
+		}
 	})
 	if err != nil {
 		i.log.WithError(err).Error("Failed to write image to disk")
@@ -1044,7 +1051,11 @@ func RunInstaller(installerConfig *config.Config, logger *logrus.Logger) error {
 	)
 
 	// Try to format requested disks. May fail formatting some disks, this is not an error.
-	ai.FormatDisks()
+	if installerConfig.CoreosImage == "" {
+		ai.FormatDisks()
+	} else {
+		logger.Info("not attempting disk format when installing to boot device")
+	}
 
 	if err = ai.InstallNode(); err != nil {
 		ai.UpdateHostInstallProgress(models.HostStageFailed, err.Error())
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -38,6 +38,7 @@ import (
 )
 
 const (
+	dockerConfigFile                = "/root/.docker/config.json"
 	coreosInstallerExecutable       = "coreos-installer"
 	dryRunCoreosInstallerExecutable = "dry-installer"
 	encapsulatedMachineConfigFile   = "/etc/ignition-machine-config-encapsulated.json"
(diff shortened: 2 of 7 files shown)


src/ops/ops.go
```diff
@@ -124,6 +125,61 @@ func (o *ops) SystemctlAction(action string, args ...string) error {
 	return errors.Wrapf(err, "Failed executing systemctl %s %s", action, args)
 }
 
+func (o *ops) WriteImageToLocalDevice(liveLogger io.Writer, ignitionPath string, rhelCoreOSImage string) error {
```
paul-maidment, line 128: Does any test provide coverage of this function?

I can see a code path in here that would conditionally execute according to the outcome of some the the calls to ExecPrivilegeCommand

I also think this is quite mockable/testable and there are pre-existing examples of how to approach this.

paul-maidment (changes requested):

src/ops/ops.go
carbonin: I didn't see much value since there's no actual logic here and every call would be mocked.

I could pull out the ostree output parsing bit since that's the only bit worth testing, otherwise it doesn't feel worth it to me.

paul-maidment: I think the parsing part should be pulled out so that we can validate the parsing

src/ops/ops.go
```diff
@@ -124,6 +125,61 @@ func (o *ops) SystemctlAction(action string, args ...string) error {
...
+		return errors.Wrapf(err, "failed creating new stateroot: %s", out)
+	}
+
+	ostreeReleasePullSpec := fmt.Sprintf("ostree-unverified-registry:%s", rhelCoreOSImage)
+	out, err = o.ExecPrivilegeCommand(liveLogger, "ostree", "container", "unencapsulate", "--authfile", "/root/.docker/config.json", "--quiet", "--repo", "/ostree/repo", ostreeReleasePullSpec)
+	if err != nil {
+		return errors.Wrapf(err, "failed to unencapsulate rhcos payload image: %s", out)
+	}
+	outputParts := strings.Split(out, " ")
+	if len(outputParts) != 2 {
```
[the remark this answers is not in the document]
carbonin: That's fair. I'll build a regex of some sort. The expected output will be in the tests when I write them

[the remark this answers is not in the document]
carbonin: FWIW I don't know what all the output could possibly be.

My guess is that if the command is successful we'd only ever see the two parts we're expecting, but better to be sure.

src/ops/ops.go
```diff
@@ -49,6 +49,7 @@ const (
 type Ops interface {
 	Mkdir(dirName string) error
 	WriteImageToDisk(liveLogger io.Writer, ignitionPath string, device string, extraArgs []string) error
+	WriteImageToLocalDevice(liveLogger io.Writer, ignitionPath string, rhelCoreOSImage string) error
```
carbonin, line 52: Any ideas for a better name for this?

Technically all the devices are "local". Maybe "BootDevice"? Is that better?

src/ops/ops.go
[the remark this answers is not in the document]
carbonin: Also just for reference, this is the line I'm parsing https://github.com/ostreedev/ostree-rs-ext/blob/b885f9468829e6af41fa95504d65975c48378d4d/lib/src/cli.rs#L710

src/ops/ops.go
paul-maidment: BootDevice seems reasonable.

We are describing the disk we will boot into once the node reboots out of discovery?

carbonin: Yes, but that's true in either case.

The distinction is if we're installing to a device that is currently being used for the rootfs or one that is not.

paul-maidment: So in both cases they are boot disks, so how about 

`WriteImageToBootDiskWithRootFs` ?

carbonin: Maybe `WriteImageToExistingRoot` to mirror the similar `bootc` command (which might be something I can use) called [`bootc install to-existing-root`](https://containers.github.io/bootc/man/bootc-install-to-existing-root.html)

src/ops/ops.go
```diff
@@ -124,6 +125,71 @@ func (o *ops) SystemctlAction(action string, args ...string) error {
...
+	o.log.Infof("imported commit %s", commit)
+
+	out, err = o.ExecPrivilegeCommand(liveLogger, "ostree", "admin", "deploy",
+		"--stateroot", "install",
+		"--karg", "$ignition_firstboot",
+		"--karg", defaultIgnitionPlatformId,
+		commit)
+	if err != nil {
+		return errors.Wrapf(err, "failed to deploy commit to stateroot: %s", out)
+	}
```
carbonin, line 173: Related to https://github.com/openshift/assisted-installer/pull/1003#discussion_r1913421655 ...

@eranco74 (or someone you think knows) can/should we use `bootc` here?

src/installer/installer.go
```diff
@@ -152,14 +152,17 @@ func (i *installer) InstallNode() error {
 		}
 	}
 
-	if i.EnableSkipMcoReboot {
-		i.skipMcoReboot(ignitionPath)
-	}
+	// we can't do any of this if we're installing to the existing root
+	if i.Config.CoreosImage == "" {
```
tsorya, line 156: maybe i.Config.CoreosImage == "" && i.EnableSkipMcoReboot ?

src/ops/ops.go
tsorya: Bootc is already part of rhcos?

carbonin: I'm not sure, that's part of what I'm asking, but I can check.

carbonin: Yeah, seems to be there in 4.18 at least 
image

carbonin: It's also present in 4.17 and 4.16. 4.15 is the latest version that doesn't have it.

carbonin: I think MCE won't support anything earlier than 4.16 by the time this change gets in so we might be safe there, but if we also want to use this on the SaaS it might not be worth limiting the versions we can install.

Even if `bootc` works for this and is the better option maybe we can leave it as a follow up for now?

src/ops/ops.go
```diff
@@ -124,6 +125,71 @@ func (o *ops) SystemctlAction(action string, args ...string) error {
...
+	commit, err := o.importOSTreeCommit(liveLogger)
+	if err != nil {
+		return err
+	}
+	o.log.Infof("imported commit %s", commit)
+
+	out, err = o.ExecPrivilegeCommand(liveLogger, "ostree", "admin", "deploy",
+		"--stateroot", "install",
+		"--karg", "$ignition_firstboot",
+		"--karg", defaultIgnitionPlatformId,
```
rccrdpccl, line 169: Shouldn't we pass platform ID based on the platform of the target image?

src/ops/ops.go
```diff
@@ -124,6 +125,71 @@ func (o *ops) SystemctlAction(action string, args ...string) error {
 	return errors.Wrapf(err, "Failed executing systemctl %s %s", action, args)
 }
 
+var ostreeOutputRegex = regexp.MustCompile(`Imported: (\w+)`)
+
+func (o *ops) importOSTreeCommit(liveLogger io.Writer) (string, error) {
+	ostreeReleasePullSpec := fmt.Sprintf("ostree-unverified-registry:%s", o.installerConfig.CoreosImage)
+	out, err := o.ExecPrivilegeCommand(liveLogger, "ostree", "container", "unencapsulate", "--authfile", "/root/.docker/config.json", "--quiet", "--repo", "/ostree/repo", ostreeReleasePullSpec)
```
rccrdpccl, line 132: nit: this is probably not gonna change often (or possibly never), however wouldn't the docker config file path be better in a constant? (there's already one in installer package, should we share such constant - in a different package of course - as we are relying that this file has been extracted there?)

src/ops/ops.go
carbonin: What do you think for this one? We can always address bugs for this later if we run into anything.

[The other place that does this](https://github.com/openshift/assisted-installer/blob/ba4f15f38a7493893701bdd9c0e90ce5a41adb42/src/ops/ops.go#L953) is looking at the current kernel command line for platform to check, but I'm not sure we're really doing the same thing.

src/ops/ops.go
carbonin: Leaving this as a potential follow up

eranco74 (commented):
/lgtm

eranco74:
@javipolo PTAL

PR #1014, opened 2025-01-22
MGMT-19621: Fix Race condition between the OLM operators monitor to the cluster installation monitor

danmanor:
## Introduction

The purpose of this PR is to resolve a bug caused by a race condition between the main thread
of assisted installer controller and a goroutine responsible for monitoring the status of 
OLM operators.

When the main thread reaches the stage where cluster installation has been deemed complete it 
will cancel the contexts of any Go routines it spawned, including the OLM operator monitor. 

We use cache upon the status of monitored OLM operators and this is relevant to this bug.
Upon shutdown, this monitor thread performs final checks of the installed operators, this final
check depends on a cached result, which, coupled with a bug that causes the operator installation status to
be inconsistent with cache content, leads to a scenario where the installer will report a complete installation
while also reporting incomplete operator installation.

The solution is twofold, 

* Firstly to fix the bug that causes the inconsistent installation status
* Secondly, as a precaution, anywhere we are taking our 'last chance' to obtain the status of the operators, this call
should not depend upon the cache. This will ensure that we make an 'up to date' assessment of the situation.

## More Details

The controller flow runs in parallel Go routines executing different actions. Two key routines are:

- **Cluster Installation Monitor**: Every minute, it queries assisted-service to check if the cluster installation is finished by evaluating its status - https://github.com/openshift/assisted-installer/blob/e9ae12da779dd7d2610f515605ea3ff66e8c4f4b/src/main/assisted-installer-controller/assisted_installer_main.go#L224.
- **OLM Operators Monitor**: Every 30 seconds, it checks the current status of each monitored operator resource. When any operator status changes, it updates assisted-service - https://github.com/openshift/assisted-installer/blob/e9ae12da779dd7d2610f515605ea3ff66e8c4f4b/src/assisted_installer_controller/assisted_installer_controller.go#L987 and https://github.com/openshift/assisted-installer/blob/e27ef9e2a68813b0167169439005d16682bdc889/src/assisted_installer_controller/operator_handler.go#L51. It is important to note that the controller uses a cache for querying cluster's monitored operators, with 20 seconds TTL. 

The issue lies in the predicate logic - https://github.com/openshift/assisted-installer/blob/e9ae12da779dd7d2610f515605ea3ff66e8c4f4b/src/assisted_installer_controller/assisted_installer_controller.go#L1001. This causes a scenario where there is always a 30-second delay between the time all operators are ready (and the assisted-service is updated accordingly) and when the function acknowledges the update. This leads to the possibility of the following sequence:

1. The operators monitor updates assisted-service to indicate all operators are available, but it won't exit until the next cycle (30 seconds later). 
2. Before the cache data becomes stale (20 seconds), the cluster installation monitor runs, querying assisted-service, which has changed its status from finalizing to installed (this happens when all hosts and operators are installed, and the state machine progresses).
3. The cluster monitor detects the installation as complete and cancels the controller context across all threads -https://github.com/openshift/assisted-installer/blob/e9ae12da779dd7d2610f515605ea3ff66e8c4f4b/src/main/assisted-installer-controller/assisted_installer_main.go#L195
4. The operators monitor receiving the cancel context, and attempts to update all operators which are still in `progressing` status as `failed` in assisted-service - https://github.com/openshift/assisted-installer/blob/e27ef9e2a68813b0167169439005d16682bdc889/src/assisted_installer_controller/assisted_installer_controller.go#L542 and https://github.com/openshift/assisted-installer/blob/e27ef9e2a68813b0167169439005d16682bdc889/src/assisted_installer_controller/assisted_installer_controller.go#L971
5. The controller is getting the last updated operator back in the list of `progressing` operators as the cache contains data from the last query - https://github.com/openshift/assisted-installer/blob/e27ef9e2a68813b0167169439005d16682bdc889/src/inventory_client/inventory_client.go#L275
6. The controller updates assisted-service with the operator status as failed - https://github.com/openshift/assisted-installer/blob/e27ef9e2a68813b0167169439005d16682bdc889/src/assisted_installer_controller/assisted_installer_controller.go#L977
7. The result is a cluster marked as installed with failed operators.

This PR resolves the issue by several steps:
1. Avoid using cache in queries which are not part of waiting predicates, as they occur once and stale data might have irreversible impact.
2. moving the check for non-ready operators inside the predicate after the list reduction, thereby eliminating the 30-second delay. This change ensures that the moment the cluster is ready in assisted-service, the operators monitor acknowledges it immediately.
3. Adds a log to increase observability.

### DIFF
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -546,11 +546,14 @@ func (c *controller) waitForOLMOperators(ctx context.Context) error {
 			c.log.WithError(err).Error("Timeout while waiting for some of the operators and not able to update its state")
 		}
 	}()
+
 	// Get the monitored operators:
 	c.updateFinalizingProgress(ctx, models.FinalizingStageWaitingForOlmOperatorsCsvInitialization)
 	retryErr = utils.WaitForeverForPredicateWithCancel(ctx, FetchRetryInterval, func() bool {
 		var err2 error
-		operators, err2 = c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID, c.OpenshiftVersion)
+		// We use the cache here to retrieve results from the assisted-service, as this function is executed repeatedly.
+		// Over time, this approach ensures that we ultimately get the correct result.
+		operators, err2 = c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID, c.OpenshiftVersion, true)
 		if err2 != nil {
 			c.log.WithError(err2).Errorf("Error fetching the monitored operators from assisted-service")
 			return false
@@ -560,6 +563,7 @@ func (c *controller) waitForOLMOperators(ctx context.Context) error {
 	if retryErr != nil {
 		return errors.Wrapf(retryErr, "Failed to fetch monitored operators")
 	}
+
 	if len(operators) == 0 {
 		c.log.Info("No OLM operators found.")
 		return nil
@@ -951,9 +955,9 @@ func (c *controller) addRouterCAToClusterCA() bool {
 
 }
 
-func (c *controller) getProgressingOLMOperators() ([]*models.MonitoredOperator, error) {
+func (c *controller) getProgressingOLMOperators(useCache bool) ([]*models.MonitoredOperator, error) {
 	ret := make([]*models.MonitoredOperator, 0)
-	operators, err := c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID, c.OpenshiftVersion)
+	operators, err := c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID, c.OpenshiftVersion, useCache)
 	if err != nil {
 		c.log.WithError(err).Warningf("Failed to connect to assisted service")
 		return ret, err
@@ -968,11 +972,17 @@ func (c *controller) getProgressingOLMOperators() ([]*models.MonitoredOperator, 
 
 func (c *controller) updatePendingOLMOperators(ctx context.Context) error {
 	c.log.Infof("Updating pending OLM operators")
-	operators, err := c.getProgressingOLMOperators()
+
+	// This call is executed only once during the application's graceful shutdown.
+	// Therefore, it is essential to retrieve up-to-date results on which operators are
+	// still progressing from the assisted-service. To ensure accuracy, we avoid using the cache in this case.
+	operators, err := c.getProgressingOLMOperators(false)
 	if err != nil {
 		return err
 	}
+
 	for _, operator := range operators {
+		c.log.Infof("Updating operator %s status to 'failed'", operator.Name)
 		c.status.OperatorError(operator.Name)
 		err := c.ic.UpdateClusterOperator(ctx, c.ClusterID, operator.Name, operator.Version, models.OperatorStatusFailed, "Waiting for operator timed out")
 		if err != nil {
@@ -980,12 +990,16 @@ func (c *controller) updatePendingOLMOperators(ctx context.Context) error {
 			return err
 		}
 	}
+
 	return nil
 }
 
 // waitForCSV wait until all OLM monitored operators are available or failed.
 func (c *controller) waitForCSV(ctx context.Context) error {
-	operators, err := c.getProgressingOLMOperators()
+	// This function is executed only once during the wait for OLM operators to become ready.
+	// As such, we need up-to-date information on which operators are still progressing.
+	// While no operators should yet be in a 'failing' or 'available' status at this point, avoiding the cache helps mitigate potential edge cases.
+	operators, err := c.getProgressingOLMOperators(false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -546,11 +546,12 @@ func (c *controller) waitForOLMOperators(ctx context.Context) error {
 			c.log.WithError(err).Error("Timeout while waiting for some of the operators and not able to update its state")
 		}
 	}()
+
 	// Get the monitored operators:
 	c.updateFinalizingProgress(ctx, models.FinalizingStageWaitingForOlmOperatorsCsvInitialization)
 	retryErr = utils.WaitForeverForPredicateWithCancel(ctx, FetchRetryInterval, func() bool {
 		var err2 error
-		operators, err2 = c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID, c.OpenshiftVersion)
+		operators, err2 = c.ic.GetClusterMonitoredOLMOperators(context.TODO(), c.ClusterID, c.OpenshiftVersion, true)
```
paul-maidment, line 554: Should we be passing context.TODO() here or should we update this to pass the context that was passed into the parent function?

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -968,24 +970,30 @@ func (c *controller) getProgressingOLMOperators() ([]*models.MonitoredOperator,
 
 func (c *controller) updatePendingOLMOperators(ctx context.Context) error {
 	c.log.Infof("Updating pending OLM operators")
-	operators, err := c.getProgressingOLMOperators()
+
+	// We don't want to use cache in this case as this function is executed once and we need to get up-to-date results.
```
paul-maidment, line 974: I understand that a function only executing once reduces the impact of not using cache.

But I think the primary motive is that we need up to date results. 

Does that sound fair?

paul-maidment: Why is it not important to get up to date results in the other locations?
Can we maybe add a comment in those locations to explain why we choose or do not choose cache for a given case. I think that this change is quite complex to read and understand so anything that aids clarification is good.

src/assisted_installer_controller/assisted_installer_controller.go
paul-maidment: Why is it important to use cache here?

src/assisted_installer_controller/operator_handler.go
```diff
@@ -62,7 +76,7 @@ func (c *controller) isOperatorAvailable(handler OperatorHandler) bool {
 }
 
 func (c *controller) isOperatorAvailableInService(operatorName string, openshiftVersion string) (*models.MonitoredOperator, bool) {
-	operatorStatusInService, err := c.ic.GetClusterMonitoredOperator(utils.GenerateRequestContext(), c.ClusterID, operatorName, openshiftVersion)
+	operatorStatusInService, err := c.ic.GetClusterMonitoredOperator(utils.GenerateRequestContext(), c.ClusterID, operatorName, openshiftVersion, true)
```
paul-maidment, line 79: For clarification, the reason we can use cache here is because this is about the availability/non availability of a given operator and this stays fairly constant throughout install, so no need to have an uncached query here?

paul-maidment (changes requested):
I think that the comments around cached/uncached queries need to be really clear so that anyone reading this code can clearly understand the intent.

The topic is quite complex and I'd maybe also take a look at how it's described in the description to see if we can communicate the concept more clearly.

paul-maidment:
Based on our discussion around this, thank you for clarifying the bugfix to me

Might I suggest that we prepend the PR description with some text along the lines of the following?
We can refine the actual wording but the general idea is to provide a more condensed summary prior to additional details, does this sound reasonable?

```
The purpose of this PR is to resolve a bug caused by a race condition between the main thread
of assisted installer controller and a goroutine responsible for monitoring the status of 
OLM operators.

When the main thread reaches the stage where cluster installation has been deemed complete it 
will cancel the contexts of any goroutines it spawned, including the OLM operator monitor

We use cache upon the status of monitored OLM operators and this is relevant to this bug.
Upon shutdown, this monitor thread performs final checks of the installed operators, this final
check depends on a cached result, which, coupled with a bug that causes the operator installation status to
be inconsistent with cache content, leads to a scenario where the installer will report a complete installation
while also reporting incomplete operator installation.

The solution is twofold, 

* Firstly to fix the bug that causes the inconsistent installation status
* Secondly, as a precaution, anywhere we are taking our 'last chance' to obtain the status of the operators, this call
should not depend upon the cache. This will ensure that we make an 'up to date' assesment of the situation.
```

paul-maidment (changes requested):
Looks generally good, I would just update the description to make the whole problem/ticket a bit clearer.

paul-maidment (approved):
/lgtm

src/inventory_client/inventory_client.go
```diff
@@ -270,10 +270,12 @@ func (c *inventoryClient) ListsHostsForRole(ctx context.Context, role string) (m
 	return ret.Payload, nil
 }
 
-func (c *inventoryClient) getMonitoredOperators(ctx context.Context, clusterId string) (models.MonitoredOperatorsList, error) {
+func (c *inventoryClient) getMonitoredOperators(ctx context.Context, clusterId string, useCache bool) (models.MonitoredOperatorsList, error) {
 	cacheKey := fmt.Sprintf("getMonitoredOperators-%s", clusterId)
-	if val, err := c.cache.Get(cacheKey); err == nil {
-		return val.(models.MonitoredOperatorsList), nil
+	if useCache {
```
paul-maidment, line 275: We could reduce levels a bit by making this part of the if condition on line 276

PR #1016, opened 2025-01-24
MGMT-19101: Add support to existing root install for installer kargs args

carbonin:
The installer args are now passed to the existing root install function and if any are --append-karg or --remove-karg the corresponding arg is added to the ostree command args.

https://issues.redhat.com/browse/MGMT-19101


carbonin:
Closing this, going to create a new one including the kargs and networking stuff.

PR #1018, opened 2025-01-28
MGMT-19101: Respect installer args

carbonin:
This PR adds support to the existing rootfs install strategy for some of the `coreos-installer` args that we allow users to specify.

Specifically `--append-karg`, `--remove-karg`, and `--copy-network`.

For the karg ones, the corresponding arg is added to the ostree command args which handles editing the boot loader configs.

For the copy network, the network manager keyfiles  are copied into `/boot/coreos-firstboot-network` which is the same thing `coreos-installer` does when provided `--copy-network`

Resolves https://issues.redhat.com/browse/MGMT-19101

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -16,6 +16,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"text/template"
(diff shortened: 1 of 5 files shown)


src/ops/ops.go
```diff
@@ -165,15 +187,24 @@ func (o *ops) WriteImageToExistingRoot(liveLogger io.Writer, ignitionPath string
...
 	}
 
+	if slices.Contains(installerArgs, "--copy-network") || slices.Contains(installerArgs, "-n") {
+		o.log.Info("copying network files from /etc/NetworkManager/system-connections to /boot/coreos-firstboot-network")
+		out, err = o.ExecPrivilegeCommand(liveLogger, "mkdir", "/boot/coreos-firstboot-network")
+		if err != nil {
+			return errors.Wrapf(err, "failed to create firstboot network directory: %s", out)
+		}
+
+		out, err = o.ExecPrivilegeCommand(liveLogger, "sh", "-c", "cp /etc/NetworkManager/system-connections/* /boot/coreos-firstboot-network")
```
carbonin, line 202: FWIW this was done this way because without the additional shell the glob was getting evaluated in the container, then the command failed when run on the host. I had to do it this way to ensure the glob was evaluated on the host.

I can add a comment if folks think it's worth it.

rccrdpccl, line 202: nit: probably paths should be a constant

PR #1024, opened 2025-01-29
MGMT-19819: Add the commit reference from which the image is built to the image

danmanor:
Currently, in most of assisted installer components CI images we don't have a way to tell from which commit reference the image was built. Since We use an image stream for each component, and we import these streams from one CI component configuration to another, we might end up with images to are not up-to-date. In this case, we would like to have the ability to check if this is actually the case.

### DIFF
--- a/Dockerfile.assisted-installer
+++ b/Dockerfile.assisted-installer
@@ -16,6 +16,9 @@ COPY . .
 RUN git config --global --add safe.directory '*'; \
     TARGETPLATFORM=$TARGETPLATFORM make installer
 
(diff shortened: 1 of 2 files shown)


carbonin:
Would it make sense to add this as an image label in addition to (or instead of) the file?

It would save a step when checking this I guess.

danmanor:
@carbonin I tried that approach initially as this makes more sense to me as well, but because of Prow limitations I didn't find a way to dynamically add this label :(

CrystalChun (approved):
/lgtm

PR #1027, opened 2025-02-06
MGMT-19421: phase 1 of replacing the refference HighAvailabilityMode with ControlPlaneCount

andrej1991:
phase 1 of replacing the refference HighAvailabilityMode with ControlPlaneCount
in the first stage it still accepst --high-availability-mode as a command line argument, but it is being converted to ControlPlaneCount in the following maner:
- ControlPlaneCount is being set to 3 if --high-availability-mode is set to full
- ControlPlaneCount is being set ot 1 if --high-availability-mode is set to none
After the phase 1 of assisted-installer-agent is being merged --high-availability-mode will be fully removed from the valid parameters.

### DIFF
--- a/deploy/assisted-installer-controller/assisted-installer-controller-cm.yaml.template
+++ b/deploy/assisted-installer-controller/assisted-installer-controller-cm.yaml.template
@@ -11,5 +11,5 @@ data:
   skip-cert-verification: '{{.SkipCertVerification}}'
   ca-cert-path: '{{.CACertPath}}'
   check-cluster-version: '{{.CheckCVO}}'
-  control-plane-count: {{.HaMode}}
+  control-plane-count: '{{.ControlPlaneCount}}'
   must-gather-image: '{{.MustGatherImage}}'
--- a/deploy/assisted-installer-controller/assisted-installer-controller-cm.yaml.template
+++ b/deploy/assisted-installer-controller/assisted-installer-controller-cm.yaml.template
@@ -11,5 +11,5 @@ data:
   skip-cert-verification: '{{.SkipCertVerification}}'
   ca-cert-path: '{{.CACertPath}}'
   check-cluster-version: '{{.CheckCVO}}'
-  high-availability-mode: {{.HaMode}}
+  control-plane-count: {{.HaMode}}
   must-gather-image: '{{.MustGatherImage}}'
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -91,6 +91,7 @@ type ControllerConfig struct {
 	Namespace               string `envconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
 	OpenshiftVersion        string `envconfig:"OPENSHIFT_VERSION" required:"true"`
 	HighAvailabilityMode    string `envconfig:"HIGH_AVAILABILITY_MODE" required:"false" default:"Full"`
+	ControlPlainCount       int
 	WaitForClusterVersion   bool   `envconfig:"CHECK_CLUSTER_VERSION" required:"false" default:"false"`
 	MustGatherImage         string `envconfig:"MUST_GATHER_IMAGE" required:"false" default:""`
 	DryRunEnabled           bool   `envconfig:"DRY_ENABLE" required:"false" default:"false"`
@@ -1255,7 +1256,7 @@ func (c *controller) logHostResolvConf() {
 // It will patch router to add access logs
 // It will run http router health check to see if router is healthy on host network
 func (c *controller) logRouterStatus() {
-	if !c.status.HasError() || c.HighAvailabilityMode != models.ClusterHighAvailabilityModeNone {
+	if !c.status.HasError() || c.HighAvailabilityMode != models.ClusterHighAvailabilityModeNone || c.ControlPlainCount > 1 {
 		return
 	}
 	c.log.Infof("Start checking router status")
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -90,7 +90,6 @@ type ControllerConfig struct {
 	CACertPath              string `envconfig:"CA_CERT_PATH" required:"false" default:""`
 	Namespace               string `envconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
 	OpenshiftVersion        string `envconfig:"OPENSHIFT_VERSION" required:"true"`
-	HighAvailabilityMode    string `envconfig:"HIGH_AVAILABILITY_MODE" required:"false" default:"Full"`
 	ControlPlainCount       int
 	WaitForClusterVersion   bool   `envconfig:"CHECK_CLUSTER_VERSION" required:"false" default:"false"`
 	MustGatherImage         string `envconfig:"MUST_GATHER_IMAGE" required:"false" default:""`
@@ -1256,7 +1255,7 @@ func (c *controller) logHostResolvConf() {
 // It will patch router to add access logs
 // It will run http router health check to see if router is healthy on host network
 func (c *controller) logRouterStatus() {
-	if !c.status.HasError() || c.HighAvailabilityMode != models.ClusterHighAvailabilityModeNone || c.ControlPlainCount > 1 {
+	if !c.status.HasError() || c.ControlPlainCount != 1 {
 		return
 	}
 	c.log.Infof("Start checking router status")
--- a/src/assisted_installer_controller/assisted_installer_controller.go
+++ b/src/assisted_installer_controller/assisted_installer_controller.go
@@ -90,7 +90,7 @@ type ControllerConfig struct {
 	CACertPath              string `envconfig:"CA_CERT_PATH" required:"false" default:""`
 	Namespace               string `envconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
 	OpenshiftVersion        string `envconfig:"OPENSHIFT_VERSION" required:"true"`
-	ControlPlainCount       int
+	ControlPlaneCount       int    `envconfig:"CONTROL_PLANE_COUNT" required:"false" default:"3"`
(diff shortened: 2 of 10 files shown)


andrej1991:
@danmanor

src/assisted_installer_controller/assisted_installer_controller.go
```diff
@@ -90,7 +90,7 @@ type ControllerConfig struct {
 	CACertPath              string `envconfig:"CA_CERT_PATH" required:"false" default:""`
 	Namespace               string `envconfig:"NAMESPACE" required:"false" default:"assisted-installer"`
 	OpenshiftVersion        string `envconfig:"OPENSHIFT_VERSION" required:"true"`
-	HighAvailabilityMode    string `envconfig:"HIGH_AVAILABILITY_MODE" required:"false" default:"Full"`
+	ControlPlainCount       int
```
pastequo, line 93: nit: ControlPlaneCount ?

What about adding an `envconfig` and `required` to be consistent ?

src/assisted_installer_controller/assisted_installer_controller_test.go
```diff
@@ -1701,15 +1701,15 @@ var _ = Describe("installer HostRoleMaster role", func() {
 		})
 
 		It("Validate router check will not run with no error", func() {
-			assistedController.HighAvailabilityMode = models.ClusterHighAvailabilityModeNone
+			assistedController.ControlPlainCount = 1
```
pastequo, line 1704: same nit as above for this file

src/assisted_installer_controller/assisted_installer_controller.go
pastequo: If the e2e are ok with the typo, does it mean that we lack some coverage somewhere ?

src/config/config.go
```diff
@@ -85,6 +85,7 @@ func (c *Config) ProcessArgs(args []string) {
 	var installerArgs string
 	flagSet.StringVar(&installerArgs, "installer-args", "", "JSON array of additional coreos-installer arguments")
 	h := flagSet.Bool("help", false, "Help message")
+	highAvailability := flagSet.String("high-availability-mode", "", "valid values: full/none")
```
pastequo, line 88: We are keeping this to be compatible while doing the migration ? What about adding a deprecation notice ?

src/ops/ops.go
```diff
@@ -491,7 +489,7 @@ func (o *ops) renderControllerPod() error {
 		params["ServiceIPs"] = strings.Split(o.installerConfig.ServiceIPs, ",")
 	}
 
-	if o.installerConfig.HighAvailabilityMode == models.ClusterHighAvailabilityModeNone {
+	if o.installerConfig.ControlPlaneCount == 1 {
```
pastequo, line 492: out of curiosity, is it the long term plan ? Or could we have a cluster that is not a SNO with 1 node control plane & several node workers ?

src/assisted_installer_controller/assisted_installer_controller_test.go
andrej1991: fixed

src/assisted_installer_controller/assisted_installer_controller.go
andrej1991: the test which should test it had the same typo

src/ops/ops.go
andrej1991: I'm not sure about it, currently the web GUI does not allow such deployment

deploy/assisted-installer-controller/assisted-installer-controller-cm.yaml.template
```diff
@@ -11,5 +11,5 @@ data:
   skip-cert-verification: '{{.SkipCertVerification}}'
   ca-cert-path: '{{.CACertPath}}'
   check-cluster-version: '{{.CheckCVO}}'
-  high-availability-mode: {{.HaMode}}
+  control-plane-count: '{{.ControllerCount}}'
```
danmanor, line 14: Lets name this `ControlPlaneCount` like everywhere else

src/config/config.go
```diff
@@ -73,7 +73,7 @@ func (c *Config) ProcessArgs(args []string) {
 	flagSet.StringVar(&c.HTTPSProxy, "https-proxy", "", "A proxy URL to use for creating HTTPS connections outside the cluster")
 	flagSet.StringVar(&c.NoProxy, "no-proxy", "", "A comma-separated list of destination domain names, domains, IP addresses, or other network CIDRs to exclude proxying")
 	flagSet.StringVar(&c.ServiceIPs, "service-ips", "", "All IPs of assisted service node")
-	flagSet.StringVar(&c.HighAvailabilityMode, "high-availability-mode", "", "high-availability expectations, \"Full\" which represents the behavior in a \"normal\" cluster. Use 'None' for single-node deployment. Leave this value as \"\" for workers as we do not care about HA mode for workers.")
+	flagSet.IntVar(&c.ControlPlaneCount, "control-plane-count", 0, "The number of controller nodes in the cluster")
```
danmanor, line 76: Lets change to `The number of control plane nodes in the cluster`

src/config/config.go
```diff
@@ -85,6 +85,7 @@ func (c *Config) ProcessArgs(args []string) {
 	var installerArgs string
 	flagSet.StringVar(&installerArgs, "installer-args", "", "JSON array of additional coreos-installer arguments")
 	h := flagSet.Bool("help", false, "Help message")
+	highAvailability := flagSet.String("high-availability-mode", "", "WARNING: this parameter will be shortly deprecated! valid values: full/none")
```
danmanor, line 88: Lets keep it as is for now

```
flagSet.StringVar(&c.HighAvailabilityMode, "high-availability-mode", "", "high-availability expectations, \"Full\" which represents the behavior in a \"normal\" cluster. Use 'None' for single-node deployment. Leave this value as \"\" for workers as we do not care about HA mode for workers.")
```

src/config/config.go
```diff
@@ -123,6 +124,15 @@ func (c *Config) ProcessArgs(args []string) {
 		utils.SetNoProxyEnv(c.NoProxy)
 	}
 
+	if highAvailability != nil && *highAvailability != "" {
```
danmanor, line 127: I guess once you will use `flagSet.StringVar` you won't need the nil check ? also you will have "" default so yo will not need the `if` at all

danmanor (changes requested):

deploy/assisted-installer-controller/assisted-installer-controller-cm.yaml.template
```diff
@@ -11,5 +11,5 @@ data:
   skip-cert-verification: '{{.SkipCertVerification}}'
   ca-cert-path: '{{.CACertPath}}'
   check-cluster-version: '{{.CheckCVO}}'
-  high-availability-mode: {{.HaMode}}
+  control-plane-count: '{{.ControlPlaneCount}}'
```
danmanor, line 14: What should `ControlPlaneCount` match ?

andrej1991: I guess it should match https://github.com/andrej1991/assisted-installer/blob/MGMT-19421-phase-1/src/ops/ops.go#L462

src/config/config.go
```diff
@@ -82,9 +82,10 @@ func (c *Config) ProcessArgs(args []string) {
 	flagSet.BoolVar(&c.NotifyNumReboots, "notify-num-reboots", false, "indicate number of reboots should be notified as event")
 	flagSet.StringVar(&c.CoreosImage, "coreos-image", "", "CoreOS image to install to the existing root")
 
-	var installerArgs string
+	var highAvailability, installerArgs string
 	flagSet.StringVar(&installerArgs, "installer-args", "", "JSON array of additional coreos-installer arguments")
 	h := flagSet.Bool("help", false, "Help message")
+	flagSet.StringVar(&highAvailability, "high-availability-mode", "", "high-availability expectations, \"Full\" which represents the behavior in a \"normal\" cluster. Use 'None' for single-node deployment. Leave this value as \"\" for workers as we do not care about HA mode for workers.")
```
danmanor, line 88: NIT - Probably best just keep 

```
flagSet.StringVar(&c.HighAvailabilityMode, "high-availability-mode", "", "high-availability expectations, \"Full\" which represents the behavior in a \"normal\" cluster. Use 'None' for single-node deployment. Leave this value as \"\" for workers as we do not care about HA mode for workers.")
```

as it is, in effort to avoid creating unnecessary diff that can cause confusion along the way

src/config/config.go
```diff
@@ -82,9 +82,10 @@ func (c *Config) ProcessArgs(args []string) {
 	flagSet.BoolVar(&c.NotifyNumReboots, "notify-num-reboots", false, "indicate number of reboots should be notified as event")
 	flagSet.StringVar(&c.CoreosImage, "coreos-image", "", "CoreOS image to install to the existing root")
 
-	var installerArgs string
+	var highAvailability, installerArgs string
```
danmanor, line 85: See comment below, then this is redundant

src/config/config.go
```diff
@@ -73,7 +73,7 @@ func (c *Config) ProcessArgs(args []string) {
 	flagSet.StringVar(&c.HTTPSProxy, "https-proxy", "", "A proxy URL to use for creating HTTPS connections outside the cluster")
 	flagSet.StringVar(&c.NoProxy, "no-proxy", "", "A comma-separated list of destination domain names, domains, IP addresses, or other network CIDRs to exclude proxying")
 	flagSet.StringVar(&c.ServiceIPs, "service-ips", "", "All IPs of assisted service node")
-	flagSet.StringVar(&c.HighAvailabilityMode, "high-availability-mode", "", "high-availability expectations, \"Full\" which represents the behavior in a \"normal\" cluster. Use 'None' for single-node deployment. Leave this value as \"\" for workers as we do not care about HA mode for workers.")
```
danmanor, line 76: See comment below, we want to avoid this removal / addition diff

danmanor (commented):
Looks good !

danmanor (approved):

PR #1076, opened 2025-03-24
MGMT-20153: Fix ostree race between node-image-pull and installing to rootfs

carbonin:
`ostree container unencapsulate` panics if the image passed to it is already present in the repo. Use `ostree container image pull` instead. Additionally attempting to pull the same os image as the `node-image-pull` service at the same time also can lead to errors writing to the repo. So make `startBootstrap` run synchronously with the rest of `InstallNode` to ensure we are don applying the node image before attempting to pull the image for install.

Resolves https://issues.redhat.com/browse/MGMT-20153


src/ops/ops.go
```diff
@@ -126,24 +129,47 @@ func (o *ops) SystemctlAction(action string, args ...string) error {
 	return errors.Wrapf(err, "Failed executing systemctl %s %s", action, args)
 }
 
-var ostreeOutputRegex = regexp.MustCompile(`Imported: (\w+)`)
+func refFilePath(refName string) string {
+	return path.Join(ostreeRepo, "refs/heads", refName)
+}
 
 func (o *ops) importOSTreeCommit(liveLogger io.Writer) (string, error) {
```
tsorya, line 136: maybe at some point worth thinking about ostree object?

src/ops/ops.go
```diff
@@ -174,6 +200,20 @@ func (o *ops) WriteImageToExistingRoot(liveLogger io.Writer, ignitionPath string
 		return errors.Wrapf(err, "failed to remount boot: %s", out)
 	}
 
+	// Remove the node image reference if it still exists, but also ensure it isn't garbage collected just yet
+	if o.FileExists(refFilePath(nodeImageOSTreeRefName)) {
```
tsorya, line 204: Lets move this part to function

src/ops/ops.go
carbonin: :shrug: I'm not sure what it gains us right now.
Also the not-so-long-term plan I think is to rip all this out and move to bootc so I don't know how much we really need to change.

carbonin: I can put this stuff in a separate file if that makes it a bit more organized

cgwalters:
I only skimmed this but FYI a lot of this I would consider internal implementation details subject to change, especically the ostree ref structures.

I'd like to have a good understanding of exactly what this is trying to do because it will need to be supported directly in bootc - and for bootc we have plans to switch to a new composefs-native store and stop doing the ostree-container flow, so directly invoking `ostree` will break when that happens.

src/ops/ops.go
```diff
@@ -126,24 +129,67 @@ func (o *ops) SystemctlAction(action string, args ...string) error {
...
+		return errors.Wrapf(err, "failed to update temp node image checkout: %s", out)
 	}
 
-	return matches[1], nil
+	return nil
 }
 
-func ostreeArgs(commit string, installerArgs []string) []string {
+func ostreeArgs(ref string, installerArgs []string) []string {
 	ostreeArgs := []string{"admin", "deploy",
```
cgwalters, line 193: We had a live chat on this and I think you can drop a lot of the low-level ostree commands if you switch to using `ostree container image deploy` instead

carbonin: Does this command have something similar to the `--karg-append` and `--karg-delete` that I'm using below?

I see `--karg`, but not the others.

carbonin: I don't see something in the `--help`

cgwalters: `--karg` is the same as `--karg-append`, now you're right it doesn't have `--karg-delete` right now; would be easy to add but can you elaborate on why it's needed?

We include almost no default kernel arguments so I am uncertain when one would need to delete any.

carbonin: Also I went down this path for a while and it seems like it's creating the content in the stateroot, but it doesn't seem to be creating the boot entry. Is that expected? Do I still need to do something else to handle that?

cgwalters: Should definitely be creating the boot entry too

carbonin: >can you elaborate on why it's needed?

Adding and deleting kargs is an API we've exposed through assisted because they were options for coreos-installer. I'm just trying to keep parity here.

carbonin: I'm (effectively) running this in my last attempt:

`ostree container image deploy --stateroot install --sysroot /sysroot --image <rhcos-content-image-from-release>`

Then the next boot fails.

I booted a live ISO and mounted the boot and sysroot partitions and I see the old boot entry in the boot partition's entry, but I see the new one in the sysroot `/boot/loader/entries` directory ... This doesn't seem right. Should I have maybe specified `/` for `--sysroot` in the ostree command?

jlebon:
Let's work on getting https://github.com/openshift/installer/pull/9570 tested and merged so we can clean up the ref manipulations here a bit.

src/ops/ops.go
carbonin: >Yeah, try with --sysroot=/

This seems to have helped. Doing some more testing then will update the PR.

>rpm-ostree kargs --os=install --append/--delete ... should work to edit the kargs post-deployment

Nice, thanks. I'll try this out too.

carbonin:
>Let's work on getting https://github.com/openshift/installer/pull/9570 tested and merged so we can clean up the ref manipulations here a bit.

I'll take a look at this after I get a reasonable general approach working. Once I have something that works and folks are happy with I'll test pulling out my addition to remove the node image ref in favor of the installer change.

I suspect if we want to test with 4.19 ec releases right after this is merged though I'll need to at least keep the `touch` in there then remove it after we get a build with your changes (or just after GA to ensure we're not using releases without the change)

carbonin:
This requires https://github.com/openshift/assisted-installer/pull/1076. Without it the cleanup service fails on the installed node.

src/ops/ops.go
```diff
@@ -151,6 +151,40 @@ func (o *ops) removeNodeImage(liveLogger io.Writer) error {
...
+	for i, arg := range installerArgs {
+		if arg == "--append-karg" && i < lastArgIndex {
+			commandArgs = append(commandArgs, "--append", installerArgs[i+1])
+			continue
+		}
+		if arg == "--delete-karg" && i < lastArgIndex {
+			commandArgs = append(commandArgs, "--delete-if-present", installerArgs[i+1])
+			continue
+		}
+	}
```
jlebon: Feels like we should error out if `--append-karg` or `--delete-karg` _is_ last rather than do our best to work with it.

jlebon (approved):
Looks sane overall. One thing that might be worth mentioning in the first commit message is that with the switch to the layered node image soon, the unencapsulate flow will just not work anyway so it's good that we're switching away from that.

src/ops/ops.go
carbonin: I _think_ we have a validation around this in assisted (where the user adds these). And if we don't I'd rather it fail there than here.

src/ops/ops.go
```diff
@@ -130,6 +130,19 @@ func (o *ops) SystemctlAction(action string, args ...string) error {
 	return errors.Wrapf(err, "Failed executing systemctl %s %s", action, args)
 }
 
+func (o *ops) remountFilesystems(liveLogger io.Writer) error {
```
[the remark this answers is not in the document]
carbonin: At the very least we still need boot because we're putting networking files and the ignition there. I never considered that I might be able to remove sysroot

carbonin:
/hold

until the assisted-service PR is merged.

PR #1079, opened 2025-03-25
MGMT-11622: operator status should be logged

rivkyrizel:
In this PR, we improve the logging by adding the operator's availability status alongside its name. This way, the logs will not only show that we are checking the operator's status, but also what the actual result is. This makes the logs more informative and helps us better understand the outcome of the availability check without having to look at additional data.


eliorerz (approved):
LGTM

PR #1106, opened 2025-04-24
MGMT-20479: Update assisted-installer to allow TNA Clusters

giladravid16:
Currently the assisted-installer running on the bootstrap waits for 2 control plane nodes to be ready, but with TNA it's possible that there will only be 1 control plane node (without the bootstrap).
On clusters like that we need to wait for only 1 control plane node to be ready, and then wait for at least 1 arbiter node to be ready.

Closes [MGMT-20479](https://issues.redhat.com//browse/MGMT-20479)

### DIFF
--- a/src/installer/installer.go
+++ b/src/installer/installer.go
@@ -588,6 +588,10 @@ func (i *installer) waitForControlPlane(ctx context.Context) error {
 		return err
 	}
 
+	if err = i.waitForMinArbiterNodes(ctx, kc, cluster.Platform, hasValidvSphereCredentials); err != nil {
+		return err
+	}
+
 	i.waitForBootkube(ctx)
 	if err = i.waitForETCDBootstrap(ctx); err != nil {
 		i.log.Error(err)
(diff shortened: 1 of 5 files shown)


src/installer/installer.go
```diff
@@ -661,7 +661,12 @@ func (i *installer) workerWaitFor2ReadyMasters(ctx context.Context) error {
 }
 
 func (i *installer) waitForMinMasterNodes(ctx context.Context, kc k8s_client.K8SClient, platform *models.Platform, hasValidCredentials bool) error {
-	i.waitForMasterNodes(ctx, minMasterNodes, kc, platform, hasValidCredentials)
+	controlPlaneReplicas, err := kc.GetControlPlaneReplicas()
```
carbonin, line 664: Is this correct for arbiter? Or should we be waiting for either two CP nodes or one CP and one Arbiter (if the bootstrap is on a CP node rather than the arbiter?)

giladravid16: The bootstrap is always on a CP node, arbiter nodes are not valid for it.

I don't think waiting for an arbiter node to be ready does anything at this point since it waits for etcd bootstrap after that.
But I haven't tested what happens to the installation if the arbiter fails at this point, I assume that it will be stuck on waiting for etcd bootstrap to finish but I can check.

giladravid16: After testing this, if the arbiter node has an error then the installation gets stuck immediately after this while waiting for bootkube (because of cluster-bootstrap).
But I updated it anyway to wait for at least 1 arbiter node because it would make more sense and make the logs more helpful.

src/k8s_client/k8s_client.go
```diff
@@ -287,6 +288,27 @@ func (c *k8sClient) GetControlPlaneReplicas() (int, error) {
 	return controlPlaneDecoder.ControlPlane.Replicas, nil
 }
 
+func (c *k8sClient) GetArbiterReplicas() (int, error) {
+	installConfig, err := c.getInstallConfig()
+	if err != nil {
+		return 0, err
+	}
+
+	var ArbiterDecoder struct {
```
carbonin, line 297: Probably doesn't needs to be caps, right?

carbonin (commented):
Looks good, just one very minor comment

src/k8s_client/k8s_client.go
giladravid16: yes

carbonin (approved):

PR #1113, opened 2025-04-30
MGMT-20534: Move getting Node replicas into waitForNodes

CrystalChun:
Originally, GetControlPlaneReplicas and GetArbiterReplicas were called once outside a retry. These functions can fail since they call the spoke cluster's kube api server, which most likely is not ready by the time the installer is checking for these Nodes. This leads to the cluster prematurely failing because of the error returned from these functions.

This moves the functions into the retry block that waitForNodes already uses so that the whole install doesn't fail when the spoke's kube apiserver is not ready yet.

---

Failures encountered in slack thread: https://redhat-internal.slack.com/archives/C014N2VLTQE/p1745997401995939
Original code added here https://github.com/openshift/assisted-installer/commit/c43bcb972dc310561de5649370e71172311ae090#diff-1046cc2d18cf5f82336bbad36a2d28540606e1c6aaa0b5073c545301ef60ffd4R670

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,7 +33,6 @@ require (
 	github.com/vincent-petithory/dataurl v1.0.0
 	go.uber.org/mock v0.4.0
 	golang.org/x/net v0.33.0
-	golang.org/x/sync v0.11.0
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.29.9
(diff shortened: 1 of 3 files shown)


giladravid16:
Thanks Crystal
/lgtm

PR #1252, opened 2025-09-16
NO-ISSUE: Configure konflux pipelines for ACM 2.12

giladravid16:


Dockerfile.assisted-installer-controller-mce
```diff
@@ -0,0 +1,48 @@
+FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi9/go-toolset:1.21 AS builder
```
pastequo, line 1: Could be another task but isn't it a bit old ?

gamli75 (approved):

PR #1270, opened 2025-09-29
[release-ocm-2.14] - ACM-24326,ACM-24321: CVE-2025-47907 Upgrade golang to 1.24

omer-vishlitzky:
## Jira Tickets
- https://issues.redhat.com/browse/ACM-24326
- https://issues.redhat.com/browse/ACM-24321

##
Update go version to 1.24 to address cve CVE-2025-47907

##
This PR updates tests to be compatible with Go 1.24 and golangci-lint 1.64.8.
                                                                                                                                                                                       
**TLS test changes (`ops_test.go`)**: Go 1.24 now [enforces a minimum 1024-bit RSA key size](https://tip.golang.org/doc/go1.24#crypto/rsa) for all TLS operations. The old `ghttp.NewTLSServer()` implementation generates certificates that are too weak and fail the new requirement. We've switched to using `httptest.NewTLSServer()` directly, which generates proper 2048-bit certificates. Since these certificates are dynamically generated per test, we had to change the function signatures from `withCert bool` to `ca []byte` to pass the actual certificate bytes through the test helpers.

**Loop variable changes (`assisted_installer_controller_test.go`)**: golangci-lint 1.64.8 includes the [copyloopvar](https://golangci-lint.run/usage/linters/#copyloopvar) linter which catches a common Go gotcha where loop variables are captured by reference in closures. The fix is simple - copy `value.Host` to a local variable before passing its fields to gomock expectations. This ensures each iteration uses the correct host data.

### DIFF
--- a/src/ops/ops_test.go
+++ b/src/ops/ops_test.go
@@ -5,10 +5,12 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"net/http/httptest"
 	"os"
 	"reflect"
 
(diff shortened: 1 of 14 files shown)


src/ops/ops_test.go
```diff
@@ -220,22 +219,26 @@ WkBKOclmOV2xlTVuPw==
...
 			ign, ca, err := o.getPointedIgnitionAndCA(ignitionPath)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(ign).To(Equal(source))
-			Expect(ca).To(Equal(string(localhostCert)))
+			expectedCA := ""
+			if len(caBytes) > 0 {
+				expectedCA = string(caBytes)
+			}
+			Expect(ca).To(Equal(expectedCA))
 		}
```
omer-vishlitzky, line 236: Because the helper now skips adding a CA when we pass nil, the assertion needs to tolerate an empty string; otherwise copyloopvar/gosec flagged the mismatch.

omer-vishlitzky:
This PR updates tests to be compatible with Go 1.24 and golangci-lint 1.64.8.
                                                                                                                                                                                       
**TLS test changes (`ops_test.go`)**: Go 1.24 now [enforces a minimum 1024-bit RSA key size](https://tip.golang.org/doc/go1.24#crypto/rsa) for all TLS operations. The old `ghttp.NewTLSServer()` implementation generates certificates that are too weak and fail the new requirement. We've switched to using `httptest.NewTLSServer()` directly, which generates proper 2048-bit certificates. Since these certificates are dynamically generated per test, we had to change the function signatures from `withCert bool` to `ca []byte` to pass the actual certificate bytes through the test helpers.

**Loop variable changes (`assisted_installer_controller_test.go`)**: golangci-lint 1.64.8 includes the [copyloopvar](https://golangci-lint.run/usage/linters/#copyloopvar) linter which catches a common Go gotcha where loop variables are captured by reference in closures. The fix is simple - copy `value.Host` to a local variable before passing its fields to gomock expectations. This ensures each iteration uses the correct host data.

PR #1271, opened 2025-09-29
[release-ocm-2.13] - CVE-2025-47907: Upgrade golang to 1.24.6

omer-vishlitzky:
Security fix for CVE-2025-47907:
Upgrading golang to 1.24.6

### DIFF
--- a/Dockerfile.assisted-installer
+++ b/Dockerfile.assisted-installer
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.21 AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.24.6 AS builder
 ARG TARGETPLATFORM
 ENV GOFLAGS=-mod=mod
 WORKDIR /go/src/github.com/openshift/assisted-installer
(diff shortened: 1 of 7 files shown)


src/ops/ops_test.go
```diff
@@ -188,28 +189,26 @@ grw/ZQTTIVjjh4JBSW3WyWgNo/ikC1lrVxzl4iPUGptxT36Cr7Zk2Bsg0XqwbOvK
...
 			})
 
-		if !withCert {
-			return ret
+		if len(ca) > 0 {
+			ret.Ignition.Security.TLS.CertificateAuthorities = append(ret.Ignition.Security.TLS.CertificateAuthorities,
+				types.Resource{
+					Source: swag.String(dataurl.EncodeBytes(ca)),
+				})
 		}
```
omer-vishlitzky, line 206: Lint 1.64.8 forced the production helper to accept raw CA bytes (so we only append a TLS authority when one exists). Switching the test helper to `[]byte` mirrors that behaviour; sticking with the old boolean toggle caused gosec/copyloopvar failures because the new code paths were never covered.

src/ops/ops_test.go
```diff
@@ -324,7 +330,10 @@ WkBKOclmOV2xlTVuPw==
 					_, err := io.WriteString(w, "")
 					Expect(err).ToNot(HaveOccurred())
 				})
-			checkMcsIgnition(s.URL(), true, false)
+			cert := s.HTTPTestServer.Certificate()
```
omer-vishlitzky, line 333: Grabbing the test server’s cert and encoding it mirrors the production path (and keeps gosec happy); the static localhost blob no longer matched the bytes once the helper only includes a CA when explicitly provided.

PR #1273, opened 2025-09-29
[release-ocm-2.11] - CVE-2025-47907: Upgrade golang to 1.24.6

omer-vishlitzky:
Security fix for CVE-2025-47907:
Upgrading golang to 1.24.6


src/ops/ops_test.go
```diff
@@ -187,28 +188,26 @@ grw/ZQTTIVjjh4JBSW3WyWgNo/ikC1lrVxzl4iPUGptxT36Cr7Zk2Bsg0XqwbOvK
...
+					Source: swag.String(dataurl.EncodeBytes(ca)),
+				})
 		}
 
-		ret.Ignition.Security.TLS.CertificateAuthorities = append(ret.Ignition.Security.TLS.CertificateAuthorities,
-			types.Resource{
-				Source: swag.String(dataurl.EncodeBytes(localhostCert)),
-			})
-
 		return ret
```
omer-vishlitzky, line 206: Lint 1.64.8 forced the production helper to accept raw CA bytes (so we only append a TLS authority when one exists). Switching the test helper to `[]byte` mirrors that behaviour; sticking with the old boolean toggle caused gosec/copyloopvar failures because the new code paths were never covered.

src/ops/ops_test.go
```diff
@@ -219,22 +218,26 @@ WkBKOclmOV2xlTVuPw==
...
+			if len(caBytes) > 0 {
+				expectedCA = string(caBytes)
+			}
+			Expect(ca).To(Equal(expectedCA))
 		}
 		It("from bootstrap", func() {
-			checkSource("https://abc.com")
+			checkSource("https://abc.com", localhostCert)
 		})
 		It("embedded", func() {
```
omer-vishlitzky, line 239: Because the helper can now return an empty CA when no PEM was supplied, the assertion has to tolerate both cases—otherwise the stricter lint run marked the branch as dead code.

src/ops/ops_test.go
```diff
@@ -323,7 +329,10 @@ WkBKOclmOV2xlTVuPw==
 					_, err := io.WriteString(w, "")
 					Expect(err).ToNot(HaveOccurred())
 				})
-			checkMcsIgnition(s.URL(), true, false)
+			cert := s.HTTPTestServer.Certificate()
+			Expect(cert).ToNot(BeNil())
```
omer-vishlitzky, line 333: Grabbing the test server’s cert and encoding it mirrors the production path (and keeps gosec happy); the static localhost blob no longer matched the bytes once the helper only includes a CA when explicitly provided.

omer-vishlitzky:
This PR updates tests to be compatible with Go 1.24 and golangci-lint 1.64.8.
                                                                                                                                                                                       
**TLS test changes (`ops_test.go`)**: Go 1.24 now [enforces a minimum 1024-bit RSA key size](https://tip.golang.org/doc/go1.24#crypto/rsa) for all TLS operations. The old `ghttp.NewTLSServer()` implementation generates certificates that are too weak and fail the new requirement. We've switched to using `httptest.NewTLSServer()` directly, which generates proper 2048-bit certificates. Since these certificates are dynamically generated per test, we had to change the function signatures from `withCert bool` to `ca []byte` to pass the actual certificate bytes through the test helpers.

**Loop variable changes (`assisted_installer_controller_test.go`)**: golangci-lint 1.64.8 includes the [copyloopvar](https://golangci-lint.run/usage/linters/#copyloopvar) linter which catches a common Go gotcha where loop variables are captured by reference in closures. The fix is simple - copy `value.Host` to a local variable before passing its fields to gomock expectations. This ensures each iteration uses the correct host data.

gamli75 (approved):

PR #1307, opened 2025-10-26
Fix eFi boot when device mapper is installation disk.

bkopilov:
In case the installation disk is a mapper the valid path to mount is /dev/mapper/dm device. Added retry for lsblk command , lsblk children not updated immediatly when slow disks.


bkopilov:
Duplicate of https://github.com/openshift/assisted-installer/pull/1308

PR #1308, opened 2025-10-26
MGMT-20756: assisted installer naive string concatenation for partiti…

bkopilov:
In case the installation disk is a mapper the valid path to mount is /dev/mapper/dm device. Added retry for lsblk command , lsblk children not updated immediatly when slow disks.

### DIFF
--- a/src/ops/ops.go
+++ b/src/ops/ops.go
@@ -367,9 +367,16 @@ func (o *ops) SetBootOrder(device string) error {
 	}
 
 	o.log.Info("Setting efibootmgr to boot from disk")
-	efiDirname, err := o.findEfiDirectory(device)
+	var efiDirname string
+	var findError error
+	//Retry find partitions due to delayed update on device-mappers
+	err = utils.WaitForPredicate(20*time.Second, 5*time.Second, func() bool {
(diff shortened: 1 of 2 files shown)


src/ops/ops.go
```diff
@@ -1058,11 +1072,13 @@ func (o *ops) getPartitionPathFromLsblk(device, partitionNumber string) (string,
 	if err = json.Unmarshal([]byte(ret), &disks); err != nil {
 		return "", errors.Wrap(err, "failed to unmarshal lsblk output")
 	}
+	o.log.Infof("--> lsblk output before:  %s", ret)
```
omertuc: Stray debug log, was it meant to be committed?

src/ops/ops.go
```diff
@@ -103,6 +103,19 @@ func NewOps(logger *logrus.Logger, executor execute.Execute) Ops {
 	return NewOpsWithConfig(&config.Config{}, logger, executor)
 }
 
+func RetryFindDevice(attempts int, sleep time.Duration, fn func(device string) (string, error), device string) (string, error) {
```
omertuc: Can we leverage `utils.WaitForPredicateWithContext` here?

PR #1384, opened 2025-11-26
AGENT-1216: support new HostStage (HostStageCopyingRegistryDataToDisk)

danielerez:
Added support for the new HostStage: HostStageCopyingRegistryDataToDisk.
I.e. when copying registry data to disk, move to the new stage and pass it to logger (for monitoring purposes).
Introduced in assisted-service by: https://github.com/openshift/assisted-service/pull/8423


andfasano:
Hi @fracappa why do need to backport that? It's not a bug

PR #2200, opened 2026-07-02
NO-ISSUE: Bump linter

pastequo:


pastequo:
Might be a chicken & eggs problem

PR #2230, opened 2026-07-29
OCPBUGS-96696: Wait for all nodes to join before exiting for ABI

bfournie:
Fixes the problem with ABI when installing a multi-node cluster. In this case the bootstrap (rendezvous) node is the last to reboot and join the cluster. The waitForInstallationAgentBasedInstaller() function exits as soon as ClusterVersion Available=True, which can happen with N-1 nodes. This cancels the mainContext, stopping the ApproveCsrs goroutine before it can approve the rendezvous node's kubelet CSRs, therefore leaving master-0 permanently unable to join the cluster.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved agent-based installation monitoring to use context-aware polling that stops on cancellation or timeout.
  * Now waits for cluster availability before proceeding, and—when configured—also waits for the expected control-plane node count.
  * For single-node/SNO scenarios, skips node-count checks when no nodes are expected.

* **Tests**
  * Expanded agent-based installer waiting coverage with context usage and interval tuning to verify partial-to-complete node readiness and immediate exit when no node count is expected.

### DIFF
--- a/src/main/assisted-installer-controller/assisted_installer_main.go
+++ b/src/main/assisted-installer-controller/assisted_installer_main.go
@@ -164,7 +164,7 @@ func main() {
 			wg.Add(1)
 			go assistedController.PostInstallConfigsK8sClient(mainContext, &wg, bootstrapKubeconfigForSNO)
 
-			waitForInstallationAgentBasedInstaller(kc, logger, false)
+			waitForInstallationAgentBasedInstaller(mainContext, kc, logger, false, 0)
 			return
 		}
 		// With the agent-based installer, assisted-service runs on the bootstrap node.
@@ -234,7 +234,7 @@ func main() {
 	// monitoring installation by cluster status
 	switch invoker {
 	case common.InvokerAgent:
-		waitForInstallationAgentBasedInstaller(kc, logger, removeUninitializedTaint)
+		waitForInstallationAgentBasedInstaller(mainContext, kc, logger, removeUninitializedTaint, Options.ControllerConfig.ControlPlaneCount)
 	default:
 		waitForInstallation(client, logger, assistedController)
 	}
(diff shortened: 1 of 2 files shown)


src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -310,19 +319,43 @@ func waitForInstallationAgentBasedInstaller(kubeClient k8s_client.K8SClient, log
...
+		if !clusterVersionAvailable {
+			clusterVersion, err := kubeClient.GetClusterVersion()
+			if err != nil {
+				log.WithError(err).Error("Failed to get cluster version from k8s client")
+			} else {
+				for _, condition := range clusterVersion.Status.Conditions {
+					if condition.Type == "Available" && condition.Status == "True" {
+						clusterVersionAvailable = true
+						log.Info("ClusterVersion Available=True")
+						if expectedNodes <= 0 {
```
zaneb: nit: I think the control flow would be clearer if this were inside the `if clusterVersionAvailable` block at line 339.

zaneb (commented):
/approve

src/main/assisted-installer-controller/assisted_installer_main.go
bfournie: Updated

bfournie:
Confirmed with 5 control-plane node installation. The bootstrap (rendezvous) node now appears properly.
```
$ oc get nodes
NAME       STATUS   ROLES                         AGE     VERSION
master-0   Ready    control-plane,master,worker   3h44m   v1.36.2
master-1   Ready    control-plane,master,worker   4h1m    v1.36.2
master-2   Ready    control-plane,master,worker   4h1m    v1.36.2
master-3   Ready    control-plane,master,worker   4h1m    v1.36.2
master-4   Ready    control-plane,master,worker   4h      v1.36.2
```

src/main/assisted-installer-controller/assisted_installer_main.go
```diff
@@ -310,19 +319,43 @@ func waitForInstallationAgentBasedInstaller(kubeClient k8s_client.K8SClient, log
...
 		}
-		for _, condition := range clusterVersion.Status.Conditions {
-			if condition.Type == "Available" && condition.Status == "True" {
-				// cluster install is complete, we can exit
-				log.Info("ClusterVersion Available=True")
+		if clusterVersionAvailable {
+			if expectedNodes <= 0 {
+				return
+			}
+			nodes, err := kubeClient.ListNodesByRole("master")
```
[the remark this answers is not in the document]
bfournie: That fix is out of scope for this PR. Its an existing pattern for k8s clients and not something introduced by this change. Changing it would require refactoring the interface, the mock, and all the callers in the repo.
We don't plan on making that change in this PR.

carbonin (approved):

PR #2249, opened 2026-08-06
[release-ocm-2.17] NO-ISSUE: Fix .gitignore to use /build instead of build

shay23bra:
The bare 'build' rule in .gitignore incorrectly excludes vendored packages containing build/ subdirectories (e.g. docker/docker/api/types/build), causing Konflux and verify-deps failures. Use /build to only ignore the project root build directory, matching the pattern used by assisted-service.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 .vscode/
 .venv/
 venv/
-build/
+/build
 kubevirtci/
 **/RCS/**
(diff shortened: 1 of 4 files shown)


giladravid16:
/retitle [release-ocm-2.17] NO-ISSUE: Fix .gitignore to use /build instead of build

PR #2275, opened 2026-08-25
MGMT-24885: Harden command construction and credential handling

bluesort:
Harden must-gather and log upload commands and credential handling

## Summary by CodeRabbit

- **Security**
  - Improved protection against command injection when collecting diagnostic logs, creating archives, checking host details, and applying manifests.
  - Validates diagnostic image references before execution.
  - Keeps pull-secret credentials out of command-line arguments and redacts sensitive environment values from errors and logs.
- **Reliability**
  - Uses explicit working directories for diagnostic collection to improve consistency.
  - Applies manifests and runs system commands with safer argument handling.
  - Supports secure environment and working-directory configuration for command execution.

### DIFF
--- a/src/ops/execute/execute.go
+++ b/src/ops/execute/execute.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"slices"
 	"strings"
 	"syscall"
 
@@ -17,7 +18,7 @@ import (
 //go:generate mockgen -source=execute.go -package=execute -destination=mock_execute.go
 type Execute interface {
 	ExecCommand(liveLogger io.Writer, command string, args ...string) (string, error)
-	ExecCommandWithContext(ctx context.Context, liveLogger io.Writer, command string, args ...string) (string, error)
+	ExecCommandWithOptions(liveLogger io.Writer, command string, args []string, opts ...CommandOption) (string, error)
 	Execute(command string, args ...string) (string, error)
 }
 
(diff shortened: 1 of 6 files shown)


src/ops/ops.go
```diff
@@ -655,17 +655,33 @@ func (o *ops) UploadInstallationLogs(isBootstrap bool) (string, error) {
...
 		fmt.Sprintf("-insecure=%s", strconv.FormatBool(o.installerConfig.SkipCertVerification)),
 		fmt.Sprintf("-bootstrap=%s", strconv.FormatBool(isBootstrap)))
 
 	if o.installerConfig.CACertPath != "" {
 		args = append(args, fmt.Sprintf("-cacert=%s", o.installerConfig.CACertPath))
 	}
-	return o.ExecPrivilegeCommand(o.logWriter, command, args...)
+
+	// Build nsenter command
+	nsenterArgs := []string{
```
shay23bra: This duplicates the nsenter arg construction from ExecPrivilegeCommand. 
Could we add an ExecPrivilegeCommandWithOptions instead, so the nsenter flags stay in one place?

src/ops/ops.go
```diff
@@ -750,15 +766,44 @@ func (o *ops) GetNumberOfReboots(ctx context.Context, nodeName, kubeconfigPath s
...
+
+	for _, r := range image {
+		if unicode.IsControl(r) {
+			return errors.Errorf("image reference contains control character")
+		}
+	}
+
+	// Supports: registry.io/repo:tag, registry.io:port/repo:tag, repo@sha256:digest
+	pattern := `^[a-zA-Z0-9._-]+(:[0-9]+)?(/[a-zA-Z0-9._/-]+)?(:[a-zA-Z0-9._-]+|@sha256:[a-fA-F0-9]{64})?$`
+	if !regexp.MustCompile(pattern).MatchString(image) {
```
shay23bra: Nit: the regex gets recompiled on every call - might want to pull it into a package-level var.

shay23bra (changes requested):

src/ops/ops.go
bluesort: good call, I've rolled this into a `WithPrivilege` option so they live in a single spot. we had them duplicated in the execute package too

src/ops/ops.go
```diff
@@ -48,6 +49,9 @@ const (
 	ostreeDeployPath                = "/mnt/root/ostree/deploy/rhcos"
 )
 
+// Supports: registry.io/repo:tag, registry.io:port/repo:tag, repo@sha256:digest
```
pastequo: Sometimes its convenient to use both tag and digest, like https://github.com/openshift/assisted-service/blob/master/Dockerfile.assisted-service#L11 : `registry.access.redhat.com/ubi9/ubi:latest@sha256:3816d303e75dec4da2d10eeb9e8651eef4393721598bea4690c607282635aa57` 

It can useful to help automation bot to update the sha part by looking at the tag part (Do I need to update the digest ? I pull the by tag and check if the sha differ)

src/ops/execute/execute.go
```diff
@@ -27,6 +27,67 @@ type executor struct {
...
+func WithEnv(env []string) CommandOption {
+	return func(c *commandConfig) {
+		c.env = append(c.env, env...)
+	}
+}
+
+// WithDir sets the working directory for the command
+func WithDir(dir string) CommandOption {
+	return func(c *commandConfig) {
+		if dir != "" {
```
pastequo: Are we fine that we can't overwrite it once it's set ?

src/ops/execute/execute.go
```diff
@@ -27,6 +27,67 @@ type executor struct {
...
+	return func(c *commandConfig) {
+		c.ctx = ctx
+	}
+}
+
+// WithPrivilege wraps the command in nsenter to execute it in the host
+// environment rather than inside the container.
+func WithPrivilege() CommandOption {
+	return func(c *commandConfig) {
+		c.args = append([]string{
```
pastequo, line 69: Are we fine that if we put WithPrivilege twice (or more) the result will be weird (I guess) ?

src/ops/execute/execute.go
```diff
@@ -53,7 +114,7 @@ func (e *executor) execCommand(liveLogger io.Writer, cmd *exec.Cmd) (string, err
 		cmd.Stdout = &stdoutBuf
 		cmd.Stderr = &stdoutBuf
 	}
-	cmd.Env = e.cmdEnv
+	cmd.Env = append(append([]string(nil), e.cmdEnv...), cmd.Env...)
```
pastequo: The double append looks weird at first sight. There is no way to do just one ?

src/ops/ops.go
bluesort: That's helpful to know, thanks. I'll update to allow it

src/ops/execute/execute.go
bluesort: Practically yes, but there's not benefit in restricting this. I'll remove the check

src/ops/execute/execute.go
bluesort: Yes, this would end up breaking the call so we can catch during review. We could check if `c.command` is already `nsenter` and no-op, but we're then just covering for someone accidentally calling it twice

src/ops/execute/execute.go
bluesort: Good catch, `slices.Concat` does what we need and is cleaner here

